Issue #267, opened 2026-04-07
fix: wasm-codegen-quick (http) — scalar_hello component validation fails


mizchi:
## Problem

`wasm-codegen-quick (http)` CI shard fails with:
```
FAIL: scalar_hello: component validation failed
```

In `test_wasi_p3_e2e.sh`, the P3 component for `scalar_hello` is compiled successfully but fails `wasm-tools validate --features all`.

## Context

- The test compiles a simple HTTP handler to a P3 component using `vibe compile --compose-p3`
- The component wasm is produced but invalid
- This may be a Rust adapter build issue or component composition bug

## Environment

- CI: Ubuntu 24.04 with Rust stable toolchain
- The P3 component path involves Rust adapter compilation (`wit-bindgen`)

## Impact

- `wasm-codegen-quick (http)` shard fails

mizchi:
Status update from #282 investigation:

The `wasm-codegen-quick` shard was split into three sub-causes. Two of them are fixed in #282:

- **(core)** was `test_golden_wat.sh`: the golden `examples/golden_wat/*.wat` fixtures were last regenerated in `0b8145d` (Apr 8), but `e0dd988` (Apr 10) added `compile_public_start_wrapper` which emits an additional wasm function and renames the `_start` vs `_vibe_run_tagged` split. All 12 goldens were stale. Regenerated via `scripts/test_golden_wat.sh --update` in commit f7174d8. Verified green in CI run 24249189779.

- **(compare)** was actually two problems stacked together:
  1. `scripts/test_component_import_contract.sh` used deprecated builtin names (`stdout_write_char`, `http_request`) and called `run()` at the top level. After `0307bc0` introduced "top-level expression has unhandled effect" as a hard type error, both probes failed to compile. Renamed to `Stdout::write_char` / `Http::request` and restructured to `export let _start = ... with {...}` in commit f7174d8.
  2. `vibe test vibe/wasm/wasm_opt/minify_zlib_test.vibe` has been silently failing the compare shard since `d22f719` — the test uses `Fs::read_bytes`, which the compiled wasm test backend does not support, and `cli_test_cmd.mbt` reports `CompiledTestAttempt::Unsupported` as `failed=1` with no way to mark the file as skipped. Removed from the shard in commit 09daa1b with a TODO to restore once the test runner grows a skip path.

  Verified green in CI run 24249189779.

[...]

Issue #287, opened 2026-04-11
selfhost: 4 test files trap `unreachable` in compiled backend (parser_loop, builtins, component_codegen, cli_adapter_cache)


mizchi:
## Problem

Four test files in `vibe/compiler/*_test.vibe` have one-or-more test cases that compile successfully on the compiled wasm test backend but then trap with `wasm trap: unreachable instruction executed` when `wasmtime` runs them. All four pass on the native interpreter. This is the same symptom family as #268 (which turned out to be `StringBuilder::push` realloc silently losing entries), but these are distinct call sites and are probably distinct compiler bugs.

Each one is currently excluded from the compiled bootstrap shard in `scripts/test_selfhost_bootstrap_gate.sh` — restore each to the shard once the underlying bug is fixed.

### 1. `parser_loop_test.vibe`  — 1 / 7 fail

Case: `"parse parameterized loop"` (line 63).

```vibe
test "parse parameterized loop" {
  let result = handle {
    parse_print("loop (i = 10, acc = 0) { if i <= 0 { break acc } else { continue(i - 1, acc + i) } }")
  } with Error {
    Throw(e) => String::concat("ERROR: ", e)
  }
  assert(result == "loop (i = 10, acc = 0) { if (i <= 0) { break acc } else { continue((i - 1), (acc + i)) } }")
}
```

Observation: `parse_print` itself runs (the sub-helpers each produce output), but the printed string has **length ~254**, not the expected 88. When the test asserts `result == "<88-char expected>"`, `String::equals` comparing a 254-char actual against the 88-char expected fails and the surrounding `assert` traps unreachable.

Root cause: unknown — either the printer is emitting the loop body multiple times for parameterized loops, or the parser is duplicating the body in the AST, or (less likely) `String::equals` has a bug on strings with differing lengths. Needs investigation.

### 2. `builtins_test.vibe` — 6 / 62 fail

All six failing cases are `lookup_builtin String::*` probes (lines 181, 188, 195, 202, 209, 216):

```vibe
test "lookup_builtin String::index_of" {
  match lookup_builtin("String::index_of") {
    Some(_) => assert(true),
    None => assert(false)
  }
}
// same shape for String::{contains, starts_with, ends_with, from_char_code, char_code_at}
```

[...]

PR #310, opened 2026-04-19
Restore in-process mwac for --compose-p3, drop wac plug workarounds

### LINKED ISSUE #290: mwac upstream: `plug_components` does not forward typed imports' component type section
## Status

Fixed upstream in `mizchi/mwac@0.2.1`.

In-tree follow-up on branch `claude/review-merge-prs-19DAg` (commit 642bc97):

- Bump `mizchi/mwac` 0.1.6 → 0.2.1 in `moon.mod.json` and drop the local `../mwac` override.
- Revert the `wac plug` workaround in `scripts/test_wasi_p3_e2e.sh`, `scripts/compose_http_p3_handler.sh`, `scripts/vibe_serve.sh` — they now go back through `vibe compile --compose-p3`.
- Drop the #290 entry from `TODO.md`.

---

## Problem

`vibe compile --compose-p3` routes through `src/runtime_compile/compile.mbt:compose_p3_component` → `@mwac.plug_components(adapter_bytes, [app_bytes], best_effort=true)`, which is defined in [mizchi/mwac](https://github.com/mizchi/mwac) (pulled in via `.mooncakes/mizchi/mwac/src/component/plug.mbt`).

When `plug_components` needs to **forward** an import from the socket (adapter) to the outer composed component — e.g. the adapter imports `wasi:http/types@0.3.0-rc-2026-02-09` as an instance of component type 0, and the composed output needs to re-import it at the outer level — `plug.mbt:encode_import` writes out the name and the raw descriptor bytes verbatim:

```moonbit
pub fn encode_import(info : ImportInfo) -&gt; Array[Byte] {
  let out : Array[Byte] = []
  append_bytes(out, encode_externname(info.name))
  append_bytes(out, info.raw_desc.to_array())   // &lt;-- raw bytes from the socket&#39;s import descriptor
  out
}
```

The raw descriptor bytes reference **type indices from the socket's type space** — e.g. `Instance(0)` meaning "type 0 as defined inside the adapter component". `plug_components` does not also emit a component type section at the outer level declaring those referenced types, so the composed wasm has:

```
(component
  (component (;0;) ... adapter with its own type section declaring instance type 0 ...)
  (import "wasi:http/types@0.3.0-rc-..." (instance (;0;) (type 0)))   ;; refers to OUTER type 0, which doesn&#39;t exist
  ...
)
```

`wasm-tools validate --features all` rejects it with:

```
error: unknown type 0: type index out of bounds (at offset 0x17702)
```

## Confirmed: not a vibe-side bug

PR #284 verified this by switching all in-tree test scripts from `vibe compile --compose-p3` to `vibe compile --component-string-lift + wac plug` (`scripts/test_wasi_p3_e2e.sh`, `scripts/compose_http_p3_handler.sh`, `scripts/vibe_serve.sh`). `wac` correctly forwards the type section — it is the reference implementation from BytecodeAlliance — and the same adapter + plug inputs validate cleanly when composed via `wac plug`. So the bug is unambiguously in mwac.

The existing `@mwac/component/plug_test.mbt` tests only cover cases where the socket has **no** typed imports to forward (they assert `composed_info.imports.length() == 0` or equal to the socket's, but all of those socket imports are satisfied by the plug). The "socket imports something the plug cannot satisfy, so it must be forwarded to the outer component" path is untested, which is how this slipped through.

## Proposed fix (upstream in `mizchi/mwac`)

`ComponentDetail` already tracks `types : Array[ComponentTypeDef]`, so the information needed is present. The fix is to walk the forwarded imports, compute the transitive closure of type indices they reference, re-encode those types into a new outer component type section (with re-mapped indices), re-map the `raw_desc` bytes of the forwarded imports to point at the new outer indices, and emit the type section before the import section in the composed output.

A targeted test case: take `deps/wasmtime/crates/wasi-http/src/p3/wit/world.wit` (WASI HTTP P3), build it into an adapter component that imports `wasi:http/types`, and compose it with a trivial plug that provides `handler`. The composed output should validate under `wasm-tools validate --features all`.

Until this is fixed, the in-tree workaround (using `wac plug` from test scripts) is in place — `vibe compile --compose-p3` itself is still broken for any adapter with typed imports.

## Workaround in place

PR #284 switched three test scripts over to the wac-based pipeline:
1. `vibe compile --component-string-lift  -o plug.wasm`
2. `wac plug --plug plug.wasm -o out.component.wasm adapter.wasm`

`wac` is already a hard prerequisite of `build_selfhost_cli_direct_component.sh` (and therefore `selfhost-gates (cli)`), so assuming it on PATH is consistent with the rest of the component-model test fleet.

## Should we track this in `mizchi/mwac` or stay here?

The fix must land in mwac upstream (it's the only repo with the bug). I'd suggest opening a mirror issue at `mizchi/mwac` and having this issue here reference it. When the mwac fix lands and is released, bump `mizchi/mwac` in `moon.mod.json`, revert the `wac plug` workaround in the three scripts, and close this issue.

## Related

- #267 (closed) — symptom report
- PR #284 (merged 72cacc2) — test-side workaround

---

mizchi:
## Summary
Closes #290.

The upstream mwac bug (imports' referenced component type sections not forwarded to the outer composed component) was fixed by [mizchi/mwac@6dc1ae2](https://github.com/mizchi/mwac/commit/6dc1ae2) (type section forwarding for WASI HTTP P3 adapter composition). With that in place we can drop the external `wac plug` workarounds introduced by #267 / #284 / #299.

## Changes
- `src/cmd/vibe/cli_compile_cmd.mbt`: `--compose-p3` goes back through in-process `@runtime_compile.compose_p3_component` (mwac) instead of shelling out to `wac plug`. Removes `compose_p3_component_with_wac` helper.
- `src/cmd/vibe/cli_wbtest.mbt`: drop the wac-command builder test.
- `scripts/test_wasi_p3_e2e.sh`, `scripts/compose_http_p3_handler.sh`, `scripts/vibe_serve.sh`: invoke `vibe compile --compose-p3` instead of assembling `wac plug` commands themselves; `wac` no longer required on `PATH` for these scripts.

## Test plan
- [x] `vibe compile --compose-p3` against the WASI HTTP P3 adapter + a trivial scalar handler plug produces the same bytes as `@mwac.plug_components(adapter, [app], best_effort=true)` run from mwac's own test suite.
- [x] `wasm-tools validate --features all` exits 0 on the composed component.
- [ ] CI (ci-required)
- [ ] `scripts/test_wasi_p3_e2e.sh`

## Follow-up
- Bump `mizchi/mwac` in `moon.mod.json` once 0.2.0 is published to mooncakes (currently the `overrides.path = "../mwac"` resolves against the local checkout).

### DIFF
--- a/moon.mod.json
+++ b/moon.mod.json
@@ -13,11 +13,8 @@
     "mizchi/flatbuffers": "0.1.3",
     "mizchi/wite": "0.6.5",
     "mizchi/similarity": "0.2.1",
-    "mizchi/mwac": "0.1.6",
+    "mizchi/mwac": "0.2.1",
     "moonbitlang/parser": "0.2.5"
-  },
-  "overrides": {
-    "mizchi/mwac": { "path": "../mwac" }
   },
   "readme": "README.md",
   "repository": "https://github.com/mizchi/vibe-lang",
--- a/scripts/compose_http_p3_handler.sh
+++ b/scripts/compose_http_p3_handler.sh
@@ -30,7 +30,6 @@ require_cmd() {
 
 require_cmd moon
 require_cmd wasm-tools
-require_cmd wac
 
 INPUT=""
 OUTPUT=""
@@ -63,16 +62,10 @@ else
   echo "[compose] using cached adapter"
 fi
 
-# Step 2: Compile vibe source to a sync component, then compose with the
-# adapter via wac. We use wac rather than `vibe compile --compose-p3` because
-# the built-in mwac plug_components does not forward component type
-# definitions referenced by the socket's imports (issue #267) and produces
-# an invalid component for non-trivial adapter imports.
-PLUG="${OUTPUT%.wasm}.plug.wasm"
-echo "[compose] compiling $INPUT to plug component..."
-$VIBE compile --component-string-lift "$INPUT" -o "$PLUG" 2>/dev/null
-echo "[compose] composing plug into adapter via wac plug..."
-wac plug --plug "$PLUG" -o "$OUTPUT" "$ADAPTER"
+# Step 2: Compile vibe source and compose with the adapter via the built-in
+# mwac plug path (vibe compile --compose-p3).
+echo "[compose] compiling $INPUT and composing with adapter..."
+$VIBE compile --compose-p3 "$INPUT" --adapter "$ADAPTER" -o "$OUTPUT" 2>/dev/null
 
 # Step 3: Validate
 wasm-tools validate --features all "$OUTPUT"
--- a/scripts/test_wasi_p3_e2e.sh
+++ b/scripts/test_wasi_p3_e2e.sh
@@ -88,7 +88,6 @@ fi
 test_scalar_handler() {
   local name="scalar_hello"
   local src="$OUT_DIR/${name}.vibe"
-  local plug_component="$OUT_DIR/${name}.plug.wasm"
   local component="$OUT_DIR/${name}.p3.component.wasm"
 
   cat > "$src" << 'VIBE'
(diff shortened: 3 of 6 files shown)


mizchi:
Superseded by:
- #311 (Bump mwac 0.2.1) — merged
- #314 (Use `mwac.plug_components` in-process for `--compose-p3`, follow-up to #287) — merged

Closing as stale.

---

PR #348, opened 2026-04-27
codegen: __perform(Ctor(args)) function-call form mirrors keyword for…

mizchi:
…m (+1 test)

`perform(Ctor(args))` (the function-call form) was lowered as a plain `Call("__perform", [Ctor(args)])` whose codegen just compiled and threw the inner ctor object directly. The keyword form `perform Effect::Op(args)` instead stores the payload in `effect_perform_arg_global` and throws a string tag `"Effect::Op"` for the surrounding handle's catch to dispatch on.

Because the two paths disagreed, an inner `handle … with Ask` caught a tagged enum-object pointer while the catch-side dispatch compared against a string-literal pointer. They never matched, the handler retry loop ran to its 64-slot ceiling, and the `i64.store` to `resume_buf[count*8]` tripped a `memory access out of bounds` trap.

Symptoms in `examples/perform_handle.vibe`:

    test "perform+handle typed payload": failed (wasm OOB)
    test "perform+handle multi-layer":   failed (same root cause
                                                  + a separate
                                                  nested-handle
                                                  resume-state
                                                  issue, not fixed
                                                  here)

This patch makes the function-call form act like the keyword form when the inner expression is a ctor call AND a resume-handler is in scope:

1. Compile and store the ctor's payload (Unit / single arg / Tuple for arity ≥ 2) in `effect_perform_arg_global`.
2. Look up the qualified key `"Effect::Op"` from `ctx.effect_handled_ops` (populated during scan from in-scope handle arm patterns; ctor name is suffix-matched). Fall back to the bare ctor name when no enclosing handle owns the op.
3. Mirror `Expr::Perform`'s resume-read path: on retry iterations, read `resume_buf[idx++]` instead of throwing — without this the handler's outer retry loop would re-throw forever and overflow the 64-slot resume buffer (the original OOB trap).
4. Throw the qualified key so the catch-dispatch matches.

Result on examples/ suite:
  before: 720 total, 718 passed, 2 failed
  after:  720 total, 719 passed, 1 failed
  (`perform+handle typed payload` now passes; `multi-layer`
   still fails — that needs handler-local resume buffers or a resume-state stack, separate change)

Verified:
- moon test (all packages): 1259/1259
- examples/perform_handle.vibe: 2/3 (was 1/3)
- examples/effect*: all green (no regressions)
- Full examples/ suite: zero regressions

https://claude.ai/code/session_01CHCYdAQXmEBeDfDk9L2y1U

### DIFF
--- a/src/codegen/wasm_codegen_builtin_misc.mbt
+++ b/src/codegen/wasm_codegen_builtin_misc.mbt
@@ -85,13 +85,111 @@ fn compile_builtin_misc(
     }
     "__perform" => {
       let pos_args = extract_positional_exprs_with_arity(args, "__perform", 1)
-      if ctx.use_js_strings {
-        compile_expr(ctx, buf, pos_args[0])
-      } else {
-        compile_expr_i32(ctx, buf, pos_args[0])
+      // `perform(Ctor(args))` (function-call form) needs to behave like the
+      // keyword form `perform Effect::Op(args)` so the surrounding handle
+      // can dispatch on the ctor name as a string. Without this, the handler
+      // compares a thrown enum-object pointer to a literal ctor-name string
+      // and the dispatch never matches — the retry loop runs to OOB.
+      let inner = pos_args[0]
+      let ctor_payload : (String, Array[@core.Expr], @core.Span)? = match inner {
+        @core.Expr::Call(name~, args=ctor_args, span~, ..) =>
+          if @core.is_ctor_name(name) &&
+            ctx.need_effect_resume &&
+            ctx.effect_perform_arg_global >= 0 {
+            let exprs = extract_positional_exprs(ctor_args, name)
+            Some((name, exprs, span))
+          } else {
+            None
+          }
+        _ => None
       }
-      buf.push((8).to_byte()) // throw
-      write_u32_leb(buf, ctx.error_tag_index)
+      match ctor_payload {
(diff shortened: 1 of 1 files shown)


mizchi:
main で既に #349 (`codegen: per-effect resume globals for nested handle multi-layer`) によって本PRと同等のロジック（function-call form の `__perform(Ctor(args))` を keyword form と同じく qualified key throw / resume buffer 読み出しで処理する）が、より進んだ per-effect resume globals 実装で取り込まれています。

main 側の `src/codegen/wasm_codegen_builtin_misc.mbt` は既に：
- ctor 名 → qualified key の suffix lookup
- `effect_handled_ops` からの dispatch
- retry 時の resume buffer 読み出し分岐

を備えていて、さらに `get_effect_resume_globals(ctx, effect_name)` で effect ごとに独立した globals を引いてきます（multi-layer の clobber を防ぐ #349 の改善）。

一方この PR は古い singleton の `ctx.effect_resume_*_global` 前提で書かれているため、rebase 時の conflict を取り込むと #349 をリグレッションさせる形になります。`examples/perform_handle.vibe` も main 側に nested-twice / save-restore のテストが追加済みで、本PRブランチではそれらが消えています。

そのため本PRは superseded として close します。

merged on main: 854c9eb (#349)

---

Issue #544, opened 2026-06-14
codegen: `throw` unwinding through a non-matching effect handler traps (memory access out of bounds)


mizchi:
## Summary

When a `throw` (Error effect) unwinds **through a `handle` for a *different* effect** on its way to the matching `Error` handler, codegen produces wrong code that **traps** (`memory access out of bounds`) instead of letting the throw pass through the unrelated handler. Reproduces with **both** the host (`src/`) and selfhost (`vibe/compiler`) backends.

## Minimal repro

```vibe
effect Prof {
  Now -> Int
}
let work: () -> Int with { Error } = () -> {
  throw("boom")
  0
}
export let answer = () -> Int {
  handle {
    handle {
      work()
    } with Prof {
      Now => resume(0)
    }
  } with Error {
    Throw(_) => 42
  }
}
```

`work()` throws an `Error`; the inner `handle … with Prof` handles a *different* effect, so the throw should pass through it to the outer `Error` handler and yield `42`.

- host-compiled (`vibe compile --wasm`) → `RuntimeError: memory access out of bounds`
- selfhost-compiled → traps (no result)
- Expected: `42`

## Working variants (isolates the bug)

- Error handler **innermost** (throw caught before crossing the other handler) → works:
  ```vibe
  handle { handle { work() } with Error { Throw(_) => 42 } } with Prof { Now => resume(0) }   // => 42
  ```
  …and this still works when `work` additionally does `perform Prof::Now` (perform unwinds through the inner Error handler to the outer Prof handler fine).
- `throw`/`handle` with **only** the `Error` effect in scope (no intervening foreign handler), including across deep recursion / nested `with { Error }` calls → works (→ 42).

So the trigger is specifically **throw unwinding across a handler frame for an unrelated effect**.

## Impact

This made the selfhost compiler hang/crash on *every* compile error, because `cli_main` wrapped the pipeline with the `Profiler` handler inside the `Error` handler (#543). Worked around there by reordering the handlers (Error innermost), but the codegen bug itself remains and will bite any `throw` that crosses a foreign effect handler.

## Notes

- Likely the effect/exception lowering does not correctly thread the unwind past a non-matching handler frame (the trap address pattern suggests a bad continuation/exnref tag or stack/handler-table offset).
- Both backends affected → check the shared lowering strategy (host `src/codegen` effect/exnref handling and the selfhost `vibe/compiler/codegen` counterpart).

Found while root-causing #543 during the SIMD lexer-scan work (#536).

[...]

Issue #543, opened 2026-06-14
host-built (skip-opt) selfhost dist hangs on ALL compile errors — throw/Error not propagating to cli_main handler


mizchi:
## Summary

Compiling a program that uses the reserved keyword `test` as a binding/function name causes the selfhost compiler to **hang (infinite loop)** instead of emitting a parse error.

## Repro

```vibe
export let test = () -> Int { 42 }
```

Compile via the selfhost compiler (`cli_main`, entry `test`):
- `export let test = ...` → **hangs** (40s+ timeout, no output, no diagnostic), no wasm produced.
- `export let foo = ...` (identical body) → compiles in ~1s, wasm produced.

Reproduced cleanly in isolation (single process, files confirmed present, low system load), so it is not CPU contention.

## Notes

- `test` is a reserved keyword (`TTest`; `syntax/lexer.vibe:148` maps `"test"` → `TTest`, used for `test "name" { ... }` blocks).
- Reading the parser, `export let test` routes `TExport → parse_export_stmt → TLet → parse_let_stmt`, and `parse_let_stmt` / `parse_binding_name` both have `_ => throw("expected name after 'let'")` for a non-`TIdent` name — which *should* be a fast clean error. The observed behavior is an infinite loop instead, so the actual control flow diverges from that (the throw is not reached, or an inner loop spins). Worth pinning the exact non-advancing loop.
- `parse_program` (`syntax/parser.vibe:376`) sets `pos = next` from `parse_stmt` with **no guard against `next <= pos`**, so any sub-parser that returns without advancing will spin forever. A defensive guard there (throw on non-advancing) would convert this class of hang into a clean error regardless of the specific sub-parser.

## Impact

- A trivial source typo (using a keyword as a name) hangs the compiler with no diagnostic, rather than reporting a parse error.
- Also wasted significant debugging time when validating other work (programs/entries happened to be named `test`).

## Suggested fix

1. Add a non-advancing guard in `parse_program` (and `parse_module_body_with`): if `parse_stmt` returns `next <= pos` for a non-`TEof`/`TSemicolon` token, `throw` a parse error.
2. Ensure `let <keyword>` reports "expected name after 'let'" (and similar for other binding positions) rather than looping.

Found while implementing the SIMD lexer-scan builtins (#536).

https://claude.ai/code/session_01HcJ24wBXbRWZHE2RZjPtJi

mizchi:
[...]

PR #549, opened 2026-06-15
test(selfhost): pin cli_main Profiler/Error handler nesting

mizchi:
## What

Adds a regression test that pins the exact effect-handler nesting used by
`cli_main` in `vibe/cli/selfhost_entry.vibe`: **Profiler inner / Error outer**.

This nesting is load-bearing:
- `perform Profiler::NowUs` must be caught directly by the inner handler, not
  unwind through the unrelated `Error` handler — a resume that crosses an
  unrelated effect handler was miscompiled and corrupted the result (#543).
- `throw` must still unwind through the inner Profiler handler to the outer
  Error handler thanks to the dedicated exception tag (#544).

Flipping the two handlers makes `compile-lite --profile-tsv` wrongly exit
non-zero (the selfhost-perf-kpi regression). The existing compile-lite tests
handled only `Profiler` (with `Error`/`Fs` propagating out), so they never
exercised the two-handler interaction and missed the regression.

## Tests added (`vibe/cli/selfhost_test.vibe`)

A new `run_cli_main_nesting` helper mirrors `cli_main`'s nesting exactly, with:

- `cli_main nesting: profiled compile succeeds (exit 0)` — the regressed case:
  a successful `--profile-tsv` compile must return 0.
- `cli_main nesting: compile error reports exit 1 with profiling` — a compile
  error must unwind through the inner Profiler handler to the outer Error
  handler and report exit 1 (the throw crosses the Profiler handler), no hang.
- `cli_main nesting: compile error reports exit 1 without profiling` —
  companion case where the Profiler effect is never performed.

A cross-reference comment is added next to `cli_main` so the two stay in sync.

Test-only change (plus one comment in `selfhost_entry.vibe`); no behavior change.

---

### DIFF
--- a/vibe/cli/selfhost.vibe
+++ b/vibe/cli/selfhost.vibe
@@ -31,7 +31,7 @@ let own_host_string: (String) -> String = (value) -> {
   String::concat("", value)
 }
 
-let collect_cli_args: () -> Array[String] with { Env } = () -> {
+export let collect_cli_args: () -> Array[String] with { Env } = () -> {
   let out = Array::slice([
     ""
   ], 0, 0)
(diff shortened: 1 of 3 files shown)


mizchi:
Addressed the "exercise the real cli_main handler stack" P2 in `c3c9c4c`.

The handler nesting now lives in a single wrapper that both the production entry and the test go through:
- `selfhost_entry.vibe`: exported `run_cli_main_with_args(args)` holds the Profiler-inner / Error-outer nesting; `cli_main` is now just `run_cli_main_with_args(collect_cli_args())`.
- `selfhost.vibe`: added `selfhost_cli_dispatch_args(args)` (argv-driven command routing) so the wrapper can be driven with injected args; `selfhost_cli_main` routes through it.
- `selfhost_test.vibe`: the regression test calls `run_cli_main_with_args` directly, so flipping the order in `selfhost_entry.vibe` now fails the test.

Note on the suggested "share the entry wrapper via a thunk": passing the inner computation as an effect-annotated function parameter isn't an option — the selfhost compiler traps (memory OOB, no wasm) when compiling a HOF with an effect-row function-type param. The wrapper therefore takes a plain `Array[String]` and dispatches internally. `vibe test vibe/cli/selfhost_test.vibe` → ok (25/25).

---

Issue #550, opened 2026-06-15
parser: `perform` binds looser than binary operators (`perform X + perform Y` parses as `perform (X + perform Y)`)


mizchi:
## Summary

`perform` is parsed as a prefix operator that swallows the **whole** following expression, so it binds looser than binary operators:

```vibe
perform A::Op + perform B::Op
// parses as:  perform A::Op(... no, as ...)  ->  perform (A::Op + perform B::Op)
```

i.e. `perform X + perform Y` is grouped as `perform (X + perform Y)` instead of `(perform X) + (perform Y)`. This affects nullary effect ops as well as typed ones.

Noted as out-of-scope while landing the effect-system fixes in #547 (and on #546).

## Repro

```vibe
effect A { Op -> Int }
effect B { Op -> Int }
export let answer = () -> Int {
  handle {
    handle {
      perform A::Op + perform B::Op
    } with A { Op => resume(2) }
  } with B { Op => resume(3) }
}
```

Expected `5` (`2 + 3`); the current precedence groups the `+` inside the `perform A::Op` operand instead.

## Workaround

Parenthesize each `perform`: `(perform A::Op) + (perform B::Op)`.

## Fix direction

Give `perform` a precedence/operand grammar consistent with other prefix forms so it binds tighter than binary operators (its operand should be the effect-op application, not an arbitrary trailing expression). Add a parser test pinning `(perform X) + (perform Y)` grouping.

## Refs
- #547 (effect-system fixes; this quirk called out as out-of-scope)
- #546

mizchi:
Fixed in #551 (merged).

`parse_perform_primary` now parses `perform`'s operand at `min_prec` 10 (above the highest binop precedence), so `perform A::Op + perform B::Op` groups as `(perform A::Op) + (perform B::Op)`. Verified against a self-built stage2 (both the bare and parenthesized forms → 5), and stage1 == stage2 byte-identical confirms the compiler's own builtin `perform Env::…` usages are unaffected.

Closing as resolved.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #548, opened 2026-06-15
selfhost: multi-argument / tuple-payload effect ops not fully supported


mizchi:
Follow-up from Codex review on #547 (the #546 typed-payload effect-op parse fix). Two related gaps for effect ops with more than one payload value. Both are **pre-existing** (before #546, typed-payload ops crashed at parse, so this is not a regression), and the common single-typed-param case (`Op(Int) -> Int`) works correctly.

### 1. `perform` drops all but the first argument
`vibe/compiler/codegen/expr/compile_call.vibe` (the `perform` branch) emits only `op_args[0]` as the thrown exception payload:
```vibe
effect Random { NextInt(Int, Int) -> Int }
... perform Random::NextInt(lo, hi)   // only `lo` is thrown; `hi` is dropped
```
`compile_expr_tail6` also binds a single thrown value, so a handler can't receive the second payload. Either pack all op arguments into one payload (e.g. a tuple/heap value the handler destructures) or reject arity > 1 with a clear compile error instead of silently dropping.

### 2. A single tuple-typed param is read as multiple params
`parse_effect_stmt`'s `#546` workaround parses the op's parameter list via `parse_type_impl(..., 1)` and splits `TyTuple` into multiple params:
```vibe
effect E { Send((Int, Int)) -> Unit }   // intended: one param of type (Int, Int)
```
parses the same as `Send(Int, Int)` (two params), because `((Int, Int))` and `(Int, Int)` both yield `TyTuple([Int, Int])` at that point — the workaround can't distinguish "one tuple param" from "two params". (Effect-op param *types* appear largely unused downstream today — the checker yields a fresh var for `perform` and codegen uses the call args, not the declared types — so this is mostly latent, but it should be made faithful if/when those types are consumed.)

### Suggested direction
Implement multi-value effect payloads end-to-end: `perform` packs N op args into a single payload value, and `handle` arm patterns destructure it. That also subsumes the tuple-param case (a tuple param is just one value). Until then, consider rejecting arity > 1 to avoid silent miscompilation.

Refs: #547 review comments by `chatgpt-codex-connector`.

mizchi:
Partial: the **arity > 1 guard** (item 1's silent arg-drop) landed in #551 — a non-builtin effect op performed with more than one argument now fails with a clear compile error (`perform: effect operations with more than one argument are not supported yet (see #548); pack the values into a single argument`) instead of silently dropping the extra args. Builtin ops keep their existing multi-arg path.

[...]

PR #551, opened 2026-06-15
fix(selfhost): perform precedence (#550) + reject multi-arg effect ops (#548 guard)

mizchi:
Follow-up to the effect-system cluster (after #547/#549), implementing the two items agreed as the clean stopping point.

## #550 — `perform` precedence

`parse_perform_primary` parsed `perform`'s operand at `min_prec` 0, so it swallowed trailing binary operators:

```vibe
perform A::Op + perform B::Op   // parsed as: perform (A::Op + perform B::Op)
```

Parse the operand at `min_prec` 10 (above the highest binop precedence, 9 — see `binop_prec`) so `perform` takes only the effect-op application (primary + postfix):

```vibe
perform A::Op + perform B::Op   // now: (perform A::Op) + (perform B::Op)
```

This fixes a user-effect pattern that was previously **uncompilable** (the seed compiler can't compile `perform E::Get > 5`; the new parser yields the correct result). The compiler's own source only uses builtin `perform Env::…` forms, which parse identically under old and new precedence.

## #548 (guard) — reject multi-argument effect ops

A non-builtin effect op with arity > 1 had all but the first argument **silently dropped** at codegen (`compile_call.vibe`). Reject arity > 1 with a clear compile error instead. Builtin ops keep their existing multi-arg path (the `op_is_builtin` branch above the guard). Full multi-value payload support remains tracked in **#548**.

## Validation

Built a self-built generation from these sources (seed compiles → stage1 → stage2) and tested against **stage2** (the environment with #544 fixed, so error reporting is correct):

- `perform A::Op + perform B::Op` and the parenthesized form → **5**
- sanity: `40+2` → 42, single `perform` → 2, typed 1-arg `perform E::Op(9)` → 9
- arity-2 `perform E::Op(1,2)` → **rejected cleanly** (exit 1, no wasm, no hang), with and without an enclosing `handle`
- control: an `unknown name` error reports cleanly too
- **stage1 == stage2 byte-identical** — compiling the whole selfhost compiler with the old parser vs the new parser yields an identical compiler, proving #550 does not alter the compiler's own behavior

Note: a plain host-built dist (built by the `src/` host codegen, which intentionally still carries the #544 throw-through-foreign-handler bug) hangs on *any* compile error during validation — including `unknown name` — so error-path testing must use a self-built stage, which is what the selfhost CI gates use.

Scope: `vibe/compiler/` only (selfhost source of truth); `src/` untouched per project policy.

---


mizchi:
Fixed in `3e9122c` — regenerated `selfhost_sources_bundle.vibe` and `selfhost_cli_adapter_bundle.vibe` via `scripts/generate_selfhost_bundle.sh`, so the bundled sources now carry the new `parse_recur(tokens, pos + 1, 10)` (#550) and the arity-guard (#548). `scripts/check_selfhost_bundle_sync.sh` reports `ok` locally; CI is re-running on the new commit.

---

PR #552, opened 2026-06-15
feat(selfhost): multi-value effect-op payloads (#548)

### LINKED ISSUE #548: selfhost: multi-argument / tuple-payload effect ops not fully supported

mizchi:
Completes the #548 follow-up: effect ops with more than one argument now thread all values end-to-end, replacing the arity guard landed in #551.

## Change

- **`perform` (`compile_call.vibe`)** — pack N op args into a tuple (`ETuple`), one heap ref threaded as the single thrown payload, when arity > 1. Arity 0/1 keep the existing raw-value path (no wrapping), so single-arg and nullary effects are unchanged.
- **`handle` (`compile_expr_tail6.vibe`)** — when the arm ctor pattern has > 1 args, destructure the tuple payload and bind each element. Offsets match `compile_match`'s tuple destructure: `payload + i*8` (linear) / `payload + 8 + i*8` with the low tag bit cleared (RC). `_` skips an element; other nested sub-patterns are rejected with a clear error for now.

`resume` is unchanged — the resumed value stays single; only the `perform → handler` payload direction is multi-value.

Codegen keys off **use-site arity** (number of call args / pattern binds), not the declared op param types, so a single tuple-typed param performed as `Op((a, b))` (one call arg) binds to one pattern var — resolving #548 item 2 (the tuple-vs-multi-param ambiguity) for the runtime as well.

## Validation

Self-built generation from these sources (seed → stage1 → stage2), **stage1 == stage2 byte-identical**:

- `effect E { Op(Int, Int) -> Int }`, `perform E::Op(3, 4)` with `Op(x, y) => resume(x*10 + y)` → **34**
- 3-arg `Op(a, b, c)` / `perform E::Op(1, 2, 3)` → **123**
- single-arg `Op(Int)` / `perform E::Op(9)` with `Op(x)` → **10** (regression intact)
- #550 nested cross-handler `perform A::Op + perform B::Op` → **5** (no regression)

Bundles (`selfhost_sources_bundle.vibe`, `selfhost_cli_adapter_bundle.vibe`) regenerated; `check_selfhost_bundle_sync.sh` → ok.

Scope: `vibe/compiler/` only; `src/` untouched per policy.

Closes #548.

---


mizchi:
Thanks — both addressed:

**P1 (continuation slot, `compile_expr_tail6.vibe:93`)** — fixed in a1c222d. Correct catch: keying off pattern arity collided with the `Op(payload, k)` continuation-slot form (e.g. `fixtures/effect_handle_resume_in_block.vibe`'s `Question(q, k)` on a 1-arg op). The handler now keys off the op's **declared** arity, threaded into `CompileCtx` (`effect_op_keys` / `effect_op_arities`, built from `SEffectDef`). Tuple-destructure happens only when declared arity > 1; extra binds (continuation slots) and unknown ops take the single-payload path. Verified `Question(q, k)` → `12` again on a self-built stage2 (stage1 == stage2), alongside multi-arg → 34/123.

**P2 (RC payload drop, line 100)** — deferring. This is the pre-existing behavior of the effect-payload path under `VIBE_RC=1`: the single-value handler never `rc_drop`ped its `cargs[0]` payload either, so the multi-arg tuple is the same class of limitation (one extra spine), not a new regression. The default `vibe test` / `vibe build` backend is linear (bump-allocated, no `rc_drop`), so this only affects the opt-in RC mode. A Perceus-correct destructure-consume for synthetic effect payloads (dup heap fields on load, drop the spine) is worth a dedicated follow-up rather than a partial fix here.

---

Issue #678, opened 2026-06-28
codegen: bare-ident Int `==` mis-compares two distinct multiples of 2^31 (type-directed eq needed)


mizchi:
## Context

Follow-up to the integer-equality soundness fix on branch
`claude/vibe-lang-handoff-trait-iterator-eoomjq`
(commit `81aa598`, "integer `==`/`!=` is sound for values >= 2^32 (+ large-literal trap)").

That commit fixed the root cause of the previously-reported "2^61 literal leb128
overflow": the polymorphic `eq` builtin (`i64.eq(a,b) ? 1 : str_eq(a,b)`) falls
back to `str_eq`, which reads each operand as a string fat-pointer
`(ptr<<32)|len`. For two bit-different **integers** this misfires — two unequal
ints with a zero low word compare EQUAL, and a large low word reads OOB (which
also made literals >= 2^39 compile-trap, because `leb128_encode_s32`'s loop
terminates on `v == 0` / `v == -1`).

The fix routes any `==`/`!=` with a **syntactically-numeric** operand (`Int`/`Bool`
literal, bitwise/comparison/logical op, non-float arithmetic) to a direct
`i64.eq`, and bounds-guards the `str_eq` fall-back (`ptr >= 64`,
`ptr + len <= memory_size`). This covers the overwhelming majority of integer
comparisons and eliminates all compile traps.

## Residual hole

When **both** operands are *bare integer variables* (no syntactic hint that
they are numeric), the comparison still routes through the bounds-guarded `eq`,
and that fall-back cannot resolve the Int/String ambiguity for the `len == 0`
case:

```vibe
export let _start: () -> Int = () -> {
  let a = 1 << 40        // tagged value has low32 == 0, high32 == 512
  let b = 1 << 41        // tagged value has low32 == 0, high32 == 1024
  if a == b { 111 } else { 222 }   // BUG: returns 111 (claims equal)
}
```

Both tagged values look like two **empty strings** at different (in-range)
offsets, and two distinct empty strings legitimately compare equal — so the
fall-back has no sound way to tell them apart from two integers that are exact
multiples of 2^31.

Trigger conditions (all required), so this is narrow in practice:
- both operands are bare local idents (not literals / arithmetic / a `<<` etc.),
- both are exact multiples of 2^31 (tagged low 32 bits == 0),
- both high words land in `[64, memory_size >> 16]`.

## Proper fix

Type-directed equality codegen. Either:
[...]

Issue #672, opened 2026-06-28
codegen: `==` / `!=` is non-structural for aggregates (enum/struct/tuple) — silent miscompile, always false


mizchi:
## Summary

`==` (and `!=`) on any **aggregate** value — a nullary enum variant, a `struct … derive(Eq)`, or a tuple — compiles without error but **always returns `false`** (the codegen lowers `==` to a primitive i64/pointer compare; two distinct allocations are never pointer-equal). This is a silent miscompile: code that looks correct produces wrong results. Only primitive operands (`Int`, `String`) compare correctly.

Surfaced while building the selfhost unit-test runner (#531/#671): several allowlist-candidate tests compile and run but trap on a `==` assert.

## Repros (selfhost compiler, current `main` and the committed seed — pre-existing, not a regression)

```vibe
enum C { Red; Green; Blue }
export let _start: () -> Int = () -> { if Red == Red { 1 } else { 0 } }   // => 0, expected 1
```
```vibe
struct P { x: Int; y: Int } derive(Eq)
export let _start: () -> Int = () -> { if P::{ x: 1, y: 2 } == P::{ x: 1, y: 2 } { 1 } else { 0 } }   // => 0, expected 1
```
```vibe
export let _start: () -> Int = () -> { if (1, 2) == (1, 2) { 1 } else { 0 } }   // => 0, expected 1
export let _start: () -> Int = () -> { if (1, "b") == (1, "b") { 1 } else { 0 } } // => 0, expected 1
```

Primitives are fine: `"b" == "b"` ⇒ `1`, `1 == 1` ⇒ `1`. Inequality cases return `0` only by accident (everything aggregate is `false`), so `!=` is wrong too (`(1,2) != (1,2)` ⇒ `true`).

## Root cause

`==`/`!=` codegen (`vibe/compiler/codegen/expr/compile_expr*.vibe`) emits a scalar comparison. There is **no structural-equality path** for aggregates: nullary enum variants / structs / tuples are heap values, so the scalar compare is effectively pointer identity → distinct allocations never compare equal. `derive(Eq)` does not currently generate a structural `==` for the selfhost backend either.

(The retired MoonBit host presumably provided structural equality, which is why committed tests like `examples/for_in_test.vibe` — `assert(indexed[1] == (1, "b"))` — and `vibe/prelude/cmp_test.vibe` were written expecting it.)

## Impact

- Silent wrong results anywhere `==`/`!=` is used on a tuple/struct/enum.
[...]

PR #679, opened 2026-06-28
Integer == soundness (>= 2^32) + large-literal trap fix, on top of handoff parser/codegen fixes

mizchi:
## Summary

Lands the handoff parser/codegen fixes plus a new integer-equality soundness fix.

### New in this branch (`81aa598`)
**Integer `==`/`!=` is sound for values >= 2^32, and large integer literals no longer compile-trap.**

Root cause: the polymorphic `eq` builtin is `i64.eq(a,b) ? 1 : str_eq(a,b)`, and `str_eq` reads each operand as a string fat-pointer `(ptr<<32)|len`. For two bit-different **integers** the `i64.eq` fast path fails and the string fall-back runs: two unequal ints whose low 32 bits are both zero compare EQUAL, and an int with a large low word makes `str_eq`'s byte loop read out of bounds. The latter also made integer **literals >= 2^39 compile-trap** — `leb128_encode_s32`'s loop terminates on `v == 0` / `v == -1`, and those checks (routed through `eq`) misfire once the shifted value reaches 2^32. This subsumes the "2^61 literal leb128 overflow" symptom noted on #664.

Fix (two layers):
- `compile_expr.vibe`: a new `expr_is_intish` routes any `==`/`!=` with a syntactically-numeric operand (Int/Bool literal, bitwise/comparison/logical op, non-float arithmetic) to a direct `i64.eq`, correct over the full value range.
- `bodies_core_a1a1_eq.vibe`: `gen_eq_body` gates the `str_eq` fall-back on both operands looking like real strings (`ptr >= 64` and `ptr + len <= memory_size`, the #664 heuristic), so integer operands never reach the byte loop.

Genuine string comparison is unchanged (real strings satisfy the bounds; equal empty strings still compare equal). New `ieq` smoke in `selfhost_only_gate.sh` #39. Full gate green; stage2==stage3 fixpoint maintained.

Known residual tracked in #678: two *bare* integer variables that are both exact multiples of 2^31 with in-range high words still mis-compare (needs type-directed eq).

### Carried from `claude/trait-iterator-element-type-ke3z01`
- `__to_string` misreads large ints as strings → OOB crash (#664)
- struct field patterns in `match` (`S::{ x, y }`), linear + wasm-gc backends
- negative integer literal patterns (`-5`)
- string interpolation `\{...}` / `\(...)` parses a full expression
- `((A, B)) -> R` parsed as one tuple param
- qualified ctor `Type::Variant` + recursive/nested structural `==` (#672)

## Verification
- `bash scripts/selfhost_only_gate.sh` — green (smoke #39: `10/153/6/111/11111/321/3021/11111/4321/148/4321/4321/78/4321/4321`)
- stage2 == stage3 byte-identical fixpoint

https://claude.ai/code/session_01TUp6jZhxG1w1QBVuqKK9mf

---
_Generated by [Claude Code](https://claude.ai/code/session_01TUp6jZhxG1w1QBVuqKK9mf)_


mizchi:
Thanks — all three findings were valid and are now fixed in `1cc2abf` + `76571d4`:

**P2 — struct field-pattern offsets (`compile_match.vibe`)** ✅ `1cc2abf`
The matched `PStruct` name is now threaded through `bind_struct_pat` / `struct_field_index` (and the wasm-gc `struct_field_index_gc`), so offsets come from the matched struct's own field list instead of the first struct declaring that name. Verified: `struct A { x; y }`, `struct B { y; x }`, `match b { B::{ x, y } => x*10+y }` now yields `37` (was `73`).

**P1 — large Int identifiers in `eq` (`bodies_core_a1a1_eq.vibe`)** ✅ `76571d4`
Added type information at codegen via int-slot tracking (the dual of the existing float-slot machinery): a new `int_local_slots` on `CompileCtx`, populated by the `let`/`let mut` handlers when the value is statically integral. `expr_is_intish` now classifies a bare ident whose slot is int-bound, so `let a = 1<<40; let b = 1<<41; a == b` takes a direct `i64.eq` and returns not-equal.

**P1 — large Int values in `__to_string` (`compile_call.vibe`)** ✅ `76571d4`
`__to_string` now skips the string-pointer heuristic when the argument is a known Int (Int literal, bitwise/shift expr, or int-bound local), forcing the decimal path. `let n = 1<<40; "\{n}"` now yields `"1099511627776"` (was `""`).

Remaining narrow residual (tracked in #678): an Int that codegen can't classify — a function-call return or an untracked param/capture — still uses the bounds-guarded `eq`. Fully closing it needs the checker's types plumbed into codegen. Full `selfhost_only_gate` is green and the stage2==stage3 fixpoint holds; the `ieq` smoke now covers bare-ident eq + large-int interpolation.

---

mizchi:
Correction to my previous comment (after CI feedback in `265dd8d`):

The bare-ident int-slot tracking I'd added for the #678 residual was **unsound** and is reverted. Slots are reused across scopes (match-arm / for-loop / lambda binders) without updating the per-slot "is int" log, so a reused slot holding a *string* was misread as an Int → `wasm_parser_test` / `wat_parser_test` trapped in `Env::get`. CI caught it; thanks.

Also fixed a regression the same commit introduced: the `eq` `str_eq`-fallback guard required `ptr >= 64`, which wrongly rejected a genuine **empty string** with a low pointer (a 0-length custom-section name), breaking string-keyed lookups. The guard is now just `ptr + len <= memory_size` (unsigned high-word shift), which prevents OOB without rejecting low-pointer empty strings.

What stands (all sound, full gate + unit-test runner 67/67 + fixpoint green):
- `==`/`!=` for any **syntactically-numeric** operand → direct `i64.eq` (literal, bitwise/shift/mod, comparison/logical, non-float arithmetic).
- large integer literals no longer compile-trap.
- `__to_string` of a known-int **literal or bitwise/shift expr** (`"\{1 << 40}"`) → decimal, no longer `""`.
- P2 struct-pattern offsets use the matched struct's name.

**#678 stays open** with its original narrow residual: comparing two *bare integer variables* that are exact multiples of 2^31 still mis-compares (and `"\{n}"` for such a bare-ident large int). Soundly closing it needs the checker's types plumbed into codegen (a per-slot heuristic is unsound, as shown). The `__to_string` P1 for a bare-ident large int likewise remains; the literal/bitwise forms are fixed.

---

Issue #885, opened 2026-07-14
checker: effect-row variable propagation is unenforced (deferred from #838)


mizchi:
Split off from #838 (now closed as docs-fixed).

`check_perform_effects_expr_tx` (checker/checker_effects.vibe, from #626) explicitly excludes effect ROW VARIABLES from the transitive call-graph check (`label_is_effect_var` guard — "Effect-variable rows impose nothing"). The call-graph map (`fn_names`/`fn_effs`) also only tracks named top-level bindings, not function-typed lambda parameters. So a wrapper that calls an effect-row-polymorphic callback without declaring `with { e }` compiles clean when it should be rejected.

A naive fix (dropping the row-variable exclusion) produces false positives on legitimate pure instantiations — verified against `fixtures/typecheck/generic_effect_matrix_ok_passthrough_pure.vibe`, which would incorrectly start failing. A sound fix needs call-site effect-row unification, not just label matching.

docs/vibe.md and docs/language-tour/effects.md currently mark the corresponding `// error:` examples as a known, unenforced gap (`vibe skip`) pending this fix.

## Acceptance criteria
- `check_perform_effects_expr_tx` (or equivalent) catches an effect-row-polymorphic callback invoked without the required `with { e }` declaration.
- `fixtures/typecheck/generic_effect_matrix_ok_passthrough_pure.vibe` and other legitimate pure-instantiation cases still pass.
- docs/vibe.md + docs/language-tour/effects.md `vibe skip` markers on the affected examples are removed once enforced.

Refs #838, #626.

mizchi:
Fixed and merged in PR #896 (commit `56725aca`, with a Codex-review follow-up fix in `449e7e10` for a shadowing false-positive the same PR introduced). Both acceptance criteria confirmed: a callback invoked without its declaring function propagating the effect is now rejected, and `generic_effect_matrix_ok_passthrough_pure.vibe`/other legitimate pure-instantiation fixtures still pass with no false positives — verified via direct A/B diff of `vibe diagnostics` output between unmodified `HEAD` and the fix. `vibe skip` markers removed from the now-enforced doc examples in `docs/vibe.md`/`docs/language-tour/effects.md`.

Scope note (not a gap in this issue's acceptance criteria, but worth flagging for anyone picking up related work): call-site effect-row unification for a downstream caller instantiating a row variable (e.g. `apply(risky, 1)` needing `{Error}`) is still unenforced — that's a separate, harder problem than what this issue asked for.

---
[...]

PR #896, opened 2026-07-15
#885: enforce effect-row propagation through callback parameters

mizchi:
## Summary

`check_perform_effects_expr_tx`'s call-graph map (`fn_names`/`fn_effs`) was built only from top-level `SLet`/`SLetMut` bindings, so invoking a function-typed PARAMETER directly (a callback, e.g. `f: (T) -> T with { e }`) was invisible to the transitive effect check — no lookup ever matched the parameter's name, so a wrapper that calls its own callback without declaring `with { e }` compiled clean when it should be rejected.

Adds a parallel "overlay" (`ov_names`/`ov_effs`) tracking function-typed parameters currently in scope, threaded through the existing explicit call-stack traversal and extended whenever the walk enters an `EFn`. Unlike the existing named-top-level-function check, this overlay does **not** exempt effect row variables: the callback's declared row and the enclosing function's own declared row are the identical token in the same signature (no cross-call-site instantiation involved), so a literal-name membership check is sound here, not the naive shortcut that previously broke `fixtures/typecheck/generic_effect_matrix_ok_passthrough_pure.vibe` when tried globally.

Also handles the separate-annotation binding form (`let apply: [T](f: (T) -> T with {e}, ...) -> ... = (f, x) -> {...}`), where the lambda's own AST carries no per-parameter types — `merge_param_types_with_sig` fills them in positionally from the binding's own type signature.

## Scope

Only catches a callback invoked without its **own** declaring function propagating the effect (#885's stated acceptance criteria). Resolving what a row variable is instantiated to at a downstream **call site** (e.g. `apply(risky, 1)` needing `{Error}`) still requires real call-site effect-row unification and is deliberately left alone — verified `generic_effect_matrix_fail_missing_effect_at_caller.vibe` and `generic_effect_missing_caller_effect.vibe` (which test that case) remain unchanged.

Un-skips the corresponding doctest examples in `docs/vibe.md` and `docs/language-tour/effects.vibe.md` now that the checker enforces them for real; the "should reject" half of each pair stays `vibe skip` since the doctest harness has no expect-compile-failure mode (pre-existing limitation, unrelated to this fix).

## Verification

- `bash scripts/generate_bundle.sh` (`VIBE_REGEN_MODULE_SOURCE=1`) → regenerated cleanly
- `bash scripts/check_bundle_sync.sh` → ok
- `bash scripts/check_module_source_sync.sh` → ok
- `bash scripts/compiler_gate.sh` → 44/44 sections ok, cold moon-free bootstrap chain
- `bash scripts/unit_test_runner.sh` → 417/417 allowlisted files passed (fresh cold rebuild), 0 regressions
- Both acceptance criteria confirmed via direct A/B diff of `vibe diagnostics` output between a compiler built from unmodified `HEAD` and one built from this fix, across the full generic-effect-matrix fixture set plus the doc examples.

## Test plan

- [x] `bash scripts/compiler_gate.sh` — 44/44
- [x] `bash scripts/unit_test_runner.sh` — 417/417, no regressions
- [x] Bundle/module-source sync checks pass
- [x] Both acceptance criteria (new rejection + no false positives) verified via A/B diagnostics diff

https://claude.ai/code/session_019tXS9Nb2wA5YErMjibod1H

---
_Generated by [Claude Code](https://claude.ai/code/session_019tXS9Nb2wA5YErMjibod1H)_


mizchi:
Re: the shadowing review comment on `checker_effects.vibe` — confirmed real, reproduced your exact scenario as a unit test and it did fail before this fix. Fixed in 449e7e10:

- `ELet`/`ELetRec`/`ELetMut` now shadow any overlay entry matching their own bound name before pushing the body's stack frame (`ELetRec` shadows both `val` and `body` since it allows self-recursion; `ELet`/`ELetMut` only shadow `body`).
- Also found and fixed the same class of bug on the `EFn` path: a nested lambda parameter reusing an outer callback's name with a *different* type (e.g. plain `Int`, or an effect-free function type) previously wasn't shadowed at all, since only parameters that themselves became new overlay entries were considered.

Added 3 permanent regression tests directly exercising `check_perform_effects_expr_tx` (your exact scenario, the original #885 case to confirm no regression, and the EFn-parameter variant) — all passing, plus the full `compiler_gate.sh` (44/44) and `unit_test_runner.sh` (420/420) triad.

---

PR #932, opened 2026-07-17
Fix #930: Stdout/Stderr::write_char writes doubled tagged byte under …

### LINKED ISSUE #930: codegen: Stdout::write_char writes the raw tagged value (42 prints as "hd")
Found while debugging PR #927's REPL CI failure. Reproduced identically under BOTH the node host runner and vibewt (pinned seed and fresh stage2):

```vibe
fn put_str(s: String) -> Unit with { Stdout } {
  let n = String::length(s)
  let mut i = 0
  while i < n {
    Stdout::write_char(String::char_code_at(s, i))
    i = i + 1
  }
  Stdout::write_char(10)
}

export let __repl_main__: () -> Unit with { Stdout } = () -> {
  let v = 41 + 1
  put_str("\{v}")   // expected "42\n"
}
```

Output: `hd` — i.e. bytes 104/100 instead of 52/50. Those are exactly the 62-bit-tagged representations (`52<<1=104`, `50<<1=100`, ADR-0055): the value reaches the write untagged-incorrectly — either the host import writes the raw tagged i64 byte, or the builtin body fails to untag before the write. The trailing `write_char(10)` similarly emits 20 (DC4) instead of a newline.

Consequence: odd byte values are unreachable through `Stdout::write_char`, so it cannot print arbitrary text at all. Together with #929 (println/print missing FS-mode lowering) this leaves the FS-compile lane with no working string-print primitive — the PR #927 REPL had to fall back to the Int entry-return print convention.

Fix is likely a single untag (`i64.shr_s 1`) in the builtin body or the host-import marshalling for `Stdout::write_char` / `Stdout::write_stream` (check both).

https://claude.ai/code/session_019tXS9Nb2wA5YErMjibod1H

mizchi:
…RC codegen

Under the default Perceus-RC linear backend (enable_rc), the char-code argument to Stdout::write_char / Stderr::write_char reached the raw-ABI host import still guest-tagged (n<<1), since these two host imports were missing from the untag-before-call shim group that every other raw-int-arg host import (Env::args_get, vibe_sh_capture_exit_code_raw, etc.) already has. The host wrote the tagged value directly as the byte, so e.g. write_char(52) ('4') printed 'h' (104) and write_char(50) ('2') printed 'd' (100) -- printing "42" as "hd".

Add the same untag-before-call / push-Unit-after-call shim used by the neighboring #705/#865 host-import groups. Regenerated the committed selfhost bundles (compiler_sources_bundle.vibe, cli_adapter_bundle.vibe, cli_adapter_module_source.vibe) and codegen fingerprint so the fix is actually reachable by the self-hosted build.

Verified directly: compiled and ran the exact repro from #930 through a freshly self-built stage2 compiler; output changed from "hd" to "42".


mizchi:
この修正は #933 (e5800e08) に包含される形で main にマージ済みです — 同じ `Stdout/Stderr::write_char` untag shim に加え、`Stdin::read_stream` の shim、`println`/`print` の linear FS-lane lowering(両 `VIBE_RC` lane、source shadow 優先)、gate 27f による両 lane の出力固定まで入っています。コンフリクトは同一箇所の重複によるものなので、こちらを close します(#930 も close 済み)。

---

PR #1202, opened 2026-07-29
Fix #1199 bug 2: dot-call sugar on [T: Hash]-bounded struct methods

### LINKED ISSUE #1199: Trait-dictionary-passing generics ([T: Hash]) break codegen combined with Array::get or dot-call
Found while implementing generic `Set[T]` for #1140 (naming consistency / ADR-0082). Two distinct, minimal, reproducible codegen bugs in the `[K: Trait]`-bounded generic / trait-dictionary-passing machinery (`desugar_trait_dict.vibe`). Both manifest as a wasm validation failure at load time (`Invalid input WebAssembly code ... type mismatch: expected i64 but nothing on stack`), i.e. the compiler emits structurally invalid wasm — not a checker rejection.

## Bug 1 — `Array::get` (or `arr[i]`) result fed into a `[K: Hash]`-bounded generic call

Minimal repro (single file, no Set/Map/Array-of-Set involved):

```vibe
export trait Hash {
  hash_key(Self) -> String
}

impl Hash for Int {
  hash_key(self) -> String {
    __to_string(self)
  }
}

fn map_key_to_string[K: Hash](key: K) -> String {
  K::hash_key(key)
}

fn one_key[T: Hash](values: Array[T]) -> String {
  let v = Array::get(values, 0)
  map_key_to_string(v)
}

fn main() -> Int {
  String::length(one_key([1, 2, 3]))
}
```

Compiles (exit 0) but the produced wasm fails to load:
```
Invalid input WebAssembly code at offset 2819: type mismatch: expected i64 but nothing on stack
```

- No loop/recursion needed — a single `Array::get` followed by a single Hash-bounded call reproduces it.
- Using index syntax (`values[0]`) instead of `Array::get(values, 0)` reproduces identically.
- Reproduces identically in both single-file and FS-compile (multi-file import) modes.
- Does NOT reproduce when the Hash-bounded call operates on a value that's a direct function parameter (not obtained via `Array::get`) — e.g. `fn f[T: Hash](v: T) -> String { map_key_to_string(v) }` compiles and runs fine, including inside a `while` loop that calls it repeatedly on the same parameter.
- Tried several workarounds that all reproduce identically: manual `while` loop, `let rec go = ...` recursive closure, routing through a generic HOF (`fold`), routing through an intermediate non-generic-looking helper function, using `MapBuilder` instead of persistent `Map`.
- Workaround that DOES avoid it: use the type-erased `__to_string(v)` builtin instead of a `[K: Hash]`-bounded call. This is the pre-existing trick `lib/@vibe/core/set.vibe`'s `StringSet::add_by`/`from_array_by` already uses (`set_value_to_string[T](value: T) -> String { __to_string(value) }`) — for every type currently `impl Hash` (Int/Float/Double/Bool/Char/String) this produces an identical string to `Hash::hash_key`, so it's a safe substitute today, just not a general fix.

Likely area: `desugar_trait_dict.vibe`'s witness/dictionary synthesis or the RC (Perceus) pass, specifically for a value whose origin is the generic builtin `Array::get`/`__index` rather than a plain local/parameter.

## Bug 2 — dot-call sugar (#641/#736) on a `[K: Hash]`-bounded method breaks codegen

Separate from bug 1. Given (from `lib/@vibe/core/set.vibe`, the new generic `Set[T]` added for #1140):

```vibe
export fn Set::add[T: Hash](s: Set[T], value: T) -> Set[T] {
  Set::{ entries: Map::set(s.entries, map_key_to_string(value), value) }
}
```

Calling it via the **explicit qualified form** works correctly:
```vibe
let s1 = Set::add(s0, 1)   // OK, compiles and runs correctly
```

but calling the exact same function via **dot-call sugar** (which resolves to the identical qualified call per #641/#736's `desugar_trait_dict.vibe` `ECall(EDot(...))` rewrite) breaks codegen:
```vibe
let s1 = s0.add(1)   // compiles (exit 0), but the wasm fails to load:
                      // "type mismatch: expected i64 but nothing on stack"
```

Isolated with a small sequence (`Set::new()` -> `.is_empty()` -> `.size()` -> `.add(1)` -> `.contains(1)` -> `.contains(2)` -> `.size()`): the identical sequence written with explicit `Set::method(...)` calls throughout compiles and runs correctly (verified output matches expectations); rewriting only `add`/`contains` to dot-call form (`s0.add(1)`, `s1.contains(1)`) reproduces the wasm load failure. The non-Hash-bounded methods on the same type (`size`, `is_empty`, `to_array`, `union`, `intersect`, `difference` — no trait bound) work fine via dot-call.

Likely area: the EDot-to-qualified-call rewrite in `desugar_trait_dict.vibe` runs at a pass ordering relative to `[K: Hash]` witness-dictionary synthesis that leaves the dictionary threading incomplete/incorrect for the rewritten call site specifically (vs. a call written in qualified form from the start, which the checker/dict-passing sees as such from the beginning).

## Current workaround in the codebase

`lib/@vibe/core/set.vibe`'s new `Set[T]` (added alongside `StringSet` for #1140/ADR-0082) avoids both bugs today by:
- Using `__to_string` instead of `Hash`/`map_key_to_string` in `Set::from_array` (bug 1).
- Its own test file (`lib/@vibe/core/set_generic_test.vibe`) calls the Hash-bounded methods (`add`/`contains`/`remove`/`from_array`) via explicit `Set::method(...)` form, not dot-call sugar, and documents why in a comment (bug 2).

These workarounds are correct for `Set[T]`'s current scope but paper over a real gap in the trait-dictionary-passing system that will resurface for any other `[K: Hash]`- (or other trait-)bounded generic code that either (a) extracts array elements generically before a trait-dictionary call, or (b) is called via dot-call sugar.

mizchi:
## Summary

Fixes bug 2 of #1199. Bug 1 (`Array::get` result fed into a `[K: Hash]`-bounded call) remains open — root cause traced deeper during this fix (see below) and left as a documented follow-up rather than rushed.

**Root cause of bug 2**: `desugar_trait_dict.vibe`'s EDot-derived qualified-call rewrite (the `as_recv_method` arm — what a dot-called `s0.add(1)` resolves through on its way to becoming `Set::add(s0, 1)`) built the resulting call directly, without checking whether the resolved qualified name is itself a trait-bounded generic needing a dictionary argument. The sibling `EIdent` arm (the literal `Type::method(...)` spelling written by hand) already does this via `lookup_gen`/`synth_dicts`. A method like `Set::add[T: Hash](s: Set[T], value: T)` has its own definition rewritten by `thread_dict_params` to expect a leading witness-dictionary parameter — but a dot-called `s0.add(1)` never supplied one, an arity mismatch between the (separately) rewritten callee and the call site that corrupted the wasm stack shape at load time (`type mismatch: expected i64 but nothing on stack`). The explicit `Set::add(s0, 1)` spelling always worked since it went through the `EIdent` arm's existing dict synthesis — this asymmetry between the two call spellings is what made it easy to miss.

**Fix**: route the `as_recv_method` arm through the same `lookup_gen`/`synth_dicts` step, synthesizing dicts from the ORIGINAL (unrewritten) `[obj, ...args]` — the same argument order a literal qualified call would have used.

**Bug 1's deeper root cause** (found while investigating, not fixed here): `build_gens`'s `find_recv_index` only recognizes a bare-`T`-typed value parameter as a threadable receiver. A type parameter used only inside a container type argument (e.g. `values: Array[T]`) gets ZERO dictionary threading on the enclosing function at all — it's not even attempted, so the function has no witness parameter to forward when its body later calls a `[K: Hash]`-bounded function on an array element. Properly fixing this needs new capability in `find_recv_index`/`synth_dicts` (recognizing container-nested type parameters, and inferring the concrete element type from call-site literals/annotations) — a larger, riskier change to the generics substrate than this bug fix, so it's left open.

## Commit

`codegen: fix dot-call sugar on [T: Hash]-bounded struct methods (#1199 bug 2)`

## Test plan

- [x] Verified against #1199's exact minimal repro (`Set::new` → `is_empty` → `size` → `add` → `contains` → `contains` → `size`, all via dot-call) — now compiles and runs correctly, output matches the explicit-call form
- [x] `fixtures/dot_call_hash_bound_method_test.vibe` (new) — isolated repro independent of Set/Map, 4/4 tests
- [x] `lib/@vibe/core/set_generic_test.vibe`'s new `set_generic_dot_call_hash_bound` test — chained dot-calls on `Set[Int]`
- [x] `scripts/unit_test_runner.sh` (full, not just `compiler_gate.sh`) — 464/464 allowlisted files pass
- [x] `scripts/lint_method_style_naming.sh` — clean, 0 new violations
- [x] `scripts/generations.sh build --stage3` — stage2==stage3 fixpoint
- [x] `scripts/compiler_gate.sh` — full gate green (67/67 steps, golden lang-review 10/10)

## References

- #1199
- #1140 (where the underlying `Set[T]` that surfaced this bug was added)

---


mizchi:
Superseded by #1201 (merged as `d58dd57`), which fixed both bug 1 and bug 2 of #1199 in full — including the deeper `find_recv_index`/array-element-receiver threading gap this PR's description flagged as a separate, larger follow-up. #1199 is now closed as completed.

Closing this PR without merging to avoid conflicting with the more complete fix already on `main`.

---

Issue #1129, opened 2026-07-27
checker/vibe diagnostics: detect unused imports


mizchi:
## Motivation

Found a real instance while doing package-reorg work: `lib/@vibe/compiler/compiler.vibe` had
`import @vibe/shell @shell` (an aliased package import) that was never referenced anywhere
in the file — no `@shell::...` call site at all. It went unnoticed because nothing in the
compiler (checker, `vibe diagnostics`, `compiler_gate.sh`) flags unused imports; it was only
found by manual `grep` while relocating the package (see PR #1127-adjacent follow-up commit
on `claude/tail-recursion-optimization-oyfdbd`). Removing it was verified safe via a full
bundle regen + stage2/stage3 fixpoint + `compiler_gate.sh` + unit-test battery run.

## Request

Add unused-import detection to the checker, surfaced through `vibe diagnostics` (and/or a
future `vibe check`-style command). Scope to consider:

- Unused aliased package imports (`import @scope/pkg @alias` where `@alias` is never used).
- Unused named imports (`import ./file.vibe { foo, bar }` where `foo`/`bar` are never
  referenced) — including `type X` and qualified `Type::method` import list entries.
- Should probably be a warning-level diagnostic (not a hard error) initially, since some
  imports exist purely to merge a package's exports into a whole-program compilation (see
  the "aliases" pattern in `lib/@vibe/compiler/compiler.vibe`) without a local call site —
  need to make sure the checker doesn't false-positive on that pattern if it's still in use
  elsewhere, or document it as legitimately flagged so callers switch to a clearer mechanism.

## Reference

- Package reorg session: moved `lib/@vibe/shell` → `lib/@vibex/shell`, `lib/@vibe/io` →
  `lib/@vibe/console`, flattened `lib/@vibex/wasm/*` → `lib/@vibex/wasm_*`, etc.
- `docs/adding-modules.md` / checker source: `lib/@vibe/compiler/checker/`

PR #1205, opened 2026-07-29
#1129 follow-up: fix Codex review findings on unused-import detection

mizchi:
## Summary

Follow-up to #1204 (merged as 7fd0f2b before Codex's review comments landed). Fixes three P2 findings from that review, all in the unused-import detector added there:

- **Shadowing false-negative**: `collect_referenced_names_expr` now tracks lexical scope (function params, `let`/`let rec`/`let mut`, `for`/`loop` vars, `match`/`handle` pattern binds — each pushed/truncated around the scope it covers). A local binder with the same name as an imported value no longer shadow-suppresses that import's own unused warning.
- **ERecord false-positive**: `Type::{ .. }` struct-literal construction now counts `Type` as referenced (previously only its field values and explicit type-argument list were walked, missing the constructor name itself) — fixes a false-positive unused-import warning on a type used only via struct-literal syntax.
- **Missing location on unused-import warnings**: they now carry a `line L:C:` prefix, via a new `locate_by_at_marker` heuristic locator (for `@alias`-shaped names, which aren't plain identifiers) plus two new marker branches (`"imported name '"`, `"import alias '"`) in the existing first-occurrence heuristic in `typecheck_fs.vibe`. No AST/parser change needed — an *unused* name by definition has no other occurrence in the file, so the existing first-occurrence search unambiguously finds the import statement itself.

## Test plan

- [x] `bash scripts/compiler_gate.sh` — 67/67, stage2==stage3 fixpoint, golden lang-review 10/10.
- [x] `bash scripts/unit_test_runner.sh` — 464/464 allowlisted files pass.
- [x] `bash scripts/vibe_fmt.sh --check` on all touched files — clean.
- [x] New unit tests added to `lib/@vibe/compiler/tests/checker_unused_import_and_discard_test.vibe`: shadowed-param regression, ERecord-reference regression, and two `locate_type_error` location tests (named import + alias).
- [x] Manual end-to-end smoke test with a fresh stage2 build and `VIBE_DIAGNOSTICS=1`, confirming an import shadowed by a same-named function parameter is still correctly flagged, with a precise location:
  ```
  line 2:12: warning: imported name 'shadow_target' from './b.vibe' is never used in this file
  line 2:27: warning: imported name 'unused_fn' from './b.vibe' is never used in this file
  line 5:20: warning: import alias '@shell' for '@vibe/shell' is never used in this file (referenced elsewhere in a whole-program merge? consider a named import instead)
  line 22:3: warning: expression result of type `Int` is unused -- bind it (e.g. `let _ = ...`) if this is intentional
  ```

---


mizchi:
Pushed a follow-up commit (`e513f1b`) addressing both new Codex review comments:

- `EForIn`'s optional index name (`for x, i in items { .. i .. }`) is now tracked as a shadow-scope binder alongside the item name — previously only `x` was pushed, so an import shadowed solely by the index `i` was silently treated as used.
- Unused-import warnings now anchor their location on the import's own path first (embedded in the same message), then search for the name/alias only from that anchor forward — a plain first-occurrence search could land on an earlier, unrelated occurrence (e.g. a comment mentioning the same identifier before the import). Verified with a regression test reproducing exactly that scenario.

`compiler_gate.sh` (67/67) and `unit_test_runner.sh` (464/464) both green; added regression tests for both fixes plus an end-to-end smoke test confirming the comment-collision case now locates correctly.

---

Issue #1185, opened 2026-07-28
Double `==` gives wrong result comparing two explicitly `: Double`-annotated let-bindings


mizchi:
## Summary

`a == b` for two `Double` values is wrong specifically when BOTH `a` and `b` come from separate `let` bindings carrying an **explicit `: Double` type annotation** — even though the underlying float values are identical. Comparing an annotated variable against an inline literal, or against an unannotated `let`, works fine; only two independently-annotated bindings trigger it.

## Repro

```vibe
import @vibe/prelude { stdout_write }

fn show(label: String, v: Bool) -> Unit with { Stdout } {
  stdout_write(label + "=" + (if v { "true" } else { "false" }) + "\n")
}

fn main with { Stdout } {
  let a: Double = 1.0
  let b: Double = 1.0
  show("annotated a==b", a == b)                 // false (WRONG -- should be true)

  let c: Double = 1.0
  show("annotated c == inline 1.0", c == 1.0)     // true  (correct)

  let d = 1.0
  show("annotated a == unannotated d", a == d)    // true  (correct)
}
```

Run via `bash scripts/vibe_run.sh` against `bootstrap/seed/compiler.wasm` (linear backend).

## Impact

Found while porting `scripts/wasm_feature_levels.py` to `.vibex`: `Json::number(j) -> Double` (from `@vibe/json`) returns a value through this same buggy path, so comparing two `Json::number(...)` results with `==` also silently returns `false` for equal numbers — confirmed with the exact same symptom (`Json::number(Json::field(j1,"a")) == Json::number(Json::field(j2,"a")))` for two structurally-identical parsed JSON `1`s gives `false`).

Any user code doing `Double` equality where both operands go through an explicit `: Double` annotation (very common for numeric struct/record fields, JSON number accessors, parsed config values, etc.) silently gets wrong results — no error, no crash.

## Workaround

Difference-based comparison instead of `==` (subtraction does not appear to hit the same bug — confirmed the workaround gives correct results for both equal and unequal inputs):

```vibe
fn double_eq(a: Double, b: Double) -> Bool {
  let d = a - b
  d < 0.0000001 && d > (0.0 - 0.0000001)
}
```

## Possible relation to other findings this session

Two other real comparison-operator bugs were found and filed while doing the same porting work:
- #1184: `String </>` fall back to length-based comparison when an operand's value originated from `Array::get`.

[...]

Issue #1184, opened 2026-07-28
String `<`/`>` give length-based (not lexicographic) results when an operand comes from Array::get


mizchi:
## Summary

`String < String` / `String > String` compare correctly (lexicographically) when both operands are literal expressions, but silently fall back to a length-based comparison when at least one operand's value originated from `Array::get` (even after being re-bound to a plain `let`) — reproduced against `bootstrap/seed/compiler.wasm` via `scripts/vibe_run.sh` (linear backend).

## Repro

```vibe
import @vibe/prelude { stdout_write }

fn show(label: String, v: Bool) -> Unit with { Stdout } {
  stdout_write(label + "=" + (if v { "true" } else { "false" }) + "\n")
}

fn main with { Stdout } {
  // lexicographically "zz" > "aaa" ('z' > 'a' at the first char); by length "zz"(2) < "aaa"(3).
  show("literal zz>aaa", "zz" > "aaa")                          // true  (correct)

  let arr = ["zz", "aaa"]
  show("array-get zz>aaa", Array::get(arr, 0) > Array::get(arr, 1))   // false (WRONG)

  let a0 = Array::get(arr, 0)
  let a1 = Array::get(arr, 1)
  show("array-get-then-let zz>aaa", a0 > a1)                     // false (WRONG -- re-binding to `let` doesn't fix it)
end
```

Output:
```
literal zz>aaa=true
array-get zz>aaa=false
array-get-then-let zz>aaa=false
```

`false` is length-based (`len("zz")=2 < len("aaa")=3` would make `"zz" > "aaa"` false), not lexicographic (`"zz" > "aaa"` should be `true` since `'z' > 'a'`).

Same-length operands from `Array::get` (e.g. `"zz"` vs `"aa"`) compare correctly — the bug only shows once the two strings differ in length AND originate from an array, i.e. exactly the scenario a length-first comparison would get wrong but a proper lexicographic one wouldn't.

## Impact

Any user code that sorts or orders `Array[String]` values with `<`/`>` (insertion sort, binary search, manual comparators, etc.) silently gets wrong results whenever compared strings differ in length — no error, no crash, just wrong ordering. Found while porting `scripts/wasm_feature_levels.py` to `.vibex` (needed a simple string sort — worked around by writing a manual `String::char_code_at`-based comparator instead of `<`/`>`, confirmed correct for the same repro inputs).

[...]

PR #1207, opened 2026-07-29
Fix #1153/#1185: annotated Double/Float let-bindings misclassified for RC float boxing

### LINKED ISSUE #1153: Double show/to_string prints garbage (both \{expr} interpolation and to_string(d))
Found while migrating `docs/tutorial/` to `.vibe.md` (#1142) and trying to print a `Double` value for a demo.

## Repro

```vibe
import @vibe/prelude { stdout_write }

let _start: () -> Unit with { Stdout } = () -> {
  let d: Double = 3.14
  stdout_write("d = \{d}\n")
}
```

Expected: `d = 3.14` (or some reasonable decimal rendering).
Actual: `d = 60` (varies across runs/rebuilds — also saw `68`, `92`, ... never a plausible decimal rendering of 3.14).

Same result via the explicit generic path, so this isn't specific to string-interpolation codegen:

```vibe
import @vibe/prelude { stdout_write, to_string }

let _start: () -> Unit with { Stdout } = () -> {
  let d: Double = 3.14
  stdout_write("d = \{to_string(d)}\n")
}
```
→ also prints a small garbage integer, not `3.14`.

## Workaround used in the tutorial

Avoided printing a raw `Double` and instead demonstrated it via `Double::to_int`, which works correctly:

```vibe
stdout_write("d*100 as int = \{Double::to_int(d * 100.0)}\n")
```
→ `d*100 as int = 314` (correct).

## Notes

- `to_string[T: Show](x: T) -> String` (`lib/@vibe/prelude/builtin_traits.vibe:141`) delegates to the `__to_string` builtin — the bug is presumably in that builtin's (or the checker/codegen's `Show` derivation's) handling of the `Double` runtime representation, likely treating it as a tagged-int and reading/printing garbage bits instead of doing an actual float-to-decimal conversion.
- `Int`/`Bool`/`String`/`Char` interpolation and `to_string` all work correctly — this is Double-specific.
- Not fixed as part of #1142 (doc-tooling change, compiler left alone); tracking here since docs currently can't show a real Double value in `vibe run`/`vibe test` output until this is fixed. `docs/tutorial/01_values_functions.vibe.md`'s "値と基本型" section has an inline comment noting the workaround and pointing at this issue once filed.

Repro compiler: `_build/selfhost/generations/vpkg-structured-header-2026-07-27_*/stage2.wasm` (current `bootstrap/seed.json` pin as of this report).

mizchi:
## Summary

Root-caused and fixed #1153 and #1185, which turned out to share a single underlying bug (as #1185's own author suspected). Also verified #1184 is already fixed on `main` as a side effect of #1199 — no code change needed for it.

**Root cause:** `let x: Double = e` (or `Float`) parses to the ascription marker `((__asc: Double) -> __asc)(e)` (`parser_expr_dispatch.vibe`'s `ascribe_wrap`). `expr_is_floatish`'s `ECall` case already recognizes this wrapper shape for `Bool` (added in #815), but that recognition was never extended to `Double`/`Float` — so an explicitly-annotated Double/Float local's slot never gets logged into `float_local_slots`.

Under RC (the production default backend), a float is a **boxed heap value** (#704). The missing classification meant:

- **#1153**: `__to_string`/string interpolation on such a misclassified local read the boxed pointer as a plain `Int`, printing garbage (`d = 60` instead of `d = 3.14`).
- **#1185**: `==` between two independently `: Double`-annotated locals fell through `emit_eq_value`'s generic `eq` fallback, which compares the two *distinct boxes by pointer* instead of unboxing and comparing values — silently `false` for two equal doubles. A single annotated operand paired with a literal or unannotated one was unaffected, since that operand's own correct classification was enough for the `||` dispatch in the `==`/`!=` codegen to still route to the correct unboxing path; only two independently-misclassified operands compared together triggered the bug — which is exactly the pattern #1185 reported.
- **#1184** (`String </>` via `Array::get`): independently already fixed on `main` as a side effect of #1199's `array_elem_bind` fix — verified with a fresh build *before* starting this investigation, so no code change is included here for it.

## Fix

Added the same ascription-wrapper recognition `expr_is_boolish` already has for `Bool`, extended to `Double`/`Float`, in both:
- `expr_is_floatish` (`lib/@vibe/compiler/codegen/expr/compile_expr_tail.vibe`)
- its "local mirror" `cc_expr_is_floatish` (`lib/@vibe/compiler/codegen/expr/compile_call.vibe`), used by `__to_string`'s call-site dispatch

## Test plan

- [x] `bash scripts/compiler_gate.sh` — 67/67, stage2==stage3 fixpoint, golden lang-review 10/10.
- [x] `bash scripts/unit_test_runner.sh` — 464/464 allowlisted files pass.
- [x] `bash scripts/vibe_fmt.sh --check` on all touched files — clean.
- [x] New regression tests in `fixtures/double_to_string_test.vibe` covering both the `to_string`/interpolation and `==` cases for an annotated Double local (the one shape missing from that file's existing "dispatch is SHAPE-based" coverage list).
- [x] Manual end-to-end verification with a fresh stage2 build against all three issues' exact repros:
  ```
  d = 3.14                                    # #1153, was "d = 60"
  annotated a==b=true                         # #1185, was false
  annotated c == inline 1.0=true
  annotated a == unannotated d=true
  literal zz>aaa=true                         # #1184, already fixed by #1199
  array-get zz>aaa=true
  array-get-then-let zz>aaa=true
  ```

---


mizchi:
Pushed `9f23bc7` addressing the Codex review comment: hardened both `expr_is_floatish` and `cc_expr_is_floatish` to match the *complete* synthetic identity-lambda shape (no type params, no declared return type, no effect row, body exactly `EIdent(<param's own name>, _)`) — mirroring `ascription_lambda_ann`'s precision in the checker — instead of only checking the marker param's name and type. Added a regression test for the exact scenario raised (`((__asc: Double) -> String { "hello" })(1.0)` correctly stays a `String`, not misclassified as floatish).

Re-verified: `compiler_gate.sh` (67/67, fixpoint ok, golden 10/10) and `unit_test_runner.sh` (464/464) both green.

On the earlier perf report showing a uniform +25–64% wall-time regression across unrelated benchmarks (including `fib30` pure computation and lexer/parser benches this diff never touches): built the base commit (`bb92d59`) and this fix back-to-back on the same machine to control for CI runner noise. Same-machine comparison shows normal-range deltas (selfcompile wall_ms -4.85%, parser/lexer/checker benches -0.9% to +4.7%), with only `fib30` as a single outlier (+26.76%, consistent with known jitter on that particular sub-100ns benchmark — see #1204's perf history, where the same series spiked +72% once and returned to ±0% on the next run). The CI-reported slowdown does not reproduce locally and is not gated (perf-metrics is advisory-only), so treating it as CI-runner noise rather than a real regression from this change.

---

mizchi:
Follow-up on the perf report: the same uniform slowdown pattern (+20-43% across every benchmark, including ones this diff never touches) reproduced on this second push too. Checked #1206 (the immediately-preceding merged PR, a docs/tooling-only change with zero compiler-internal touches) for comparison — its perf report shows the **opposite**: a uniform **-15% to -27%** (🎉) across every single benchmark relative to its own baseline.

Two unrelated PRs each showing every metric move together by a similar magnitude, in *opposite* directions, is the signature of CI-runner-to-runner variance (this run happened to land on a slower/faster shared runner instance), not a code-driven regression or improvement in either PR. Combined with the earlier same-machine back-to-back comparison (which showed only normal-range deltas), I'm confident this is CI noise and not something introduced by this change. Not blocking — `perf-metrics` is advisory-only and `ci-required` already passed.

---

Issue #1114, opened 2026-07-26
is_inlined_scalar_builtin/is_inlined_async_builtin skip local-parameter shadowing of builtin names inside nested closures


mizchi:
## Summary

`is_inlined_scalar_builtin` / `is_inlined_async_builtin` in `lib/@vibe/compiler/codegen/common_analysis/common_analysis.vibe` (introduced for `not`/`add`/`sub`/`mul`/`lt` in #773, extended to `assert` in #1095/#1113) special-case bare-name call callees that are inline-lowered builtins so a nested lambda body referencing them doesn't wrongly capture them as free variables.

The doc comment on `is_inlined_scalar_builtin` only describes protection against **top-level** shadowing (a same-named top-level function, which shows up in `fn_names`). It does **not** protect against a **local** shadowing binding from an *enclosing* function/closure scope — e.g. a function parameter literally named `not` referenced from a nested inner lambda:

```vibe
fn outer(not: (Bool) -> Bool) -> Bool {
  let inner = () -> Bool {
    not(true)
  }
  inner()
}

test "shadowed parameter named not is captured, not treated as builtin" {
  let r = outer((b: Bool) -> Bool { b })
  assert(r == true)
}
```

This traps with `RuntimeError: unreachable` (the assertion fails) — `inner`'s free-variable analysis only checks its own `bound` (empty, since `inner` has no params) and the global `fn_names` list, so the inline-builtin check fires and `not` is never captured from `outer`'s scope. The call site then falls through to the actual builtin logical-not lowering instead of invoking the captured closure parameter, silently computing the wrong value instead of erroring.

## Impact

Low probability in practice (shadowing a well-known builtin name like `not`/`add`/`sub`/`mul`/`lt`/`assert` with a local parameter/binding is unusual and already discouraged), but it's a silent wrong-value bug rather than a compile error, which makes it worse than a mere crash if ever hit.

## Root cause

`collect_free_vars_expr`'s `ECall(EIdent(cname, _), args, _)` branch (around line 731) only checks `bound` (the *immediate* lambda's own parameter/local list, reset per closure — see `compile_lambda.vibe`'s `param_names_arr`) and `fn_names` (top-level function/constructor names). It has no way to see that an *enclosing* lambda/function scope binds `cname` locally, since free-variable analysis for each nested lambda starts a fresh `bound` array from that lambda's own parameters only.

## Suggested direction

[...]

PR #1211, opened 2026-07-29
#1203 investigation: fix collect_free_vars encl/global precedence gap

### LINKED ISSUE #1203: Top-level fn named exactly `f` or `g` breaks linear-backend codegen (root cause still unknown)
**Status: parked, not yet fixed.** See comments below for the full investigation trail (multiple corrected hypotheses — read bottom-to-top for the current understanding, not the original report at the top of this issue).

## tl;dr for whoever picks this up next

**Repro** (via `bash scripts/vibe_md.sh check` or `scripts/vibe_run.sh`, seed or any stage2 generation):

```vibe
import @vibe/prelude { stdout_write }

fn f() -> Int { 1 + 2 }

fn main with { Stdout } {
  stdout_write("r = \{f()}\n")
}
```

Compiles with exit 0 and no diagnostics (`vibe check` is clean), but the emitted wasm fails to instantiate:
```
CompileError: WebAssembly.instantiate(): ... expected 1 elements on the stack for fallthru, found 3 ...
```
(or, depending on arity, "not enough arguments on the stack for call"/"call_indirect".)

**Confirmed:**
- Trigger is the *exact* top-level identifier `f` or `g` (any other name, including other single letters, is fine) used as a `fn`/`let` name, whenever `@vibe/prelude` is linked in.
- Reproduces whether or not the function is ever called.
- NOT caused by `@vibe/prelude/func.vibe`'s `compose`/`flip` parameter names (disproved by renaming them and rebuilding — still crashes).
- `wasm-dis`-diffing an `f`-named build against an otherwise-byte-identical `helper`-named build shows the divergence is in an early, shared/generic builtin function compiled once per module (not in `f`'s own body, not at its call sites) — missing a whole `if` block, plus two nearby constants that differ (40→48, 3→4), suggestive of a struct-field-count/offset shift.

**Blocked on:** live-instrumenting `compile_wasi_module_linked_impl` (`lib/@vibe/compiler/codegen/wasi/linked_compile.vibe`) to dump internal state (`func_table`, etc.) failed — even a single added line makes the *compiler's own self-compile* fail with `stack overflow: expression too deeply nested` before producing a usable probe binary. That function is apparently already near some AST-nesting/statement-count limit on its own. Next attempt should extract the relevant logic into a small, separately-callable, probe-safe function first (a real refactor, not a throwaway debug edit), rather than editing that function in place.

**Workaround:** don't name a top-level function `f` or `g`.

mizchi:
## Summary

While deep-diving issue #1203 (a top-level function named exactly `f`/`g` breaks the linear-backend linked-compile pipeline), found a related-but-distinct correctness gap in `collect_free_vars_expr_sc`'s plain `EIdent` case (`lib/@vibe/compiler/codegen/common_analysis/common_analysis.vibe`):

- It checked `fn_names` (top-level globals) **before** `encl` (the enclosing scope's own locals, e.g. an outer function's parameters), so a bare identifier reference shadowed by an enclosing local was wrongly treated as "already global, no capture needed" whenever the merged program also happened to define a top-level function of the same bare name — silently dropping a real closure capture.
- This is the same gap issue #1114 already fixed for the `is_inlined_scalar_builtin`/`is_inlined_async_builtin` inline-builtin skip check, just never extended to the general plain-identifier case.
- Fixed by checking `encl` before `fn_names`, matching the precedence the `#1114` fix already established elsewhere in the same function.

**This does not fix #1203 itself** — see the issue for the full investigation writeup. A minimal, prelude-free 2-file repro still reproduces identically with this fix applied; the actual defect is upstream of `collect_free_vars_indexed_self` (which behaves correctly given the same inputs, per the new unit test). This fix is landed on its own merits as a real, isolated, regression-tested correctness fix for the same bug *class*.

## Test plan

- [x] New unit test `lib/@vibe/compiler/cfv_encl_test.vibe`: `collect_free_vars_indexed_self(body="g(f(z))", bound=["z"], fn_names=["f"], encl=["f","g"])` now correctly returns both `f` and `g` (previously only `g`). Registered in `scripts/unit_test_allowlist.txt`.
- [x] `stage2 == stage3` fixpoint holds.
- [x] `bash scripts/compiler_gate.sh` — 67/67 ok.
- [x] `bash scripts/unit_test_runner.sh` — 465/465 (incl. the new test).

---
_Generated by [Claude Code](https://claude.ai/code/session_01MoHiTv1shjrM2aFUUd8xFm)_


mizchi:
Verified this fix independently and it **does** resolve #1203, contrary to this PR's own description — the "doesn't fix #1203" conclusion looks like it came from a rebuild that silently reused the stale committed `_cli_adapter_module_source.vibe` (`generate_bundle.sh` only regenerates it from fresh source with `VIBE_REGEN_MODULE_SOURCE=1`; I hit this exact false-negative myself before catching it).

Built this branch's own commit (`028c92a`) exactly as-is and confirmed the original issue repro now compiles+runs correctly (`r = 3`, previously crashed at instantiation). Full evidence and root-cause writeup in #1203.

Opened #1212 with this same fix (independently re-derived) plus the regenerated bundle files and an end-to-end `compiler_gate.sh` step to prevent this exact stale-rebuild confusion from recurring. Feel free to close this one in favor of #1212, or merge this first — the core fix is identical either way.

---

mizchi:
Closing — #1212 (same core fix, plus the additional `EAssign`/`EAssignOp` precedence fix from Codex review and end-to-end gate coverage) merged into `main` as `2804072`.

---

PR #1303, opened 2026-08-01
Carry table_slots_used through the codegen body cache, and test the replay path (#1259)

mizchi:
## Summary

Follow-up to #1300. The replay half of #1259 step 7 shipped with **no test exercising it**, and it was wrong.

`table_slots_used` is a module-level, append-only log: codegen pushes to it while compiling a function, and it is read exactly once **after** the body loop, to decide which table slots the element section registers. It does not affect a function's own bytes — which is why a fixpoint build could never notice — but a replayed function never runs those pushes. The moment a caller supplied a non-empty cache, the module would have emitted a smaller element section than the same program compiled straight through.

Nothing passes a non-empty cache today, so this is latent rather than a live bug. It would have surfaced as "the parallel build produces different bytes" only after `--jobs N` was wired up, which is the worst possible time to find it.

### The fix

The cache now carries the slots each function recorded, keyed by owner, and replay pushes them back. The consumer builds a flag set, so order and duplicates do not matter — which is also what will let per-slice harvests merge freely once bodies are actually farmed out.

### Making it testable

`body_cache_out` becomes a parameter instead of a local: a harvest nothing can reach cannot be checked. `compile_wasi_module_replayed` then compiles a program twice — once compiling and harvesting every body, once replaying every one — and throws with the offset of the first differing byte. It is deliberately off every production path (it compiles twice on purpose); its only callers are the test and the gate.

`codegen_body_cache_test.vibe` asserts whole-module byte equality across five shapes: no lambdas, lambdas spread across several functions, closures reaching `call_indirect`, nested lambdas, and module-level thunks.

### Mutation check

Rather than assume the new test has teeth, I built a compiler whose replay deliberately drops the recorded slots — the pre-fix behaviour — and ran the cases against it:

| case | mutant |
|---|---|
| no lambdas | survives |
| lambdas across several functions | **killed** |
| closures reaching `call_indirect` | **killed** |
| nested lambdas | **killed** |
| module-level thunks | survives |

The two survivors build no lambdas at all, so they record no slots and *cannot* distinguish the two implementations. That is the expected shape, not a coverage hole, and the test header says so — they are there for the other axis, that the plain and thunk body paths replay at all.

### Also fixed

Five call sites outside `linked_compile.vibe` that #1300's signature change missed (`linked_artifacts.vibe`, `linked_library.vibe`, `cli_support.vibe`, and two in `_selfbuild_gate_block_a.vibe`). They broke ten unit-test files with an arity mismatch **while the self-build fixpoint stayed green**, because the self-build path never typechecks those files. Worth recording: fixpoint alone does not cover a signature change.

## Test plan

- stage2 == stage3 fixpoint
- bundle sync / module source sync
- size ratchet, all samples within tolerance
- `vibe-fmt-check` clean (770 files)
- **480/480** unit-test files, including the new one
- compiler-gate

Those numbers are from the pre-rebase tree. The branch was rebuilt on top of the current main (#1299, #1301) after #1300 merged, and the same battery is re-running against that base now; I will follow up here if anything moved.

## Not done

`--jobs N` is still not wired up. Remaining: slice bounds, entry points that pass a non-empty cache, `file_compile`-side serialization, and a host script.

https://claude.ai/code/session_01PENPDUZBtGEyxx27pMVtsa

---


mizchi:
Following up on the note in the PR body: the full battery finished against the rebased base (on top of #1299 / #1301) and nothing moved.

- stage2 == stage3 fixpoint
- bundle sync / module source sync
- size ratchet, all samples within tolerance
- `vibe-fmt-check` clean (770 files)
- **480/480** unit-test files
- compiler-gate

---

PR #1419, opened 2026-08-04
feat(checker): retain typed occurrence expression paths

mizchi:
## Summary

Next Phase 3 slice for #1379, building on #1413, #1415, and #1418.

Adds an opt-in opaque observation that combines, for every retained legacy typed-occurrence row:

- retained post-desugar statement path
- exact post-desugar `core::expr_children` expression path
- append-time role and checker lane
- raw checker type, canonicalized through `final_subst` in the versioned snapshot

Expression root is `[]`; each structural child appends its zero-based `expr_children` index. Provenance is captured adjacent to the legacy append and is never reconstructed from offsets.

## Complete-or-none boundary

The API returns `Some` only when every retained legacy row has an honest structural path and counts align. It returns `None` rather than publishing partial/fabricated provenance when:

- a synthetic rewrite appends a row
- an auxiliary resume-value recheck appends a row
- traversal frame accounting is incomplete or imbalanced

Special checker traversal is represented explicitly for:

- iterative right-nested `ESeq` / `ELet` / `ELetMut` / `ELetRec` spines
- nested operation calls inside `perform`
- specialized `ECall(EDot(...))` receiver checking
- specialized `__region_run` literal-lambda body checking
- recognized call paths that intentionally skip structural callees

## Compatibility and scope

Unchanged:

- `CheckedProgram` and legacy `typed_occurrences`
- offset/type/order/duplicate/negative-offset semantics
- statement-owner observation and role/lane v1 snapshots
- throwaway checker tables
- ordinary checker API behavior

Normal checking bypasses capture entirely and performs no path allocation. The capture and successful observation are opaque at the package boundary.

This remains observation-only. It is not edit-stable identity, typed IR, persistent transport, cache identity, or production reuse policy.

## Tests

Focused coverage includes:

- root and nested call paths
- `ESeq` / dot projection and shared offsets
- mixed deep iterative let/sequence spines
- nested `perform` call frames
- dot-call callee/object frames
- `__region_run` argument/lambda-body frames
- final-substitution type snapshot
- deterministic snapshots and invalid indexes
- ordinary/with-env parity
- synthetic and auxiliary fail-closed behavior
- compatibility with #1413 and #1418 suites

## Validation

- matching fresh-stage2 focused tests: 4 files / 31 tests
- stage2/stage3 Wasm validation
- stage2 == stage3
- bundle/module-source sync
- formatter and diff checks
- `pkf run release-check`
- independent adversarial review and post-fix re-review: no remaining P0/P1/P2

The local release gate handled the installed wasmtime async-stackful capability probe as its existing supported skip and otherwise completed successfully.


mizchi:
main更新（#1417/#1416）へrebaseし、generated artifactsをcanonical regenerateしました。rebase後もfresh stage2 focused 4 files/31 tests、stage2==stage3、bundle/module-source sync、fmt、`pkf run release-check`が成功しています。

Issue #1549, opened 2026-08-07
incremental P0-2: normal import assembly の value-only 境界 (`env_flat_bindings`) を解消する


mizchi:
#1379 の分解 (P0 の 2 本目)。

## 問題

通常の import assembly は依然 **value-only の `env_flat_bindings` path** を通る。
traits / impls / typedefs / effects を lossless に運ばない。

一方 persistent `TypeEnv` v3 transport は trait headers / method `TypeExpr` /
supertraits / concrete・generic impl / bounds / shadow chain を保持する。
**同じ意味情報について権威が 2 つある状態**で、production reuse へ進む前に
一本化しておかないと、どちらを鍵にしたのか分からなくなる。

## スコープ

- [ ] `TypeEnv` v3 transport / module interface v2 / normal import assembly の
      **権威を一本化する**
- [ ] first-match / shadow / binder provenance を維持する
- [ ] 旧 transport を**暗黙 upgrade しない**（fail-closed で cold へ落とす）
- [ ] declaration loss が解消されたことを、trait / impl / typedef / effect の
      それぞれについて fixture で示す

## 関連

- #1379 (親)、#1397 / #1403 (TypeEnv v3 の strict recursive codec)
- docs/incremental-build.md

mizchi:
A trait-projection prototype was implemented and independently reviewed but deliberately not merged. Review exposed two correctness prerequisites: (1) the FS/import checker currently calls bounds checking in `send_only` mode, so non-Send user bounds can succeed without an impl; (2) explicit trait aliases must override value-derived mappings independent of import-item order and rewrite selected `CtForAll` bound labels. The prototype did establish selected/exported authority filtering and a real root-cache-cleared warm dependency test. Next slice should fix these P0 semantics before transporting trait impl nodes; typedef/effect transport remains later scope.

mizchi:
Trait-authority slice merged in #1608. Selected/exported trait defs plus concrete/generic impls now travel through FS and in-memory import assembly; explicit aliases are order-independent, imported `CtForAll` labels are rewritten, generic-impl bound traits are closed over, cross-dependency explicit alias collisions fail closed, and the old non-`Send` user-bound bypass is removed. Cold/warm missing-impl controls and full required CI passed. Keeping #1549 open for the remaining typedef/effect declaration transport and final authority unification.

mizchi:
[...]

PR #1613, opened 2026-08-10
feat(checker): transport typedef declaration authority (#1549)

mizchi:
## Summary
- make exported typedef declarations part of the transported environment across FS, worker/persistent, TDRE, and in-memory TypeDb paths
- preserve generic typedef provenance, aliases/reexports, local shadow order, and strict first-match semantics
- validate v4 transport on every cache-hit path and cold-fallback old/malformed data
- reject selected effects/effectsets explicitly until their declaration authority lands
- retire the legacy enum carrier as an imported checker authority

## Validation
- focused codec and FS cross-module tests passed
- worker scheduler parity, persistent-cache suite, and TDRE oracle passed
- formatter, generated freshness, and `git diff --check` passed
- final independent review: safe to integrate, no blocker/major/minor findings

## Remaining #1549
Effect/effectset declaration transport and final authority unification remain follow-up scope.

### DIFF
--- a/lib/@vibe/compiler/checker/artifacts/checked_effective_effect_row_observation.vibe
+++ b/lib/@vibe/compiler/checker/artifacts/checked_effective_effect_row_observation.vibe
@@ -104,6 +104,9 @@ fn owner_env_lookup(env: TypeEnv, wanted: String) -> Option[Type] {
       EnvTraitImplGen(_, _, _, _, rest) => {
         cur = rest
       },
+      EnvTypeDefs(_, rest) => {
+        cur = rest
+      },
       EnvFlat(bindings) => {
         let mut i = 0
         while i < Array::length(bindings) && !done {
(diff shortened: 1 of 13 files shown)


mizchi:
Closing rather than weakening authority after full CI exposed an architectural prerequisite. The strict selected-effect rejection is sound in isolation but breaks existing package/contract imports such as `Random`, `Fs`, and other effect names because effect/effectset declarations are not yet carried by `EnvTypeDefs`. Restoring uppercase fallback would preserve CI only by reintroducing the silent undeclared-effect path this slice is intended to remove.

The typedef carrier, strict v4 cache validation, and focused tests are useful prototype evidence, but independent review correctly found no independently sound subset, and full CI confirms typedef/effect transport cannot be ordered as initially planned in this repository. The next #1549 slice must transport typedef + effect/effectset declaration authority atomically across package contracts, FS/worker/TDRE, and in-memory TypeDb before removing the legacy fallback.

PR #1638, opened 2026-08-12
feat(async): support direct suspended selection inputs

mizchi:
## Summary
- normalize direct recognized suspension in `if` conditions and `match` scrutinees onto the existing CPS `let` spine
- cover sequence-head and tail forms with collision-safe fresh binders
- preserve fail-closed rejection for compound inputs and provenance-sensitive cases
- update ADR/spec/cheatsheet/triage documentation

## Validation
- independent verifier: accepted after one fixture fix; no implementation blocker
- `pkf run test-local`: 160 selected tests passed in source checkout
- focused Phase 3a compiler gate: passed
- formatter and generated freshness: passed
- full local gate still reaches unchanged macOS xargs command-size portability failure

Closes no umbrella issue: #1536 remains open for row-variable evidence, residual literal provenance, loops/for-in, and assignment RHS.

### DIFF
--- a/docs/cheatsheet.md
+++ b/docs/cheatsheet.md
@@ -28,7 +28,7 @@ vibe build --release app.vibe  # standalone .wasm
 
 ### CLI は IDE 相当のクエリ面 (方針)
 
-`vibe check` (`--single-file` 込み) / `vibe symbols` / `vibe type-at` /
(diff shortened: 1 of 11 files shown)


mizchi:
Closed without merge: branch construction accidentally copied full files from a stale standalone-clone baseline and reverted unrelated main updates. Reconstructed the reviewed patch as a narrow raw diff on current main; replacement PR follows.

PR #1659, opened 2026-08-13
async: normalize direct suspended compound assignment RHS

mizchi:
## Summary
- normalize only direct recognized suspension inputs in compound-assignment RHS onto the existing CPS spine
- preserve `EAssignOp` target/operator semantics and freshen the synthetic binding against target-only collisions
- retain fail-closed rejection for compound/nested RHS suspension
- migrate the positive fixture to source-owned `inspect(...)` coverage and update focused gate/docs

## Validation
- focused inspect snapshot: 1/1 passed
- compiler gate passed through phase 91 including stage2/stage3 fixpoint; local phase 92 stopped at the known macOS `xargs` command-length limitation
- formatter and generated freshness passed
- `bit diff --check` passed
- independent review initially rejected target-field hygiene; bounded fix added; independent rereview ACCEPT

## Scope
Bounded structural #1536 slice. Row-variable evidence, residual literal provenance, type-aware `for-in`, and other compound structural forms remain open.

Refs #1536


mizchi:
Closing in favor of broader PR #1658, which was already open before this PR and includes the same EAssignOp target-order fixes plus a general ANF compound-suspension implementation and green required CI. The independently reviewed bounded slice remains useful evidence, but merging both would duplicate/conflict. No commit from #1659 is integrated.

PR #1664, opened 2026-08-13
feat(async): split tail short-circuit suspension

mizchi:
Bounded #1536 slice: support a suspending RHS only when the entire `&&` / `||` expression is in suspend-CPS tail position.

The lowering rewrites that whole tail expression to its semantics-equivalent `if` before generic compound ANF, reusing existing condition/branch splitting. Non-tail assignment/call/selection contexts remain fail-closed.

Evidence:
- source-owned snapshot covers bypass and taken paths for both operators
- records handler entry, resumed source continuation, and handler-after-resume events with exact count/order
- existing short-circuit, if-branch, and match-branch non-tail negatives remain rejected

Validation:
- focused positive: 1/1
- three focused negatives rejected
- fixture accounting 242/242
- seed → stage1 → stage2 → stage3 and stage2/stage3 fixpoint passed
- formatter, generated freshness, and `bit diff --check` passed
- independent review rejected the first evidence shape, one bounded test-only fix was applied, and final rereview ACCEPTED

The broader #1536 remains open for Array-typed `for`, loop-body return, row-variable evidence, residual literal provenance, and non-tail conditional positions.

### DIFF
--- a/docs/effect-evidence-passing.md
+++ b/docs/effect-evidence-passing.md
@@ -2890,7 +2890,12 @@ pass で拾われる。callee は名前を付けない — by-name call を loca
 変えてしまうと、それこそ suspend lowering が see-through できない唯一の形になる。
 
 **必ず評価されるとは限らない位置だけがこの ANF では fail-closed のまま**:
-`if` / `match` の枝、`&&` / `||` の右辺。closure literal は
+`if` / `match` の枝、`&&` / `||` の右辺。後続スライスでは、式全体が tail の
+`l && r` / `l || r` だけをそれぞれ `if l { r } else { false }` /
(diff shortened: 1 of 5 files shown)


mizchi:
Conflict resolved against current main `6ebc491` by clean reconstruction, preserving #1662 fixtures and behavior. New head: `7308bd0`. Independent conflict review first caught malformed shell continuations; those were fixed in the single bounded conflict-resolution pass. Final rereview: ACCEPT. Verified both #1662 positives remain in their stage2 invocation, the tail fixture runs separately, loop-control collision remains gated, `bash -n`, fixture accounting 244/244, generated freshness, and `bit diff --check` pass. Fresh required CI is mandatory after this force update.

PR #1665, opened 2026-08-13
test(gc): pin Array[Int] control-flow join fallback

mizchi:
## Summary

- add a dedicated explicit `Array[Int]` control-flow join fallback fixture
- verify true-path identity by mutating the selected alias and reading the original
- compile/validate/run it through the existing GC fallback helper
- require decoded `array.new_default` count to remain zero

No ABI or codegen eligibility changes. This is evidence-only follow-up for #1541.

## Validation

- `bash scripts/vibe_fmt.sh --check fixtures/gc_direct_array_join_fallback_test.vibe`
- `bash -n scripts/test_gc_heap_accounting.sh`
- `bash scripts/check_fixture_execution.sh` (245/245)
- `bit diff --check HEAD^ HEAD`
- independent review: ACCEPT

Full local GC execution is unavailable with local Rust 1.92 / missing current stage2; fresh CI must provide authoritative GC execution.

Refs #1541

### DIFF
--- a/fixtures/gc_direct_array_join_fallback_test.vibe
+++ b/fixtures/gc_direct_array_join_fallback_test.vibe
@@ -1,0 +1,25 @@
+// #1541 fallback boundary: joining native-reference candidates through control
+// flow is outside the private direct Array[Int] ABI proof. The complete
+// component must retain the tagged-i64 representation.
+
+fn choose_array(values: Array[Int], choose_original: Bool) -> Array[Int] {
+  if choose_original {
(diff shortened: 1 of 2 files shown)


mizchi:
Addressed the inline review in `d883a232`: both new expected values now use updateable `inspect(...)` snapshots instead of `assert`. Formatter, Bash syntax, fixture accounting (245/245), and diff checks pass. Fresh required CI is running on the amended head.

Issue #1916, opened 2026-08-16
codegen: statement-position `for` materializes a discarded comprehension array per execution — and it leaks under RC


mizchi:
Found while attributing #1875 R4 heap costs (the +5.76MB/used-effect was mostly this, paid by per-effect whole-program walks written with `for child in children { visit(child) }`).

## Symptom

`for x in xs { … }` is a comprehension in the residual builtin lowering: the linear-lane `EForIn` codegen (`compile_expr_tail2.vibe`) unconditionally allocates a result array and pushes every body value, **including in statement position where the value is discarded**. Measured on stage2 (linear/RC lane), 10k-element `Array[Int]`, `Profiler::heap_bytes` deltas:

| shape | heap delta per execution |
|---|---:|
| `for x in xs { acc = acc + x }` (statement position) | 131,164 — **and never reclaimed** (×100 → 13.1MB) |
| `let ys = for x in xs { x + 1 }; Array::length(ys)` (used) | first 131,164, then 0 (drop + exact-fit free list recycle) |
| `while` twin | 0 |

So a discarded `for` pays twice: the array nobody reads, and (because the discarded value never gets a drop — see also #1915 for the call-boundary variant) the allocation is permanent bump-heap growth.

The same shape in match-arm tail position of a `-> Unit` visitor (the recursive-walk idiom all over the compiler) reproduces identically: a 6^6-node tree walk costs 783,804 bytes per call with `for`, 0 with `while`.

## Why it matters

This makes the natural spelling of a read-only AST walk the expensive one, which is an implementation detail leaking into the language's cost model. #1875 R4 (PR #1909) hand-rewrote one hot walk (`edp_collect_handle_sites_expr`) from `for` to `while` for a 4× per-used-effect improvement — that class of rewrite should not be necessary.

## Fix directions

1. **Discard-position lowering**: the checker already classifies a direct `EForIn` at an `ESeq` head as statement-shaped (`ce_discarded_root`, #1679). Codegen has no such notion — a discard-context signal (or an AST rewrite of discarded residual-builtin `for` to a `while` loop, where the iterand is provably `Array`) would drop both the allocation and the churn. Tail-of-`Unit`-function positions need discard propagation to be covered, which is the harder half.
2. Independently of (1), **the discarded result should get a drop** so it is at least reclaimed by the free list.

[...]

Issue #1915, opened 2026-08-16
RC: builder-shaped returns are never dropped by callers — the #707 bare-EIdent-tail fallback classifies them borrow-returning


mizchi:
Found while attributing #1875 R4 heap costs. This is likely the single largest systematic contributor to the ~578MB self-compile heap high-water.

## Symptom

Any function whose body tail-returns a local binding by name — the standard builder shape —

```vibe
fn make_arr(n: Int) -> Array[Int] {
  let out: Array[Int] = []
  let mut i = 0
  while i < n {
    Array::push(out, i)
    i = i + 1
  }
  out                    // bare EIdent tail
}
```

has its **call results never dropped by any caller**, used or not. Measured on stage2 (linear/RC lane), `Profiler::heap_bytes` around 100 calls with n=1000 (~8.3KB per array):

| shape | heap delta ×100 |
|---|---:|
| `let x = make_arr(1000); Array::length(x)` | 828,400 (leaks every call) |
| `let x = make_arr(1000); ()` (unused) | 828,400 |
| `make_arr(1000)` statement position | 828,400 |
| same loop written **inline** (no call boundary), used or unused | **0** (scope-end drop + exact-fit free list reclaim everything) |

The call boundary is the discriminator: inline construction is fully reclaimed, returned construction is never reclaimed.

## Cause

`compute_borrow_returning_names` (#707) classifies a function as borrow-returning when every tail it can produce comes from a known borrow-returning name **or a "safe to never drop" fallback — and a bare `EIdent` tail is accepted as that fallback** (`is_borrow_ret_fname` / the fixed-point doc comment in `lib/@vibe/compiler/perceus/perceus.vibe`). The fallback was designed for nullary-ctor-ish sentinels (`TEof`), where the worst case is leaking one small value. But a bare EIdent tail is also the *most common owned-return shape in the codebase* (`out` / `result` / `acc` builders), so every such function joins `pctx_names`, `is_scalar_expr` marks caller bindings of its results scalar, and no caller ever plans a drop.

The direction of the tradeoff is right (leak, never double-free) — the *reach* of the fallback is what's wrong.

## Proposed fix (R5 candidate)

In the fixed-point, resolve a bare `EIdent` tail to its binding inside the function body instead of accepting it unconditionally:

- tail resolves to a `let` whose value is a **heap literal / builder construction / call to an owned-returning fn** → the function is owned-returning (callers keep their scope-end drop);
[...]

PR #1923, opened 2026-08-16
perf(codegen): compile discard-position for-in without the comprehension result array (#1916)

mizchi:
## Summary

Fixes the #1916 lowering gap: a statement-position `for` compiled as a full comprehension — the linear `EForIn` lowering materialized a result array per execution, and the discarded array never got a drop, so it was both churn and a permanent leak. The measured costs (micro-fixtures on stage2, `Profiler::heap_bytes`):

| shape | before | after |
|---|---:|---:|
| discarded `for` over 10k `Array[Int]`, ×100 calls | 13,116,400 (leaks) | **0** |
| recursive `-> Unit` visitor with `for`-over-children arms (6^6-node tree), per call | 783,804 | **0** |
| value-position `for` (`let ys = for …`) | unchanged | unchanged |

This is the structural version of what #1909 hand-fixed for one hot edp walk: the natural spelling of a read-only walk no longer pays an allocation per node.

## How

- `compile_expr_tail2.vibe`: the `EForIn` lowering is extracted into `compile_forin_loop(collect: Bool)`. `collect=false` keeps the identical block/local layout (break/continue depths, slot numbering) and differs only in: no result array (the slot holds i64 0), the body value is dropped, and the loop's value is 0. `collect=true` is byte-for-byte the old lowering.
- `compile_expr_tail.vibe`: an `ESeq`-head `for` — the same statement-shaped classification the checker uses (#1679) — compiles through the no-collect form; its value is dropped by the very next instruction.
- `forin_discard_pass` (end of `effect_lowering_prelude`, linear lane): funnels the other syntactic discard positions — statement roots, test/bench bodies, `while` bodies, let/assign continuations, if/match tails under a discarded context, declared-`Unit` fn body tails — into that one `ESeq`-head shape by wrapping the loop as `ESeq(for …, EUnit)`. Runs last, after every effect pass, when the only surviving `EForIn`s are the residual builtin lowering (#1679). `EHandle` is left untouched (conservative).
- The pass gates its per-statement rebuild on an allocation-free `RQContainsForIn` read query and shares any subtree without a `for` (a first cut used `expr_children`, whose per-node children arrays cost 9.5MB on the KPI compile — measured and removed).

## Verification

- stage2 == stage3 fixpoint on the rebuilt generations.
- Micro-fixture behavior identical (sums/lengths byte-for-byte, only heap deltas change).
- Self-compile KPI (codegen_lexer_test, warm, deterministic): 581,323,128 with this change; baseline comparison against current main and the full operation gate are running now — results will be posted on this thread before this is ready to merge.

Note: unlike #1909 this is deliberately **not** output-preserving (that is the point — discarded loops compile to less code), so the validation story is fixtures + gates rather than byte-compare.

## Related

- #1915 (R5 candidate) remains: builder-shaped returns are never dropped by callers; that leak class is untouched here.

https://claude.ai/code/session_01CunRRnvmDfVWw5MzzmjjbQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01CunRRnvmDfVWw5MzzmjjbQ)_

### DIFF
--- a/lib/@vibe/compiler/codegen/common_base/index.vpkg
+++ b/lib/@vibe/compiler/codegen/common_base/index.vpkg
@@ -244,6 +244,11 @@ fn inline_direct_performs(stmts: Array[Stmt]) -> Unit
 // --- evidence_dict_pass
 fn evidence_dict_pass(stmts: Array[Stmt], entry_name: String) -> Array[String]
 
+// --- forin_discard_pass (#1916): wrap every discard-position `for` as
+// `ESeq(for .., EUnit)` so the linear ESeq lowering compiles it without the
+// comprehension result array. Runs at the end of effect_lowering_prelude.
+fn forin_discard_pass(stmts: Array[Stmt]) -> Unit
+
 // --- suspend_cps_pass (ADR-0076 Phase 3a, #817)
 fn suspend_cps_pass(stmts: Array[Stmt]) -> Array[String]
 
--- a/lib/@vibe/compiler/codegen/common_base/inline_direct_perform.vibe
+++ b/lib/@vibe/compiler/codegen/common_base/inline_direct_perform.vibe
@@ -1349,7 +1349,11 @@ enum EdpReadQuery {
   RQLiteralNeeds((String, Array[String], Array[String], Int, Array[String]));
   RQTrigger;
   RQHandlesEffect(String);
-  RQNeedingValueArg(Array[String])
+  RQNeedingValueArg(Array[String]);
+  // #1916: does the tree contain any EForIn? Gates forin_discard_pass's
+  // per-statement rebuild without allocating (expr_children-based recursion
+  // allocates a children array per node, which defeats the pass's purpose).
+  RQContainsForIn
 }
 
 fn edp_rq_is_count(q: EdpReadQuery) -> Bool {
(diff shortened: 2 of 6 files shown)


mizchi:
Pushed `c2a53eb`:

- **Codex review addressed**: `forin_discard_pass` now walks let/seq spines iteratively (`fdp_spine`) — each head value is gated and rebuilt exactly once and the nodes fold back right-to-left, so a deep right-nested body costs O(N) with no per-node recursion (same hazard class `check_seq_spine` guards in the checker).
- **vibe-fmt-check**: both flagged files formatted; local `check_vibe_fmt.sh` is clean (954 files, 0 violations).

Verification on the rebuilt tree: stage2 == stage3 fixpoint; micro-fixtures unchanged (discarded `for` ×100: 0 bytes, visitor walk per call: 0 bytes, values identical).

Honest KPI framing, measured against a stage2 built from current main (warm, deterministic): self-compile heap is **neutral** — 581,139,600 (main) vs 581,521,664 (this PR), +0.07%, matching the perf bot's +0.19% cold-series reading. The compiler's own hot walk was already hand-fixed in #1909, and the new pass's one-time rebuild roughly offsets the remaining compiler-side savings. The win is the language-level cost model: statement-position `for` no longer allocates-and-leaks per execution in **user** programs (13.1MB → 0 in the ×100 fixture), and future compiler walks don't need the #1909-style hand rewrite. The residual builder-return leak class is #1915.

---

PR #1969, opened 2026-08-16
Fix named stream values in non-RC components

### LINKED ISSUE #1930: named host stream under `VIBE_RC=0` silently doubles every byte read
Found while triaging #1924. Independent of it.

The named-host-stream adapter (`comp_generate_hostfuture_adapter_core_module`) untags its handle with `param >> 1` and tags the byte it returns with `byte << 1`. That is correct for a core the CLI compiled with RC **on** — `linked_compile`'s tag_mode 1, where an Int is `value << 1`. Compile the same program with `VIBE_RC=0` and an Int is a plain i64, so the adapter's tag is one step too many and **every byte read comes back doubled**.

It does not trap. It returns a plausible number.

## Repro

```vibe
let run: () -> Int with Async = () -> {
  let s = host_stream_named("body")
  host_stream_next(s)
}
```

```
VIBE_RC=1 ... -> 10   correct (VIBE_ASYNC_STREAMS="body=10|15|17")
VIBE_RC=0 ... -> 20   doubled
```

Measured across four shapes, all under `wasmtime 47.0.2` / `viberun`, with both the eager and the parking (`@60`) producer:

| program | `VIBE_RC=1` | `VIBE_RC=0` | expected |
|---|---|---|---|
| single read | 10 | **20** | 10 |
| two reads | 25 | **50** | 25 |
| drain to end | 42 | **84** | 42 |
| read 2 then stop | 25 | **50** | 25 |

Parking makes no difference — this is purely the value convention.

## Why it went unnoticed

`scripts/test_named_hoststreams_component_gate.sh` compiles every lane with the CLI default (RC on) and asserts 42. Nothing in that gate ever exercises the non-RC core, so the doubling has no lane that can see it.

## Relation to #1540 / spec §3.18.4

This is the same hazard the serve lane already hit and documented: **one import name, `vibe.host_stream_read`, with two value representations decided by the compile mode.** The serve lane resolved it by compiling non-RC and emitting an untagged adapter (`comp_generate_serve_stream_adapter_module`). The named lane is still tagged-only, so it is correct for exactly one of the two modes and silently wrong for the other.

## Options

1. **Make the tagging follow the core.** The emitter already receives the core wasm; if the compile mode were recoverable from it (a custom section, the way `vibe.dbgargs` publishes `tag_mode` in break builds only), the adapter could emit the matching shift. This is the honest fix and it makes the hazard structurally impossible for future lanes.
2. **Reject the combination.** If a non-RC core imports `vibe.host_stream_read` and the composer is about to emit the tagged adapter, throw instead of composing. Cheap, and turns a silent wrong answer into a message.

Option 1 subsumes the serve lane's separate untagged adapter and is worth doing before a third lane repeats this.

## Suggested gate

Add an `VIBE_RC=0` lane to `test_named_hoststreams_component_gate.sh` asserting the same 42. It is one more `run_case` and it fails today.

mizchi:
## Summary

- publish a compiled core's value representation in a required `vibe.tagmode` custom section when it uses a host future/stream component adapter
- fail closed on missing, malformed, truncated, unsupported, or conflicting declarations; the serve stream lane accepts exactly plain-i64 mode 0
- centralize adapter value conversion and cover RC modes across eager/parked streams and concurrent/blocked host futures
- emit the serve lane's mode-0 declaration from the validated named `handler` signature even when the body never uses the stream or another Async helper is exported; unrelated HostStream readers stay byte-neutral even when they call the raw read builtin
- record the component-boundary ABI contract in ADR-0105 and the WASI P3 async specification

## Root cause

The named host-stream adapter was generated after the guest core was compiled, but it did not know that core's value convention. It always decoded handles with `>> 1` and encoded bytes and EOS with `<< 1`, which is correct only when RC is enabled. With `VIBE_RC=0`, the core uses plain i64 integers, so every returned byte was silently doubled.

The core now carries the decision structurally when it selects a host future/stream adapter. The composer treats `vibe.tagmode` as an ABI contract rather than a compatibility hint, and never guesses a historical default. The shared host-future/named-stream adapter accepts modes 0 and 1; the serve stream adapter accepts only its actual plain-i64 mode 0. Ordinary core wasm does not cross either adapter boundary and remains byte-identical because it emits no section.

## Validation

- Red on `origin/main`: RC=0 eager named stream returned 84 instead of 42
- focused component-codegen tests: RC mode 0/1, missing, malformed, truncated, unsupported, duplicate, conflicting, and serve mode 0/1/missing
- focused unused-`HostStream` serve componentization with an additional exported Async helper: component validates and exports `body: stream<u8>`; an ordinary non-handler reader has no `vibe.tagmode`; a named stream reader still composes and imports `body: stream<u8>`
- named host-stream component gate: RC=1 and RC=0 both return 42 for eager and parked producers
- named host-future component gate: RC=1 and RC=0 both return 42 for warm and blocked/concurrent producers
- size ratchet: all samples remain within +2%; ordinary wasm remains byte-identical
- affected gate: 665/665 passed
- generation gate after rebasing onto `46d1bd1d7`: stage2 and stage3 are byte-identical (`8c61f3db32ae23501b79e28868049b9346fccf90dfa6126d27da807de7a591af`)
- pre-commit review and architecture lint gates

Closes #1930


mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

PR #1971, opened 2026-08-16
Prevent region capture provenance escapes

### LINKED ISSUE #1938: checker: region escape を row / taint に統合し closure stopgap の取りこぼしをなくす
## バグ

#1262 ADR-0090 と closed #1725 から分離。現在の closure escape 防御は、region body の結果が直接 closure literal で、token由来の名前を直接または裸aliasで捕獲する形に限定された stopgap。outer binding、container、helper 関数経由などは exact な region provenance を失いうる。arena 値が scope 外へ出ると use-after-free / silent wrong になり得る。

## 方向

- region skolem / provenance を型または effect row の正規の情報として運ぶ
- closure、alias、container、helper call、generalization を同じ source of truth で判定
- lowering の保守的 escape と checker の厳密 escape を分離したまま、strict 側を TypeEnv に保持
- #1863 の signature-level region kind と整合させる

## 完了条件

- Red: outer binding / container / helper / nested closure / alias chain の各 escape witness
- positive: region 内完結 closure、copy-out、shadowing は受理
- parser名やliteral形に依存しない判定
- linear bump / wasm-gc arena で escaping pointer を生成できない
- generation-gate、full affected tests、precommit が Green

関連: #1725 #1770 #1863
親: #1262

mizchi:
## Summary

- carry source-unforgeable region capture provenance through checked function rows and ordinary type substitution
- preserve provenance across helper calls, explicit return annotations, nested closures, aliases, and containers
- reject assignments that store active-region captures into outer bindings and remove the syntax-specific closure stopgap
- add negative and positive regression fixtures for all escape witnesses

## Tests

- `pkf run generation`
- `pkf run test-affected -- --changed-from origin/main` (full fallback, 665/665 files passed)
- focused region provenance fixtures (5 negative, 3 positive)
- `bash scripts/ensure_generated.sh --check`
- pre-commit review-derived lint gates
- `pkf run test` passed through the complete #1938 region gate and later stopped at the unrelated existing BSD `xargs: command line cannot be assembled, too long` failure in fixture typecheck gate 92

Closes #1938

### DIFF
--- a/docs/region-mutable-state.md
+++ b/docs/region-mutable-state.md
@@ -235,22 +235,19 @@ region の外に出られない。検査は TaskGroup の雛形を一般化す�
   (`fixtures/region_arena_ok.vibe`、42)と negative
   (`fixtures/err_region_escape_return_value.vibe`、needle
   "region escapes its scope")を固定。
-- **closure capture (#1725、2026-08-14)**: §4 の「捕獲した closure が region
-  型に感染する」は **row 統合前提なので Phase 1 では働かない** —— 型は捕獲を
-  記録しないので、`region r { let l = MutList::empty(r); () -> Array[Int] {
-  MutList::to_array(l) } }` の結果型は `() -> Array[Int]` で skolem 走査から
-  見えず、region 値が closure の環境に隠れて脱出できた。arena が入れば
-  use-after-free なので stopgap を入れてある:
-  `checker/checker_escape.vibe :: region_token_escapes_in_closure` が
-  「region body の**結果そのもの**が closure literal で、region token から
-  直に束縛された名前(とその別名)を捕獲している」形だけを reject する。
-  **capture 自体は正当**(region 内で完結する closure は普通に書く)なので
-  過剰approx にしないことが要件で、汚染は「token に言及する初期化式」と
-  「汚染名の裸の別名」に限り、shadowing は最後の束縛が勝つ。取りこぼし:
-  outer binding / container 経由、helper 関数経由。gate 75 が両方向
-  (`fixtures/err_region_escape_closure_capture.vibe` /
-  `fixtures/region_ok_closure_local.vibe`)を pin。**正しい規則は §4 の
-  row 感染**で、arena スライスの前提はそちら。
+- **Closure capture provenance (#1725, #1938)**: a checked function row carries
+  an internal, source-unforgeable region provenance label. A direct capture is
+  `#region_n`; a capture through an inference variable is `#capture[v]` and
+  resolves when ordinary type substitution binds `v`. These labels are not
+  runtime effects and are excluded from effect-set compatibility, but they
+  survive generalization, explicit return annotations, helper calls and
+  containers. The region boundary therefore uses the same zonked-type scan for
+  direct values and closures. Assignment into an outer cell is rejected while
+  the active-region marker is present in `TypeEnv`. Capture itself remains
+  legal: a closure consumed inside the region, a copied-out value, and a
+  shadowing binding carry no escaping provenance. Gate 75 pins the five escape
+  witnesses (outer assignment, container, helper, nested closure, alias chain)
+  and their positive counterparts.
 - **既知ギャップ(Phase 1 で許容、ADR 本文の設計は不変)**: effect row 統合
   (§3)は未着手 — 別 pass の effect 検査は lambda wrapper を見るため、
   region body 内の effect は enclosing fn に帰属しない。outer-binding scan は
--- a/docs/region-mutable-state.md
+++ b/docs/region-mutable-state.md
@@ -235,19 +235,6 @@ region の外に出られない。検査は TaskGroup の雛形を一般化す�
   (`fixtures/region_arena_ok.vibe`、42)と negative
   (`fixtures/err_region_escape_return_value.vibe`、needle
   "region escapes its scope")を固定。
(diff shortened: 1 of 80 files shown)


mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
Independent verification against #1938's completion criteria, using witnesses written from the issue text rather than from this PR's fixtures. Built `e76784ebd` (seed → stage1 → stage2) and threw laundering shapes at it.

**Most of the class is genuinely closed** — these are all rejected with `region escapes its scope`, and none of them are in the fixture set: tuple payload `(1, closure)`, `match`-arm merge, `if`/`else` merge, a write-only closure `(x: Int) -> Unit { MutList::push(values, x) }`, `Some(closure)`, a generic user enum `Cell[T]::Of(closure)`, and a struct literal with a concrete function-typed field.

**One shape still launders provenance: a monomorphic user enum constructor.**

```vibe
enum Box { Hold(() -> Array[Int]) }
fn build() -> Box {
  region r {
    let values = MutList::empty(r)
    Box::Hold(() -> Array[Int] { MutList::to_array(values) })
  }
}
```

Compiles clean. The identical capture rejected through every other carrier:

| carrier | verdict |
|---|---|
| `(1, closure)` tuple | rejected ✅ |
| `Wrap::{ f: closure }` struct, concrete field type | rejected ✅ |
| `[closure]` array | rejected ✅ |
| `Some(closure)` | rejected ✅ |
| `Cell[T]::Of(closure)` generic enum | rejected ✅ |
| **`Box::Hold(closure)` monomorphic enum** | **accepted ❌** |

Two more in the same family: `[Box::Hold(closure)]` (accepted — the array carrier does not recover what the constructor dropped), and `enum Box { Hold(MutList[Int, r]) }` holding the handle itself with no closure at all (accepted).

The pattern fits "Model parameter-dependent region provenance": substitution carries the rider through a payload that *is* a type variable, and a concrete declared payload type has no variable to carry it. Structs with concrete field types are still caught, so the two nominal carriers disagree — which is the property #1938 asks for (`parser名やliteral形に依存しない判定`).

Not a regression: the same witness compiles on `main` too, so this is a survivor of the old stopgap rather than something this PR broke. But it is inside the class this PR closes the issue on, and the fixture set has no enum-payload witness, so nothing currently pins it.

Repro (from a stage2 built on this branch):

```bash
VIBE_PREOPEN_DIR="$PWD" VIBE_FS_COMPILE=1 VIBE_IMPORT_ABI=raw \
  bash scripts/run_wasm_vibe_host_runner.sh --invoke cli_main "$stage2" \
  fixtures/err_region_escape_enum_payload.vibe /tmp/out.wasm __no_entry__
```

Runtime today prints `41, 42` rather than corrupting: the arena is not reclaimed yet (ADR-0090's watermark is unimplemented), so the escaped pointer still reads intact memory. That makes this latent rather than observable — it becomes a real use-after-free the moment reclamation lands, which is the case for pinning it in the checker now.

One unrelated note: the perf bot flags `parse_only_checker_vibe` B/op **+9.65%** and selfcompile heap **+3.62%**, both in the deterministic table where the repo treats any drift as real.

---

mizchi:
@codex review

mizchi:
Follow-up to the witnesses above: #1993 is stacked on this branch and closes the carrier-erasure half.

Two things worth knowing here regardless of what you do with that PR:

**1. The struct case was wider than my first comment said.** A parameterless struct erases provenance exactly like a parameterless enum — `err_region_escape_named_struct.vibe` passes only because the literal sits in the region body's tail position, where `tail_named_record_mentions_region` inspects it syntactically. Move the same literal behind a `let` alias or into an array and it compiles:

```vibe
struct Wrap { f: () -> Array[Int] }
fn build() -> Wrap {
  region r {
    let values = MutList::empty(r)
    let wrapped = Wrap::{ f: () -> Array[Int] { MutList::to_array(values) } }
    wrapped
  }
}
```

**2. `bash scripts/compiler_gate.sh` does not pass on this branch's HEAD**, and not because of anything in the region work. `check_fixture_execution.sh` (#1587) accounts for a fixture by grepping `scripts/` for its basename, and the `for r90_escape in outer_assignment container ...` loop builds paths from stems, so those names never appear literally. Eight of the new fixtures are reported as `has test blocks but no lane runs it`, and the gate exits there — before any region check runs. Reproduced on a clean checkout of `e76784ebd` with no other changes. Listing the fixtures by full path fixes it (#1993 does that).

---

mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
@codex review

mizchi:
Required CI is still the selfcompile heap KPI (`heap_ptr_bytes` ~1.17 GiB vs +10% gate). Unit shards are green.

Not addressing the Codex P1/P2 hamster wheel in this pass — those are more provenance holes on the same #1938 surface, stacked into #1993. Heap needs its own cut, not another rider.

mizchi:
@codex review

mizchi:
@codex review

PR #1972, opened 2026-08-16
Return non-zero status for uncaught exceptions

### LINKED ISSUE #1945: Return a non-zero process status for uncaught Exception values
## Bug

A program whose entry point declares `Exception` and throws an uncaught value prints `vibe: uncaught error: ...` but `vibe run` exits with status 0. CI and shell scripts therefore treat a failed program as successful.

## Acceptance criteria

- Red: an uncaught Exception prints a diagnostic and exits 0 on the old runtime
- Green: uncaught Exception exits non-zero through every launcher/runtime lane
- Handled exceptions and successful programs still exit 0
- Explicit process-exit values preserve their requested status
- Linear and wasm-gc/component launchers have parity
- Add a shell-level regression that checks both output and exit status

Extracted from the tutorial DX investigation in #1826.

mizchi:
## Summary

- move the uncaught-exception entry boundary into the shared normalization path so linear and wasm-gc backends agree
- diagnose an uncaught value and trap instead of returning a value that `_start` discards
- keep handled exceptions successful and preserve explicit process-exit statuses
- make basic async components valid when the boundary adds the stderr import
- add a shell-level regression covering output and status across linear, wasm-gc, and component launchers

## Validation

- `pkf run generation-gate`
- `pkf run test-affected -- --changed-from origin/main` (665/665)
- `pkf run test-uncaught-exception-exit`
- `pkf run test-async-component`
- `pkf run check-host-runtime-contract`
- `pkf run pre-commit`
- `pkf run check-taskfile-fmt`
- `scripts/check_vibe_fmt.sh` with the bundled modern Bash
- `pkf run release-check` passed the changed exception gates and continued through compiler gate 91; gate 92 hit the existing macOS command-size limit: `xargs: command line cannot be assembled, too long`

Closes #1945

### DIFF
--- a/Taskfile.pkl
+++ b/Taskfile.pkl
@@ -454,6 +454,8 @@ local testMeasureFsHeap = scriptTask("test-measure-fs-heap", "scripts/measure_fs
 local testRunCachedVibe = scriptTask("test-run-cached-vibe", "scripts/run_cached_vibe_test.sh")
 local testSelfhostGenerations = scriptTask("test-generations", "scripts/generations_test.sh")
 local testVibeRun = scriptTask("test-vibe-run", "scripts/vibe_run_smoke.sh")
+local testUncaughtExceptionExit =
+  scriptTask("test-uncaught-exception-exit", "scripts/test_uncaught_exception_exit.sh")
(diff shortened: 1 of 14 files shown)


mizchi:
@codex review

PR #1992, opened 2026-08-16
fix: single-file Show interpolation and block-local optional params

### LINKED ISSUE #1948: Make single-file check enforce Show interpolation like full check
## Bug

Full `vibe check` rejects interpolation of a value without a Show renderer, but `vibe check --single-file` can return empty output and exit 0 for the same source. The editor-oriented path therefore certifies invalid code as clean.

## Acceptance criteria

- Red fixture demonstrates full-check reject versus single-file false clean
- Both paths run the same interpolation/Show eligibility contract
- Diagnostics identify `derive(Show)` or `Type::to_string` as the fixes
- Imported renderers follow the documented single-file limitation without silently accepting unknown evidence
- CLI, LSP diagnostics, and JSON output agree

Extracted from #1826 and related to #1445.

### LINKED ISSUE #1952: An optional parameter `b?` on a block-local lambda is rejected: `function arity mismatch` (top level works)
A function declaring an optional parameter can be called with that argument omitted — at the top level. Bind the same function with `let` inside a block and the call no longer compiles.

## Symptom

```vibe
fn top_level(a: Int, b?: Int) -> Int {
  match b { Some(v) => a + v, None => a }
}

test "optional param" {
  let o = (a: Int, b?: Int) -> Int {
    match b { Some(v) => a + v, None => a }
  }
  inspect(o(7), "7")          // <- rejected
  inspect(top_level(7), "7")  // <- fine
}
```

```
FAIL (compile) opt.vibe
       opt.vibe: line 15:11-12: function arity mismatch for o: expected 2 args, got 1
```

The same declaration at the top level compiles and runs: `top_level(7)` = 7, `top_level(7, 5)` = 12 (measured).

## Cause

`desugar_labeled_args` fills in the omitted `None` for optional parameters, driven by a table built by `collect_optional_param_fns` (`lib/@vibe/compiler/normalize/desugar_trait_dict.vibe:9206`). That table walks the **top-level** statement array only — its own doc comment says "Every **top-level** function that declares at least one optional (`z?`) parameter" — matching `SLet(_, _, fname, _, EFn(...))` without recursing into block bodies.

Then the `ECall` arm skips the fill for anything locally bound (`desugar_trait_dict.vibe:9604`):

```vibe
let opt_args = match callee {
  EIdent(fname, _) => if str_array_contains(shadows, fname) {
    final_args                      // <- a block-local binding is always in `shadows`
  } else {
    match optional_param_flags(fname) {
      Some(flags) => optional_fill_args(final_args, flags),
      None => final_args
    }
  },
  _ => final_args
}
```

The bail-out is right as far as it goes — do not apply the top-level table to a local binding of the same name — but there is no local table to consult instead, so a block-local optional parameter gets no fill at all. The call reaches the checker as a 1-argument call against a 2-parameter function, and the checker reports exactly that.

## Relation to #1899

Same code, same arm, same `shadows` short-circuit: the labeled-argument reordering immediately above (`desugar_trait_dict.vibe:9585-9598`, `lookup_param_labels` + `reorder_labeled_args`) bails out for the same reason and produces #1899's still-open second half. Three tables are top-level-only and a block-local lambda is invisible to all three: `collect_fn_param_labels`, `collect_optional_param_fns`, and the checker's equivalent.

They differ in **how they fail**, which is why they are triaged differently:

| | symptom | triage |
|---|---|---|
| #1899 (2nd half) | labels silently applied in written order — `s(y = 3, x = 10)` on `x - y` returns -7 instead of 7 | **P0** (silently wrong) |
| this | rejected at compile time with `line:col` naming the call | **P1** (can't write it) |

A fix that gives block-local lambdas a scoped table would very likely resolve both. Worth attempting together; I reverted a first attempt at the #1899 half because the implementation reached the built compiler and provably did not fire (see PR #1925), so the shape of the local table is still unestablished.

## Measured with

`bootstrap/seed/compiler.wasm` sha256 `bfd18afa…` (seed `simd-line-comment-2026-08-16`), linear backend, `scripts/vibe_test.sh`.

mizchi:
Fixes #1948.
Fixes #1952.

## #1948 (P0)

`vibe check --single-file` used `collect_all_diagnostics` and stopped after `check_program`. Full `vibe check` then runs `expand_interp_stmts`, which rejects interpolating a declared aggregate with no Show renderer. Same source was therefore clean in the editor lane and rejected in the compile lane.

`collect_all_diagnostics` now runs the same `expand_interp_stmts` pass after typecheck.

## #1952

`collect_optional_param_fns` only walked top-level `fn`, and the `ECall` arm skipped any name in `shadows`. A block-local `let o = (a: Int, b?: Int) -> ...` therefore kept arity 2 and `o(7)` failed.

Local `EFn` flags are pushed on `ELet`/`ELetRec`/`ELetMut` and consulted before the top-level table.

## Tests

- `lib/@vibe/compiler/tests/interp_show_single_file_diag_test.vibe`
- `lib/@vibe/compiler/tests/optional_param_local_lambda_test.vibe`
- `fixtures/optional_param_test.vibe` (end-to-end once stage2 is rebuilt)

### DIFF
--- a/fixtures/optional_param_test.vibe
+++ b/fixtures/optional_param_test.vibe
@@ -44,3 +44,14 @@ test "a partial suffix of optionals is filled" {
   assert(two(1, 2) == 3)
   assert(two(1, 2, 3) == 6)
 }
+
+test "a block-local lambda optional may be omitted" {
+  let o = (a: Int, b?: Int) -> Int {
+    match b {
+      Some(v) => a + v,
+      None => a
+    }
+  }
+  assert(o(7) == 7)
+  assert(o(7, 5) == 12)
+}
--- a/lib/@vibe/compiler/normalize/desugar_trait_dict.vibe
+++ b/lib/@vibe/compiler/normalize/desugar_trait_dict.vibe
@@ -9370,6 +9370,86 @@ fn optional_param_flags(fname: String) -> Option[Array[Bool]] {
   }
 }
 
+// #1952: block-local `let o = (a, b?) -> ...` is in `shadows`, so the
+// top-level table must not apply — but the local function still needs fill.
+// Innermost binding wins; leave pops with the matching `shadows` mark.
+let optional_param_local_cell: Array[(String, Array[Bool])] = array_empty()
+
+fn efn_optional_flags(e: Expr) -> Option[Array[Bool]] {
+  match e {
+    EFn(_, _, params, _, _, _) => {
+      let flags = array_empty()
+      let mut any = false
+      let mut pi = 0
+      while pi < Array::length(params) {
+        let (pn, _) = Array::get(params, pi)
+        let pl = String::length(pn)
+        let is_opt = pl > 0 && pn[pl - 1: pl] == "?"
+        if is_opt {
+          any = true
+        } else {
+          ()
+        }
+        Array::push(flags, is_opt)
+        pi = pi + 1
+      }
+      if any {
+        Some(flags)
+      } else {
+        None
+      }
+    },
+    _ => None
+  }
+}
+
+fn optional_local_enter(name: String, v: Expr) -> Int {
+  let mark = Array::length(optional_param_local_cell)
+  match efn_optional_flags(v) {
+    Some(flags) => Array::push(optional_param_local_cell, (name, flags)),
+    None => ()
+  }
+  mark
+}
+
+fn optional_local_leave(mark: Int) -> Unit {
+  Array::truncate(optional_param_local_cell, mark)
+}
+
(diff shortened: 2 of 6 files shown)


mizchi:
Addressed the Codex review on nearer shadows: `b89ad7bef`.

Every nearer binder (non-optional `let`, match/for/loop/param, and `let mut` assign) now pushes a sentinel onto the local optional stack so `o(7)` is not filled from an outer `(a, b?)`. Regression is in `optional_param_local_lambda_test.vibe`.

PR #1993, opened 2026-08-16
Refuse region provenance erasure at the carrier (#1938)

mizchi:
**Stacked on `codex/fix-1938` (#1971)** — it needs that branch's provenance rider, so it targets that branch rather than `main`. Follow-up to the witnesses in https://github.com/mizchi/vibe-lang/pull/1971#issuecomment-5307431555.

## What was open

#1971 carries region provenance as a rider on the checked function row. Every carrier that can *name* what it holds keeps it — tuple, array, `Option`, generic enum. A carrier declared **without type parameters** has nowhere to put it, so the rider is dropped and the value stops being trackable.

Measured on `e76784ebd`, every one of these compiled clean:

| route | enum carrier | struct carrier |
|---|---|---|
| literal directly in the region's tail | ❌ accepted | ✅ rejected |
| `let` alias | ❌ accepted | ❌ accepted |
| inside an array | ❌ accepted | ❌ accepted |
| assigned to a binding that predates the region | ❌ accepted | — |

The struct spelling was rejected in exactly one position because the fallback that catches it (`tail_named_record_mentions_region`) inspects that one syntactic shape — which is the literal-form dependence #1938's criteria rule out.

## The rule

Reject the **erasure**, not the escape. At an enum constructor call or a named struct literal, if the carrier declares no type parameters and a payload depends on the active region, the construction fails:

```
line 12:7: `Box` has no type parameter to record the region this value depends on,
so storing it here erases the dependency: give the carrier a type parameter
(like `MutList[T, r]`), or copy the value out first (`MutList::to_array`)
```

One rule replaces rediscovering the dependency at every later site — and there is no later site where it is still visible.

**This over-approximates on purpose**: a carrier that never leaves the region is refused too. The two ways out are in the message, and writing the field *after* construction stays legal — which is what the existing positive fixture already does. A parameterized carrier is accepted, pinned as a new positive case.

## Known residue (not closed here)

A helper that constructs the carrier internally still escapes:

```vibe
fn mk(f: () -> Array[Int]) -> Wrap { Wrap::{ f: f } }
region r { let v = MutList::empty(r); mk(() -> Array[Int] { MutList::to_array(v) }) }
```

No region is active where the construction happens, and the nominal result type carries nothing back to the call site. Closing it needs provenance on the result type itself — the representation change #1938 describes, and a decision for whoever owns the provenance model rather than something to slip in here.

## Gate

Six new witnesses (enum payload / alias / container / outer-assignment, struct alias / container) plus the moved `named_struct` case, each required to fail with the erasure diagnostic **and** a `line:col` (#1567).

## Also fixed: the region fixture list in `compiler_gate.sh`

The loop built fixture paths from stems, and `check_fixture_execution.sh` (#1587) accounts for a fixture by grepping `scripts/` for its **basename** — which an interpolated name never produces. Eight of #1971's fixtures were reported as run by no lane, so **`bash scripts/compiler_gate.sh` fails on `codex/fix-1938` HEAD for that reason alone**, before reaching any region check. The list is now explicit paths.

## Verification

- `bash scripts/compiler_gate.sh` — green end to end (region gate, new erasure gate, fixture-execution check)
- `bash scripts/generations.sh build` — green
- every non-`err_` region fixture in the corpus still compiles: **0 false positives**
- `pkf run pre-commit`, `vibe fmt --check` — green

https://claude.ai/code/session_01TwospM2pqZv8ifD8tHRbNV

---
_Generated by [Claude Code](https://claude.ai/code/session_01TwospM2pqZv8ifD8tHRbNV)_

### DIFF
--- a/docs/region-mutable-state.md
+++ b/docs/region-mutable-state.md
@@ -235,22 +235,19 @@ region の外に出られない。検査は TaskGroup の雛形を一般化す�
   (`fixtures/region_arena_ok.vibe`、42)と negative
   (`fixtures/err_region_escape_return_value.vibe`、needle
   "region escapes its scope")を固定。
-- **closure capture (#1725、2026-08-14)**: §4 の「捕獲した closure が region
-  型に感染する」は **row 統合前提なので Phase 1 では働かない** —— 型は捕獲を
-  記録しないので、`region r { let l = MutList::empty(r); () -> Array[Int] {
-  MutList::to_array(l) } }` の結果型は `() -> Array[Int]` で skolem 走査から
-  見えず、region 値が closure の環境に隠れて脱出できた。arena が入れば
-  use-after-free なので stopgap を入れてある:
-  `checker/checker_escape.vibe :: region_token_escapes_in_closure` が
-  「region body の**結果そのもの**が closure literal で、region token から
-  直に束縛された名前(とその別名)を捕獲している」形だけを reject する。
-  **capture 自体は正当**(region 内で完結する closure は普通に書く)なので
-  過剰approx にしないことが要件で、汚染は「token に言及する初期化式」と
-  「汚染名の裸の別名」に限り、shadowing は最後の束縛が勝つ。取りこぼし:
-  outer binding / container 経由、helper 関数経由。gate 75 が両方向
-  (`fixtures/err_region_escape_closure_capture.vibe` /
-  `fixtures/region_ok_closure_local.vibe`)を pin。**正しい規則は §4 の
-  row 感染**で、arena スライスの前提はそちら。
+- **Closure capture provenance (#1725, #1938)**: a checked function row carries
+  an internal, source-unforgeable region provenance label. A direct capture is
+  `#region_n`; a capture through an inference variable is `#capture[v]` and
+  resolves when ordinary type substitution binds `v`. These labels are not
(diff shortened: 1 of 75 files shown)


mizchi:
CI still red: `err_region_escape_named_projection.vibe` does not emit `region escapes its scope`, and `type_db_cross_module_test` (`imported helpers retain region dependency contracts`) traps.

That is the published-env hole: `__field_contract__*` is not on the export surface, so imported `unwrap` loses the field dependency. Not pushing a speculative restrict_env change from this pass — it did not turn the type_db test green locally. Codex review nits stay open on the same surface.

mizchi:
Superseded by #1971, which now includes these carrier witnesses under the explicit provenance model and has merged as 9072eaa33. The immediate-construction rejection here is intentionally not retained because it rejects region-local values that never escape; #1971 instead rejects at the actual escape boundary. All relevant negative and positive regressions pass on merged main.

mizchi:
Left closed. The close reason stands: #1971 already landed the escape-boundary model, and this PR's construction-time rejection is the one that was intentionally not kept.

I had started a rebase onto current main to clear the stack conflict; that is **not** a reopen. The branch `claude/region-carrier-erasure-1938` was force-pushed during that attempt (`ac062fa45`). Ignore it unless someone explicitly wants the erasure-at-construction follow-up back.

PR #2003, opened 2026-08-16
fix(checker): mark lowered let* with __try_none, not source shape

mizchi:
Follow-up to #1999 (already merged).

#1999 stopped classifying unlocated `match o { Some(v) => ..., None => None }`
as `let*` by requiring matching source offsets. That closed the
self-compile / `compile_source` hole, but it also meant a real `let*`
on the unlocated `parse_program` path lost its source-level diagnostic.

Codex review on #1999 asked for an origin marker, the same kind `__try_v`
is for postfix `?`, instead of treating the source shape as railway syntax.

- `lower_try_bind` emits `EIdent("__try_none")` on the failure arm
- `railway_kind_from_match` looks only at that name
- `__try_none` types as the prelude `None` builtin
- a user-written `None` stays an ordinary match, located or not
- unlocated `let*` still names `let*` through `first_check_error`

Verified: `checker_railway_diag_test.vibe` through the pinned seed.


mizchi:
Follow-up `3fc873d22`: `__try_none` is a checker-only origin marker. Typing remapped it to `None`, but the AST still carried the synthetic name, so codegen ICE'd (`reached code generation unresolved`) on `let*` (`pipeline_ergonomics_test`, `try_let_star_option_test`).

After a clean check, `erase_railway_origin_markers` rewrites `__try_none` back to `None`. Diagnostics still see the marker; user-written `None => None` is still not classified as `let*`.

mizchi:
Follow-up `21b6be230`: CI still ICE'd on `__try_none` after the checker erase. Check mutates its own statement array; the FS/merged compile path can still see the marker.

Codegen now resolves leftover `__try_none` as `None`, and both `compile_module` and the WASI merge entry erase again.

Issue #1946, opened 2026-08-16
Make vibe test failures show the full test name and expected/actual values


mizchi:
## Problem

`vibe test` truncates a quoted test name at the first space and `assert_eq` failures do not show expected and actual values. Documentation also disagrees about whether a directory can be passed to `vibe test`.

## Scope

- Preserve the complete source test name in failure output
- Show expected and actual rendered values for assert_eq
- Keep output line-oriented and deterministic
- Either support directory inputs or make all docs and diagnostics consistently point to the supported command

## Acceptance criteria

- Failing named tests with spaces, Unicode, and duplicate names
- Scalar, String, tuple, struct, and unrenderable assert_eq values
- Exit status remains non-zero for any failure
- Tutorial, cheatsheet, and CLI help agree

Extracted from #1826.

mizchi:
The quoted-name half landed in #1995 (`failing test: <full name>` from the Node/wasmtime stack frame).

Still open: `assert_eq` expected/actual rendering, and whether `vibe test` accepts a directory. Not closing this until those are done or split out.

mizchi:
Quoted-name half landed in #1995 (`failing test: <full name>`).

Leftover (not closing):
- `assert_eq` expected/actual rendering for scalar, String, tuple, struct, and unrenderable values
- whether `vibe test` accepts a directory, or docs/help consistently say it does not

mizchi:
Leftover slice: #2010 prints `expected:` / `actual:` on `assert_eq` failure and documents that `vibe test` accepts a directory of `*_test.vibe` files.

Quoted names remain #1995. If anything is still missing after #2010 lands, it is `runtime/vibe` `condense_test_trap` still using the old `__test_[A-Za-z0-9_]+` cut — user-facing `scripts/vibe_test.sh` is the new one.

mizchi:
Leftover slice after #1995 / #2010: `runtime/vibe` `condense_test_trap` still cut `__test_[A-Za-z0-9_]+` at the first non-word character, and neither condenser decoded percent-encoded frames (`__test_has%20spaces`).

#2015 ports the frame-delimiter extract to `condense_test_trap`, percent-decodes the displayed name in both copies, and extends `scripts/vibe_test_smoke.sh` so a live `test "has spaces"` plus canned `%20` dumps must print `failing test: has spaces`.

[...]

PR #2010, opened 2026-08-16
fix(test): print expected and actual on assert_eq failure

mizchi:
Leftover #1946 (quoted names already landed in #1995).

`assert_eq` failures now print both sides before trapping:

```
assert_eq failed
  expected: 2
  actual:   1
```

The rewrite lives in `desugar_trait_dicts` (after check, same pass as throw-site rendering), so both backends see ordinary `==` / `Stderr::write_stream` / `assert_true` calls. Scalars and `String` go through `__to_string`; tuples and `derive(Show)` structs reuse the existing renderer; anything else is `<unprintable>`. The diagnostic is on stderr because `vibe test` discards stdout. `vt_fail_detail` / `condense_test_trap` keep those lines on FAIL.

`vibe test` now accepts a directory and expands `*_test.vibe` under it, matching `scripts/vibe_test.sh`. Tutorial, install help, and `runtime/vibe` usage say the same thing.

## Test

`lib/@vibe/compiler/tests/assert_eq_failure_test.vibe` drives the shipped `compile_source_wasi` path (not a reimplementation):

- compile-time: scalar / String / tuple / Show struct / unrenderable all emit the diagnostic labels
- run-time: compile a failing `assert_eq(1, 2)`, execute it, and require stderr to contain `expected: 2` and `actual:   1`

`scripts/vibe_test_smoke.sh` also locks canned `vt_fail_detail` output and directory input.

No seed wasm.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -109,7 +109,8 @@ vibe run     <file.vibex> [-- args]   compile the fixed `main` entry then run
 vibe compile <file.vibe> -o <out>     compile to a .wasm
 vibe build   <file.vibe> -o <out>     alias of compile
 vibe check   <file.vibe|file.vibex>   parse + typecheck (no output kept)
-vibe test    <file_test.vibe>...      compile + run test {} blocks
+vibe test    <file_test.vibe|dir>...  compile + run test {} blocks
+                                      (a directory expands to *_test.vibe)
 vibe fetch   [project_dir]            vendor git/URL deps from vibe.deps + lock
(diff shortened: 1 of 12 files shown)


mizchi:
gc-gate fix `18b9f97a8`: `Stderr::write_stream` flipped the gc host-import block on SIMD-only files and shifted generated-body indices (`simd_skip_ws_test` failed to compile). Diagnostics now go through `println`; `vibe test` keeps stdout on FAIL so `expected:`/`actual:` still surface.

Issue #2012, opened 2026-08-16
Assigning a lambda to a closure-captured `let mut`, then calling it, traps


mizchi:
A `let mut` holding a function, reassigned from inside a closure, traps when it is next called.

## Symptom

```vibe
fn captured_mut_lambda() -> Int {
  let mut g = (x: Int, y: Int) -> Int { x - y }
  let bump = () -> Unit { g = (x: Int, y: Int) -> Int { x - y - 1 } }
  bump()
  g(10, 3)
}
```

Expected 6. Compiles clean, then at run time:

```
RuntimeError: table index is out of bounds
[crash debug] heap_ptr=376 (0x178), memory_size=4194304 (64 pages) / table index is out of bounds
```

## Scope

Nothing to do with labeled arguments — the repro above has none. Both lambdas are ordinary two-parameter functions with the same shape; only the reassignment site is unusual. Assigning the same lambda **outside** a closure works:

```vibe
let mut g = (x: Int, y: Int) -> Int { x - y }
g = (x: Int, y: Int) -> Int { x - y - 1 }
g(10, 3)      // 6, correct
```

So the escape/box path for a captured `let mut` (#1262, ADR-0091) appears not to survive holding a function value: the call reads a function index out of a cell the closure wrote, and lands outside the table.

## Provenance

Found while measuring #1994 (labeled arguments on a `let mut` lambda). It is not caused by that work: the **committed seed** — which contains none of it — traps identically on the program above, byte for byte the same `heap_ptr`. Reported separately because it is a different failure in a different pass.

## Why P1, not P0

It crashes rather than returning a plausible wrong number, so it cannot be mistaken for a correct run ([docs/issue-triage.md](docs/issue-triage.md)). No `line:col`, though, and the message says nothing about what in the program caused it — worth fixing along with the trap itself.

mizchi:
Claiming this. Isolated PR incoming: captured `let mut` holding a function, reassigned from a closure, then called.

mizchi:
PR: https://github.com/mizchi/vibe-lang/pull/2014 — load the boxed cell before `call_indirect`.

PR #2031, opened 2026-08-16
Drop a `let mut` lambda's labels when a branch re-labels it (#1994 follow-up)

mizchi:
Follow-up to #2011, which landed the sequential half of #1994. This is the part that half leaves open, narrowed to where it actually bites.

## What is still wrong on `main`

Continuation-only replacement is exact while the assignment **dominates** what follows — straight-line code, which is what `labeled_local_lambda_scope_test` pins. It is exact nowhere else. The walk enters an `if` arm once; execution may enter it never. A loop body may run any number of times. A closure body runs somewhere else, if ever.

```vibe
fn branch_stale(flip: Bool) -> Int {
  let mut g = (x~: Int, y~: Int) -> Int { x - y }
  if flip {
    g = (y~: Int, x~: Int) -> Int { x - y }
  } else { () }
  g(x = 10, y = 3)
}
```

`branch_stale(false)` returns **-7**; 7 is correct. After the `if`, the table still holds the arm's `(y, x)`, so the call is reordered by labels belonging to a lambda `g` does not hold. Measured on a stage2 built from `126db0e07` (the #2011 merge), linear backend.

The reporting side is also unreachable there: nothing pushes to `unstable_mut_tbl_cell`, so `unstable_mut_tbl_lookup`, the `ECall` record and the checker's message can never fire.

## The rule

Leaving a subtree whose walk does not dominate what follows drops any entry whose labels an assignment inside it **changed**, and the binding goes back to unresolved. A labeled call to an unresolved binding is then reported with a location instead of compiled.

An assignment that re-binds the **same** parameter names changes nothing a label can observe, so it is not recorded and costs the binding nothing. That is what keeps the common shapes resolving, and it is why this does not repeat the flow-insensitive reject that broke the sequential fixtures earlier in #2011.

```
line 6:3-4: cannot resolve labeled arguments for g: a branch, loop or closure
assigns it a lambda with different parameter names, so a label does not name a
position here — call g positionally, or give every lambda assigned to g the
same parameter names
```

## Measurement

Through a stage2 built from this checkout, against one built from `main`:

| shape | main | here |
|---|---:|---|
| straight-line reassign (the #2011 case) | 7 | 7 — unchanged |
| branch assigns the same parameter names | 7 | 7 |
| branch assigns different names, call after | **-7** | **reported**, with `line:col` |
| call *before* that branch | 7 | 7 — not a false rejection |
| positional call | correct | correct |

Both directions are covered by tests: `fixtures/labeled_mut_lambda_scope_test.vibe` for the shapes that must keep resolving (branch, loop body, and closure body assignments that keep the names; a call the branch has not reached yet), and `lib/@vibe/compiler/tests/labeled_mut_lambda_unstable_test.vibe` for the report and its three non-reports.

Everything in the runtime fixture subtracts on purpose: `x + y` gives the right answer whether or not the arguments were reordered, which is how the first half of #1899 survived a "fix verified" claim.

## Regressions checked

`fixtures/labeled_local_lambda_scope_test.vibe` (sequential updates) and `lib/@vibe/compiler/tests/normalize_labeled_arg_binder_authority_test.vibe` both pass — those are exactly what the earlier reject broke. `fixtures/captured_mut_lambda_test.vibe` (#2012) passes too, since a closure body is one of the subtrees this touches.

`scripts/compiler_gate.sh` and `scripts/check_vibe_fmt.sh` are running against this base; I will confirm both in a comment. (`pkf run test` cannot reach the gate in this container — a broken `/usr/bin/sort` under pkfire's nix environment, reproducible on a clean `main`.)

https://claude.ai/code/session_016v6eS7sAtn3hh16x61Spqk

---
_Generated by [Claude Code](https://claude.ai/code/session_016v6eS7sAtn3hh16x61Spqk)_

### DIFF
--- a/fixtures/labeled_mut_lambda_scope_test.vibe
+++ b/fixtures/labeled_mut_lambda_scope_test.vibe
@@ -1,51 +1,27 @@
-// #1994: a lambda bound by `let mut` got no parameter table, so a labeled call
-// to it was never reordered -- the arguments bound in WRITTEN order, with no
-// diagnostic. `let mut g = (x~: Int, y~: Int) -> Int { x - y }` called as
-// `g(y = 3, x = 10)` returned -7.
+// #1994, second half. Registering a `let mut` lambda from its initializer and
+// replacing the entry on assignment (the first half) is exact while the
+// assignment DOMINATES what follows -- straight-line code. It is exact nowhere
+// else: the walk goes through an `if` arm once, execution may go through it
+// never; a loop body may run any number of times; a closure body runs
+// somewhere else entirely. The entry such a subtree leaves behind then
+// describes a lambda the binding may not hold, and the labeled call after it
+// was reordered by those stale names, with no diagnostic.
 //
-// #1899/#1952 gave block-local `let` bindings a table of their own but left the
-// mutable case out on purpose: a `let mut` can be reassigned to another
-// type-compatible lambda whose parameter NAMES differ (labels are not part of a
-// function's type), and a table built from the initializer would then reorder a
-// later call by stale names.
+// Leaving one of those subtrees now drops the entry again, so the binding goes
+// back to unresolved -- and a labeled call to an unresolved binding is
+// reported rather than compiled (labeled_mut_lambda_unstable_test.vibe).
 //
-// The rule now is flow-insensitive and needs no analysis: register the
-// initializer's labels when the whole scope AGREES on them -- every assignment
-// to the name binds a lambda declaring the same parameter names in the same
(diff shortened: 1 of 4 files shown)


mizchi:
Gates confirmed, as promised in the description:

- `scripts/compiler_gate.sh` — exit 0, `[compiler-gate] ok` (104/104)
- `scripts/check_vibe_fmt.sh` — 971 files under `lib`, 0 known debt, 0 new violations

CI agrees so far: every check on this head is green except `compiler-gate`, which is still running.

---

PR #2058, opened 2026-08-17
fix(check): warn on old builtin spellings via a deprecation table

mizchi:
## Why

Old builtin spellings (`String::char_code_at`, PascalCase `Fs::ReadFile`, …) never got a `#deprecated` warning because builtins have no declaration site. `canonical_builtin_name` already promised `core/builtin_deprecated.vibe`; the file did not exist. That is why those names keep spreading.

## What

- Add `builtin_deprecated.vibe` and honour it from `vibe check` (always on, not import-gated).
- Warn `String::char_code_at` → `String::byte_at`, `String::from_char_code` → `String::from_byte`, identity `Char::to_int`/`from_int`, and PascalCase Fs/Env names.
- Leave `*Builder::freeze` off the table until a bootstrap bump (seed does not know `build`; compiler source must keep `freeze`).
- Type aliases in `index.vpkg` (`type HashMap = MutMap`) still cannot carry `#deprecated` (vpkg parse). Function aliases (`HashMap::*`) were already marked.

## Test

`bash scripts/vibe_test.sh lib/@vibe/compiler/runtime/deprecated_scan_test.vibe` — 19 passed.


mizchi:
Leftover, not in this PR:

- `StringBuilder::freeze` / `ArrayBuilder::freeze` / `MapBuilder::freeze` stay unmarked until a bootstrap bump (seed does not know `build`).
- Collection **type** names (`HashMap` etc.) cannot carry `#deprecated` in `index.vpkg` until the same bump — the seed contract parser rejects `#`. That slice is #2059 (language table, visibility-gated).
- `Profiler::NowUs` / `Profiler::HeapBytes` and `Map::has` / `map_has` are still accepted aliases and are not in this table yet.

PR #2059, opened 2026-08-17
fix(check): warn on old collection type aliases

mizchi:
## Why

Old collection **type** names (`HashMap`, `HashSet`, `SortedMap`, `SortedSet`, `ImmutMap`) are unmarked. Function aliases already warn; annotations do not. That is why the old spelling keeps spreading.

Putting `#deprecated` on `index.vpkg` is the honest site, but the **seed** contract parser rejects `#`, and the seed loads `@vibe/core/index.vpkg` when compiling the compiler. Marking the file today breaks every compile.

## What

- Language table in `deprecated_scan.vibe` for those five type names, **visibility-gated** (a file that did not import `HashMap` is not told to rename its own type).
- `parse_contract_program` skips `#deprecated` so a later bootstrap bump can move the marker onto the vpkg row.
- `vibe check` re-reads on-disk `.vpkg` files (the generated facade drops the directive).
- Cheatsheet updated.

`*Builder::freeze` stays unmarked until the seed knows `build`.

## Tests

- `deprecated_scan_test.vibe`: table warns; silent when the type was not imported; warns when it was.
- `contract_conformance_test.vibe`: contract parser accepts `#deprecated` / bare form, rejects unknown `#`.
- `cli_support_test.vibe`: facade-in-source_groups still warns from the on-disk vpkg.

84 tests in those three files passed via `scripts/vibe_test.sh`.

### DIFF
--- a/docs/cheatsheet.md
+++ b/docs/cheatsheet.md
@@ -174,8 +174,9 @@ contract — this is a naming *rule*, not a per-type coincidence:
 旧綴りの**型注釈**も transparent alias として残る (#1700)。たとえば
 `let m: HashMap[String, Int] = HashMap::new_string()` は `@vibe/core` の
 `index.vpkg` 境界を越えて `MutMap[String, Int]` と同じ型になる。移行先は
-引き続き `Mut-` 名で、旧関数を使えば `vibe check` が警告する。型名だけを
-旧綴りにした場合は警告されないので、新規コードでは `Mut-` を直接書く。
+引き続き `Mut-` 名。型行に `#deprecated` を書くのは bootstrap bump 待ち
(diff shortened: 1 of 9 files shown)


mizchi:
Leftover, not in this PR:

- The `#deprecated` marker still cannot live on shipped `index.vpkg` type rows until a bootstrap bump (seed parse). This PR only makes the parser ready and warns via the language table.
- `*Builder::freeze` still waits on the seed knowing `build`.
- Next unmarked builtin aliases: `Profiler::NowUs` / `Profiler::HeapBytes`, `Map::has` / `map_has`.

mizchi:
@copilot resolve the merge conflicts in this pull request

Issue #2001, opened 2026-08-16
Inventory and retire the scripts layer: split compiler gates, migrate tools to .vibex, and move product behavior into @vibe/cli


mizchi:
## Goal

Reduce `scripts/` to thin host/platform orchestration.

Repository logic should be written in `.vibex` where the required capabilities exist. User-visible command semantics belong in `lib/@vibe/cli` or reusable `lib/@vibe/*` packages. Obsolete aliases, retired-host compatibility layers, historical repro scripts, and superseded implementations should be deleted.

This issue is an umbrella/index. Implementation belongs in bounded sub-issues.

## Current position (updated 2026-08-24, main 69874d2)

**Phase 0 and Phase 1 of the `compiler_gate.sh` refactor are done** (#2060, #2062): `scripts/compiler_gate.sh` is now an **84-line** thin aggregator (it was 12,338 lines at filing) over independently runnable lanes in `tests/gates/{bootstrap,early,mid,late,tooling-accounting}`, with the fail-closed registry at `tests/gates/registry.tsv`. Three of the four initial deletion candidates (`trial_gate.sh`, `coverage_eval_sidecar.sh`, `bench_cmd_latency.sh`) were deleted in #2062.

Inventory today: 238 tracked shell files, 69 JS/MJS/CJS, 13 `.vibex` under `scripts/`.

**Still open** (the live remainder of this epic):

- Phase 2 (shared assertion harness) and Phase 3 (`.vibex` migration of deterministic repository logic; product semantics into `@vibe/cli` — publish/install/cache-clean/normalize/run/test/bench policy still live in shell).
- The public installer consolidation (`install/install.sh`).
- Phase 4 deletion ratchet / `scripts/README.md` ownership metadata.
- The `allows` dogfood sequence below still waits on #1961.

### Directory policy: CI tests belong under `tests/`

Being invoked by CI does not by itself make a file repository infrastructure. If a script's primary purpose is to assert behavior, compile a test fixture, compare expected output, or implement a gate, it belongs under `tests/` rather than `scripts/`.

Target layout:

- `tests/gates/<lane>/`: independently runnable compiler/release gate specifications *(landed)*
- `tests/integration/`: CLI, package, component, WIT/WASI, DAP/LSP, and host integration tests
- `tests/fuzz/`: fuzz drivers, generators, reducers, oracle support, corpora, and active findings
- `tests/support/`: test-only shell/JS host adapters that cannot yet be `.vibex`
[...]

PR #2062, opened 2026-08-17
chore: delete dead script hops from the MoonBit-host era

mizchi:
#2001 listed these as initial deletion candidates. Each was a hop or a
broken alias, not a live tool.

- \`scripts/trial_gate.sh\` — exec'd \`compiler_gate.sh\`. \`gate.sh\` now does that itself.
- \`scripts/coverage_eval_sidecar.sh\` — unused; the task already called \`coverage_scratch_sidecar.sh\`.
- \`scripts/bench_cmd_latency.sh\` + \`pkf run bench-cmd-latency\` — exec'd missing \`bench_compare.sh\`.

Docs that named the hops now name \`compiler_gate.sh\`.


mizchi:
Leftover from #2001: repro_715_rc_free_list_corruption.sh still has a unique diagnostic procedure. Not deleted.

PR #2071, opened 2026-08-17
refactor(normalize): extract desugar_array_spread from the 12k kitchen-sink file (#1849)

### LINKED ISSUE #1849: Split the compiler's oversized files (inline_direct_perform 15.0k / desugar_trait_dict 12.8k / gates/late 5.9k)
Three files were named when this was filed. Re-measured 2026-08-19 — **one is
done, and the other two grew.**

| file | at filing | now | |
| --- | ---: | ---: | --- |
| `lib/@vibe/compiler/codegen/common_base/inline_direct_perform.vibe` | 14.3k | **15,017** | grew; also not under `normalize/` as the old title said |
| `lib/@vibe/compiler/normalize/desugar_trait_dict.vibe` | 10.5k | **12,804** | grew ~22% |
| `scripts/compiler_gate.sh` | 11.8k | **84** | ✅ **split** |

`compiler_gate.sh` is now an 84-line dispatcher over four lanes:

```
tests/gates/bootstrap/run.sh    176
tests/gates/early/run.sh      3,851
tests/gates/mid/run.sh        2,764
tests/gates/late/run.sh       5,941
```

So that third of the issue is closed, and the successor to watch is
`tests/gates/late/run.sh` at 5.9k — well below where it started but the largest
shell file left.

The rest of the tree, for scale (everything over 5k lines, generated bundles
excluded):

```
15017  lib/@vibe/compiler/codegen/common_base/inline_direct_perform.vibe
12804  lib/@vibe/compiler/normalize/desugar_trait_dict.vibe
 9232  lib/@vibe/compiler/checker/checker.vibe
 8590  lib/@vibe/compiler/codegen/wasi/linked_compile.vibe
 8446  lib/@vibe/compiler/entry/source_compile/wasi_only/component_codegen.vibe
 5941  tests/gates/late/run.sh
 5430  lib/@vibe/compiler/codegen/expr/compile_call.vibe
 5223  lib/@vibe/compiler/runtime/typecheck_fs.vibe
 5194  lib/@vibe/optimizer/wasm_opt.vibe
```

## Why the two that remain keep growing

Both are the file every new case lands in, and each new case is individually
small and correct to put there — so the file grows without any single change
looking wrong. `desugar_trait_dict.vibe` gained ~150 lines in #2128 alone
(tuple shape resolution), and that change was the right one; it just had nowhere
smaller to go.

That makes "split it" the wrong framing on its own. What each file needs first
is a **seam**: for `desugar_trait_dict.vibe` the shape/`var_types` machinery
(`infer_shape`, `shape_bind`, `seed_var_types`, the `*_key` sentinels and their
readers) is a self-contained sub-language with its own grammar, used by dict
synthesis, interpolation rendering, `==` dispatch, and now `<` — four consumers
that only touch it through lookups. That is a package boundary, not a file
boundary.

Splitting on line count without finding the seam produces files that import
each other's internals, which is worse than one long file.

mizchi:
One bounded slice of #1849: split a real pass out of `desugar_trait_dict.vibe` (12,708 lines).

## Why not `desugar_derives`

That was the first target. Its generators (`gen_eq_fn`, `show_for_typed`, `collect_bare_array_eq_keys`, …) call helpers the `==` rewrite still owns (`eq_for_typed`, `type_key`, `arr_equals_name`, `is_scalar_type_name`, `concat2`). Normalize siblings do **not** share private names. Importing the helpers from the kitchen-sink file while that file imports the new pass is a cycle; promoting the helpers to `export fn` without a vpkg row is a contract violation.

`desugar_hoist_local_lambdas` has the same shape: `dlh_marker_only_called_directly` calls `dinsp_scan`, and the hoist walk calls `collect_pat_binders`.

`desugar_array_spread` is the next self-contained large-enough pass. It only needs `Expr`/`Pat`/`Stmt` and `array_empty`.

## What moved

- New `lib/@vibe/compiler/normalize/desugar_array_spread.vibe` (`das_expr` / `das_arms` / `das_stmt` / `desugar_array_spread`).
- `desugar_trait_dict.vibe` imports and re-exports the pass so `import ./desugar_trait_dict.vibe { desugar_array_spread }` still resolves.
- Manifest row next to `desugar_trait_dict.vibe`. `index.vpkg` already declared the fn.

## Line counts

| file | before | after |
| --- | ---: | ---: |
| `desugar_trait_dict.vibe` | 12708 | 12558 (−150) |
| `desugar_array_spread.vibe` | — | 180 |

## Tests (seed compiler, `VIBE_FS_COMPILE=1`)

Red: renaming the export to `desugar_array_spread_hidden` makes the package test fail (`imported name 'desugar_array_spread' is not exported`).

Green:

- `lib/@vibe/compiler/normalize/desugar_array_spread_test.vibe` (no-op + `...xs` → `Array::concat`)
- `lib/@vibe/compiler/normalize/desugar_derives_test.vibe` (package still exports `desugar_derives`)
- `lib/@vibe/compiler/normalize/uniquify_shadowed_bindings_test.vibe`
- `fixtures/derive_eq_test.vibe`

Does not close #1849 (`inline_direct_perform.vibe` is still 14k; this is one file).

### DIFF
--- a/lib/@vibe/compiler/compiler_sources_manifest.tsv
+++ b/lib/@vibe/compiler/compiler_sources_manifest.tsv
@@ -271,6 +271,7 @@ loader	loader/loader.vibe
 normalize	normalize/index.vpkg
 normalize	normalize/normalize.vibe
 normalize	normalize/desugar_trait_dict.vibe
+normalize	normalize/desugar_array_spread.vibe
 normalize	normalize/unbox_tuple_loop.vibe
 normalize	codegen/common_base/self_tail_call.vibe
 normalize	codegen/common_base/inline_direct_perform.vibe
--- a/lib/@vibe/compiler/normalize/desugar_array_spread.vibe
+++ b/lib/@vibe/compiler/normalize/desugar_array_spread.vibe
@@ -1,0 +1,180 @@
+// Array-literal spread lowering (#790, #1849).
+// Extracted from desugar_trait_dict.vibe. `desugar_derives` and
+// `desugar_hoist_local_lambdas` both call into helpers the kitchen-sink
+// file still owns (`eq_for_typed`, `dinsp_scan`, `collect_pat_binders`),
+// so pulling either would cycle or duplicate. This pass only needs the
+// AST types and `array_empty`.
+
+//# Imports
+
+import @vibe/compiler/core {
+  Expr, Pat, Stmt
+}
+
+import @vibe/core {
+  array_empty
+}
+
+//# Functions
+
+///|
+/// #790: lower array-literal spreads `[a, ...xs, b]` into an `Array::concat` chain
+/// of segments (`Array::concat(Array::concat([a], xs), [b])`). The parser accepts
+/// `...expr` inside an array literal (ESpread), but the active desugar pipeline
+/// never lowered it, so codegen reached ESpread and threw "unsupported expression
+/// in codegen". Ported from the legacy desugar.vibe (spread-only; no other legacy
+/// transforms). `das_arms` is mutually recursive with `das_expr`.
+fn das_expr(expr: Expr) -> Expr {
+  match expr {
+    EArray(elems) => {
+      let mut has_spread = false
+      let mut si0 = 0
+      while si0 < Array::length(elems) {
+        match Array::get(elems, si0) {
+          ESpread(_) => {
+            has_spread = true
+          },
+          _ => ()
+        }
+        si0 = si0 + 1
+      }
+      if has_spread {
+        let segments = array_empty()
+        let mut buf = array_empty()
+        let mut si = 0
+        while si < Array::length(elems) {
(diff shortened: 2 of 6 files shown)


mizchi:
Not merging. `compiler-gate` failed the selfcompile heap KPI (#987):

- heap_ptr_bytes **1,165,652,320**
- gate_max **1,164,658,668** (+10% of baseline)
- delta about **+993 KiB**

Adding `normalize/desugar_array_spread.vibe` as a new compiler module is the same class of cost as #2058 (a new compiler file is not free even when the moved code is small). Folding back into `desugar_trait_dict.vibe` would undo the split. #1849 stays open; the next cut needs to shrink an existing module without adding a new bundle file, or pay for the heap on purpose after a measured win.

Leaving this PR closed so it does not keep burning CI.

Issue #1727, opened 2026-08-14
async: suspend split の残り 2 件は「call-edge instantiation が AST に無い」に帰着する (#1536 の後継)


mizchi:
#1536 の残作業の切り出し。あちらの本文が名指ししていた症状 (`async_iter_*` /
`await(Stream::next(s))` / ループ / `for` / closure param callee) は**すべて解消**した
([実測](https://github.com/mizchi/vibe-lang/issues/1536#issuecomment-5291742673))。
残る 2 件は**同じ 1 つの不足**に帰着するので、そちらを表題にして起票し直す。

## 共通の不足

suspend CPS split は AST しか見ておらず、**呼び出し辺ごとの型/row の instantiation**
を持っていない。残る 2 件はどちらも「その辺で何が入るか」が分かれば決まる。

## 残り 1: 関数を取る row 変数 callee

```vibe
fn twice[e](f: (x: Int) -> Int with e, v: Int) -> Int with e { f(v) + f(v) }
// suspend-class handle body の中:
let b = twice((x: Int) -> Int { x + 1 }, a)
```

**REFUSED** (`cannot see through`)。

引数が row 変数 `e` を instantiate しうるので、「宣言 row が空 ⇒ perform しない」が
成り立たない — **これは健全性の線であって穴ではない**。#1705 が「宣言パラメータが
関数型を一切含まない callee」(= 引数が row を運べない) だけを受理したのはそのため。

受理を広げるには、この呼び出し辺で `e` に何が入るかが要る。fixture:
`fixtures/err_effect_rowvar_hof_call.vibe`。

## 残り 2: 局所束縛経由の `for` iterand

```vibe
let mk = () -> Array[Int] { [1, 2] }
for j in mk() { acc = acc + perform Sus::Suspend(j) }
```

**REFUSED**。pass は「`mk` が inert な closure を保持している」ことは見えている
(だから呼び出し自体は受理される) が、**種別は証明しない** — #1714 の guard が
「局所が隠している名前の宣言は信じない」と決めているため。

その guard は必要 (外すと String を Array として index して黙って 0 を返す、#1714 実測)。
広げるなら「局所が何を保持しているか」を辺ごとに追う必要があり、上と同じ不足。
`cli_support_test.vibe` の不適格テストがこの形を留めている。

## 判断

**どちらも「落ちる」ではなく「書けない」で、回避策がある** (row 変数 callee は
[...]

PR #2081, opened 2026-08-17
fix(checker)+feat(async): close the row-variable effect leak (#2078), then close both #1727 eligibility gaps

### LINKED ISSUE #2078: An annotated closure literal passed to a row-VARIABLE parameter loses its effect requirement — compiles, then traps unhandled
## Symptom

A program that performs an effect with **no handler anywhere and no declared row** type-checks, compiles, and throws an unhandled effect at runtime.

```vibe
effect Ask {
  Get(Int) -> Int
}

fn eff_use(g: (Int) -> Int with e, x: Int) -> Int with e {
  g(x)
}

fn leaky() -> Int {
  eff_use((y: Int) -> Int with Ask {
    perform Ask::Get(y)
  }, 2)
}

fn main() -> Int {
  leaky()
}
```

```console
$ ... cli_main stage2.wasm repro.vibe repro.wasm main
(no diagnostic, wasm emitted)

$ ... --invoke main repro.wasm
[crash debug] heap_ptr=320 (0x140), memory_size=4194304 (64 pages) / [object WebAssembly.Exception]
[object WebAssembly.Exception]
```

`leaky()` declares no row, nothing handles `Ask`, and the checker accepts it. P0 by the triage rule: it is silently wrong, not a crash at the right place.

## Root cause, isolated to one token

Two ingredients are BOTH required. Each of the six neighbours below differs from the repro on exactly one axis.

| # | argument at the parameter | parameter's row | caller declares row | result |
|---|---|---|---|---|
| **repro** | **inline literal, `with Ask`** | **variable `e`** | no | **COMPILED → runtime trap** |
| c8 | inline literal, **no annotation** | variable `e` | no | REJECTED — `missing { Ask::Get }` |
| c7 | **named binding**, `with Ask` | variable `e` | no | REJECTED — `missing { Ask }` |
| c4 | inline literal, `with Ask` | **concrete `with Ask`** | no | REJECTED — `missing { Ask }` |
| c5 | inline literal, pure | variable `e` | no | COMPILED (correct — `e := {}`) |
| c3 | direct `perform`, no HOF | — | no | REJECTED — `missing { Ask::Get }` |

So the leak needs an **annotated inline closure literal** *and* a **row-variable parameter**. Remove the annotation (c8) and the `perform` is charged lexically to `leaky`, exactly as #1261 describes the model:

> closure リテラル内の `perform` は **リテラルが置かれた関数の row** に計上される（呼び出しサイトではなく字句的）

Hoist the same annotated literal to a named binding (c7) and the row-variable path picks it up, exactly as #1485's fix intends. Make the parameter's row concrete (c4) and it is caught too.

The annotation is what makes the row fall between the two mechanisms: the literal declaring its own row suppresses the lexical charge to the enclosing function, and the row-variable instantiation does not pick it up from an inline literal.

## This is NOT a regression of #1485

#1485's own acceptance example — the performing function as a **named** value — is still correctly rejected:

```vibe
let boom2 = (x: Int) -> Int with Ask { perform Ask::Get(x) }
let run2 = (x: Int) -> Int { apply_one(boom2, x) }
```
```
line 14:3: effect row mismatch for 'run2': missing { Ask }
```

and its fixtures are still promoted in `fixtures/typecheck/expected.tsv`:

```
generic_effect_matrix_fail_missing_effect_at_caller  reject  ...
generic_effect_missing_caller_effect                 reject  ...
generic_effect_matrix_ok_passthrough_pure            ok
```

This is the uncovered **intersection** of #1485 (row-variable parameter, named argument — fixed) and #1261 (closure literal, row-free parameter — fixed): closure literal **and** row-variable parameter, which neither covered.

## Why it matters beyond one trap

#1485 already recorded the consequence, and it applies verbatim here:

- ADR-0084's entry-row check (the host preflighting `main`'s residual row) never sees a capability that entered this way
- ADR-0088's `allows` likewise starts from the declared row, so an undeclared one is out of scope

An `Fs`/`Env` capability substituted for `Ask` in the repro is the same program.

## A note for anyone fixing it

#1485 records that the naive fix does not work — requiring the row-variable **token** in the caller's declared row produces false positives, with `fixtures/typecheck/generic_effect_matrix_ok_passthrough_pure.vibe` as the standing proof. c5 above is the same constraint in this issue's shape: an inline literal with a *pure* row must stay accepted.

The narrow reading is that an annotated closure literal in argument position should contribute its declared row to the instantiation the same way a named binding already does (c7 works), rather than being treated as self-discharging.

## Acceptance

- [ ] the repro fails to compile, naming `Ask` on `leaky`
- [ ] c5 (pure inline literal through a row variable) still compiles — no false positive
- [ ] c4 / c7 / c8 keep their current diagnostics
- [ ] `generic_effect_matrix_ok_passthrough_pure` stays `ok`

## Provenance

Found while measuring #1727 (whether the checker solves row variables per call edge — it does; `e := {}` for a pure argument, `e := Ask` for a performing one). Compiler: generation `stage2` at `6d1e38f72`; reproduced through `VIBE_FS_COMPILE=1` single-file compile.

Related: #1485 (closed, named-argument half), #1261 (closed, row-free-parameter half), #885, #1727.

mizchi:
Three changes on one branch, in the order they have to land. Closes #2078; closes **both** of #1727's remaining gaps.

All three are the same subject: **what the pass can prove about a value at the place it actually arrives**, rather than from a declaration read somewhere else. The first makes the checker stop *missing* an effect; the other two make the suspend lowering stop *refusing* code where nothing can go wrong.

---

# 1. `fix(checker)`: an annotated closure literal must instantiate a row variable (#2078)

A program that performs an effect with **no handler and no declared row anywhere** type-checked, compiled, and threw an unhandled effect at runtime:

```vibe
effect Ask { Get(Int) -> Int }
fn eff_use(g: (Int) -> Int with e, x: Int) -> Int with e { g(x) }
fn leaky() -> Int {
  eff_use((y: Int) -> Int with Ask { perform Ask::Get(y) }, 2)
}
```

```console
$ ... cli_main stage2.wasm repro.vibe repro.wasm main
(no diagnostic, wasm emitted)
$ ... --invoke main repro.wasm
[object WebAssembly.Exception]
```

`resolve_row_var_instantiation` (from #1485) only handled `EIdent` arguments; everything else bailed and kept the pre-#1485 exemption. Its doc comment listed *"a lambda literal"* among the bail-outs, on the principle that the pass never guesses. For a literal that **declares** its row there is nothing to guess — read it. An **unannotated** literal still bails: its row is `None`, which `core/types.vibe` defines as structurally distinct from `Some([])` (*"inferred and therefore inherits its surrounding context"*), and that case is already covered lexically.

| argument | parameter's row | before | after |
|---|---|---|---|
| **inline literal, `with Ask`** | **variable `e`** | **COMPILED** | **REJECTED** |
| inline literal, unannotated | variable `e` | REJECTED | REJECTED |
| named binding, `with Ask` | variable `e` | REJECTED | REJECTED |
| inline literal, `with Ask` | concrete `with Ask` | REJECTED | REJECTED |
| inline literal, pure | variable `e` | COMPILED | COMPILED |
| direct `perform`, no HOF | — | REJECTED | REJECTED |

**Not a regression of #1485** — its own acceptance example (the performing function as a *named* value) is still rejected, and `generic_effect_matrix_ok_passthrough_pure` still compiles. This is the uncovered intersection of #1485 (row-variable parameter, named argument) and #1261 (closure literal, row-free parameter).

Fixtures use a **user effect**, not `Exception`: the exception row is transparent on the transitive path, so an `Exception` version would pass for a reason unrelated to what it tests. I wrote it that way first and it silently proved nothing.

---

# 2. `feat(async)`: a row-variable HOF call whose function arguments are provably inert (#1727 gap 1)

```vibe
fn apply_twice(f: (Int) -> Int with e, x: Int) -> Int with e { f(f(x)) }
// inside a suspend-class handle body:
let b = apply_twice((y: Int) -> Int { y * 2 }, a)
```

`scps_callee_first_order` (#1705) reads the **callee alone**: it takes a function, therefore an argument *could* instantiate `e` to the handled effect. Whether it can is a property of what arrives — and the pass already has `scps_is_inert_literal` (used by the `ELet` arm) proving a literal has no `eff` in its row, no row variable of its own, is not a step literal, and no tainted call in its body.

| argument at the function-typed parameter | before | after |
|---|---|---|
| **inline literal, pure** | **REFUSED** | **COMPILES → 21** |
| inline literal declaring `with Ask` | REFUSED | REFUSED |
| unannotated literal that performs | REFUSED | REFUSED |
| NAME bound to a pure function | REFUSED | REFUSED |

**No type table, substitution, or offset key is involved.** I had measured toward a checker→codegen pipeline for this ([notes on #1727](https://github.com/mizchi/vibe-lang/issues/1727#issuecomment-5315343978)); gap 1 needed none of it.

---

# 3. `feat(async)`: a `for` iterand proved from its local binding (#1727 gap 2)

```vibe
let mk = () -> Array[Int] { [1, 2] }
for j in mk() { acc = acc + perform Sus::Suspend(j) }
```

`scps_call_ret_ty` answered `None` for any callee a local binder hides, because #1714 measured that believing the shadowed **top-level** annotation lowered a String iterand to the indexed Array form — `Array::get` reading a `(offset << 32) | length` fat pointer as a heap pointer. That guard is about reading the *wrong declaration*, not about locals: a local literal declares its return type right at the binding.

Admitted only when nothing is left to choose between: the name is **not also a top-level fn**, it has exactly **one** binder in the walked expression, and that binder is an immutable `let` holding a closure literal with a declared return type.

| iterand | before | after |
|---|---|---|
| **call through a declared local `let`** | **REFUSED** | **COMPILES → 23** |
| same, local literal with no declared return | REFUSED | REFUSED |
| same name also a top-level fn (#1714 shape) | REFUSED | REFUSED |
| two live binders of the name | REFUSED | REFUSED |

**The top-level clause is the one the gate taught me.** My own bracket passed the shadowing case — it produced 215, correct String semantics — so I thought reading the local was simply better. `err_effect_for_shadowed_callee.vibe` disagreed and was right for a reason my cases did not cover: the binder probe is scope-blind, so a lone local binder in a branch that does **not** reach the call would be read while the call resolves to the top-level fn. That over-accepts in exactly the direction that fixture says this pass must never err in.

`cli_support_test.vibe`'s ineligibility test pinned this shape as *"the cost of that guard"*. The cost is gone, so it now pins the boundary that survives: a local literal with no declared return type.

---

## Why this order

#2078 says one path silently produces *no* effect requirement. Parts 2 and 3 admit more code on the strength of what the pass can prove locally. Widening before plugging would have let the leak reach further.

## Verification

- `bash scripts/compiler_gate.sh` → **exit 0** after each commit separately
- typecheck-verdict lane: `typecheck fixtures ok (88 fixtures, 3 known debt)`; exactly one verdict changed in the `generic_effect_*` family
- `effect_rowvar_first_order_call_test.vibe` unchanged; `err_effect_for_shadowed_callee.vibe` still rejected
- every new fixture verified under the gate harness's own invocation
- `scripts/vibe_fmt.sh --check` clean; `pkf run pre-commit` passed
- re-verified after merging main three times, as the checker subsystem moved underneath this branch (one of those merges touched `checker_effects.vibe` itself)

## New fixtures

```
generic_effect_missing_caller_effect_inline_lambda   reject  (was: compiled)
generic_effect_ok_inline_lambda_pure                 ok      (false-positive guard)
effect_rowvar_hof_inert_literal_test                 21      (gap 1 positive)
effect_for_local_binding_iterand_suspend_test        23      (gap 2 positive)
err_effect_rowvar_hof_call                           repurposed to the NAME argument
```

## Known adjacent limitation, unchanged

Following #2078's new diagnostic — declare the row up the chain and handle it at the entry — still meets the ADR-0084 entry rule. Verified the old compiler behaves identically; parts 2 and 3 remove two of the reasons a handled version was refused.

https://claude.ai/code/session_01TwospM2pqZv8ifD8tHRbNV


mizchi:
`smoke (ubuntu-latest / profile)` failed on `72c941a9` for an infrastructure reason, not this change — it never reached a build step:

```
Download action repository 'Swatinem/rust-cache@v2' ...
##[warning]Failed to download action ... Error: Response status code does not indicate success: 429 (Too Many Requests). Back off 13.495 seconds before retry.
##[warning]Failed to download action ... 429 (Too Many Requests). Back off 11.869 seconds before retry.
##[error]Failed to download archive '.../Swatinem/rust-cache/...' after 3 attempts.
```

The four sibling smoke lanes (`js`, `lsp`, `library`, `debug`) pull the same action on the same run and all passed, so it is a transient rate limit on that one runner. Everything else on this SHA is green or still running: `compiler-gate` early/mid/late, all four `unit-tests` shards, `gc-gate`, `vibe-fmt-check`, `structural-lint`, `perf-metrics`.

**A re-run of that one job should clear it.** I do not have permission to trigger it (`rerun-failed-jobs` → 403, the token here is read-only), so it needs a click.

For context on why the head moved: main gained six commits after the last green run, four of them checker changes and one touching `checker_effects.vibe` itself — the file part 1 edits. I merged main in and re-verified locally rather than trusting the older base: `bash scripts/compiler_gate.sh` exit 0, both behaviour pins re-checked against the new checker (`effect_rowvar_hof_inert_literal_test` → 21, `effect_rowvar_first_order_call_test` unchanged, and #2078's repro still rejected with `missing { Ask }`).

---

Issue #2105, opened 2026-08-18
Rename @vibe/prelude to @vibe/builtin


mizchi:
## Symptom / discomfort

The package that holds `Int` helpers, `Option` combinators, and `stdout_write` is named `@vibe/prelude`. That word is a Haskell (and Scala) import: "the module the compiler sneaks into every file."

In this repo it does **not** mean that. `lib/@vibe/prelude/index.vpkg` is explicit: prelude is **not** wired into the loader as a magic always-imported package. Every user file that wants those names writes `import @vibe/prelude { … }`. What looks like an implicit prelude is compiler builtins (`Option`, some `Int`/`String` methods) plus inspect desugar — none of which read this package's facade.

Teaching `@vibe/prelude` therefore imports the wrong metaphor. `@vibe/builtin` (or another boring name) matches what the directory actually is: the language's own standard helpers.

## Current surface (measured)

- Path: `lib/@vibe/prelude/`, contract `name = @vibe/prelude`.
- Facade file is `prelude.vibe` (renamed from `index.vibe` so it can sit next to `index.vpkg`).
- README / ADR-0005 still say `vibe/prelude` and layered `trait-contract` / `pure-primitive` / `effect-boundary`.
- Cheatsheet hello: `import @vibe/prelude { stdout_write }`.
- I/O helpers in that package (`io.vibe`) are the wrong neighbors for `Int::abs` if the stream-effect sibling lands as `@vibe/console`.

## Proposal (not a grant)

Rename the package to `@vibe/builtin` (directory `lib/@vibe/builtin/`, `name = @vibe/builtin`). Keep a temporary re-export facade at `@vibe/prelude` only if import-graph churn is too large for one PR.

Do **not** treat this as a grant to auto-import the package. The current "explicit import, no magic" rule stays unless a sibling I/O issue separately decides to inject `println`.

Coordinate with:

- Hello / `println` / library-izing `Stdout` — those may move `io.vibe` *out* of this package. Rename after that cut, or the move and the rename fight.
- #1353 (`lib` package hierarchy / `@vibe` promotion). That issue is graph/cache/concurrent, **not** this rename. Cross-link only.

## Related (not duplicates)

- Sibling cluster: hello availability; `println` spelling; `Stdout` out of the language namespace.
- #1353 package hierarchy (different files, different goal).

## Labels

`refactoring` · `P2` · `lane:stdlib` · not `blocker`. Behavior-preserving rename plus docs.

## Suggested first slice

[...]

Issue #2101, opened 2026-08-18
Prefer a println-like user-facing I/O name over stdout_write


mizchi:
## Symptom / discomfort

The taught hello verb is `stdout_write`. That name is an fd-shaped implementation detail (`Stdout::write_stream` without a trailing newline). A more general user-facing name would look like `println` / `print`.

`stdout_writeln` exists in the same prelude file and *does* add `\\n`. The cheatsheet still shows `stdout_write("hello world\\n")` with a manual newline. That is the wrong default for the first program.

## Current surface (measured)

| Name | Where | What it does |
|---|---|---|
| `stdout_write` | `@vibe/prelude` `io.vibe` | `Stdout::write_stream` — no newline |
| `stdout_writeln` | same | write + `write_char(10)` |
| `print` / `println` | `@vibe/console` `io.vibe` | same pair, already spelled like every other language |
| `eprint` / `eprintln` | `@vibe/console` | `Stderr` counterparts |
| `inspect` desugar | compiler | expands to a `println` call |

So the name the language already uses internally and in `@vibe/console` is `println`. The name the docs teach is `stdout_write`.

"Effect name" here means the **user-facing I/O verb** (and, if we keep a capability row, the row/op it sits on). It does **not** automatically mean a new algebraic effect called `Println`. Capability ops stay `Effect::snake_case` (ADR-0084 / cheatsheet taxonomy). A plausible landing is `Console::println` / a builtin `println` with `with Console`, not `perform Println::…`.

## Proposal (not a grant)

Pick one taught spelling for "emit a line":

- Prefer `println` (and `print` without newline) over `stdout_write` / `stdout_writeln`.
- Point hello, cheatsheet, and book at that spelling.
- Deprecate or stop re-exporting `stdout_write` after the docs move.

Coordinate with the sibling issues: availability (builtin vs import) and whether `Stdout` remains a top-level language name. `@vibe/console` already has the names; this issue is about making them *the* names.

## Related (not duplicates)

- Sibling cluster: hello without a prelude import; keep `Stdout` out of the language namespace; `@vibe/prelude` rename.
- #1460 `Stdout` → `Console` (capability label). Naming the **function** `println` is independent of that merge, but the row on the signature should follow #1460.
- Closed #929 was codegen for a historic builtin `println`/`print`, not this docs/stdlib surface.

## Labels

`enhancement` · `P2` · not `blocker`.

## Suggested first slice

[...]

Issue #2100, opened 2026-08-18
Hello I/O should not require importing stdout_write from @vibe/prelude


mizchi:
## Symptom / discomfort

The first program in `docs/cheatsheet.md` is:

```vibe
import @vibe/prelude { stdout_write }

fn main with Stdout {
  stdout_write("hello world\n")
}
```

The helper is documented as *not* a builtin (`unknown function: String::stdout_write` if you forget the import). That is a surprising tax for "print a line": the name lives in a Haskell-flavored stdlib package, while `Stdout` itself is already an ambient language-level capability.

## Current surface (measured)

- `lib/@vibe/prelude/io.vibe` `stdout_write` is a one-line wrapper around `Stdout::write_stream`.
- `lib/@vibe/console/io.vibe` already exports `print` / `println` wrapping the same host ops (and `eprint` / `eprintln` for `Stderr`).
- `inspect(...)` desugars to `__to_string` / `println` / `assert_true` *before* import resolution (`desugar_inspect_calls`), so a `println` name already exists on the compiler path — just not as the documented hello verb.
- `@vibe/prelude` is **not** auto-imported (`index.vpkg` header): every caller that wants prelude names imports them explicitly.

## Proposal (not a grant)

Make the hello print verb available without `import @vibe/prelude { stdout_write }`. Candidates:

1. A real compiler builtin (same class as `Option` / the inspect desugar target).
2. Implicit injection of a small I/O name, still implemented in a library.
3. Teach hello to use `@vibe/console { println }` instead of prelude — still an import, but the package name matches the job.

This issue is only the *availability* question. Spelling is sibling; whether `Stdout` stays a top-level language name is sibling. Do not close those by picking (1) here.

## Related (not duplicates)

- Sibling cluster from a docs reread: naming (`println` vs `stdout_write`), library-izing stream effects, `@vibe/prelude` → `@vibe/builtin`.
- #1460 unifies `Stdin`/`Stdout`/`Stderr` into `Console` as a **language-level** capability. That is a rename/merge, not "hello should not import a helper".
- #1458 effect taxonomy / operation spelling.

## Labels

`enhancement` · `P2` · not `blocker`. Current hello compiles; this is surface honesty, not silent-wrong.

## Suggested first slice

Decide the availability rule in writing (builtin vs injected vs documented `@vibe/console` import) against the two sibling issues, then change cheatsheet + book + one fixture. Do not land a new ambient name while #1460 still has `Stdout` as the taught row.

[...]

Issue #1949, opened 2026-08-16
Make the first tutorial program work with an installed toolchain outside the repository


mizchi:
## Problem

The first tutorial program imports `@vibe/prelude` and works inside the repository, but an installed toolchain does not provide that package in `~/.vibe/lib`. Copying the example to an empty directory fails with a require-pin error, while install.md uses a different host-builtin hello that works.

## Decision needed

Choose one supported contract:

1. install the prelude with the toolchain,
2. change the first tutorial example to the host builtin form, or
3. include a complete package pin/setup path before the example.

## Acceptance criteria

- Fresh temporary directory with only the installed toolchain
- Copy/paste of the first documented program compiles and runs
- Tutorial, install guide, and package-pin diagnostic agree on the workflow
- Add an install-smoke regression independent of the repository `lib/` tree

Extracted from #1826.

mizchi:
Option 2 slice is in https://github.com/mizchi/vibe-lang/pull/2047.

The first documented program is now the host-builtin form from `docs/install.md`:

```
fn main with Stdout { Stdout::write_stream("42\n") }
```

Leftover: option 1 (install `@vibe/prelude` with the toolchain) is not done. A fresh install still cannot copy/paste the chapter-01 `import @vibe/prelude { stdout_write }` examples.

mizchi:
Leftover map (current):

**Shipped (option 2):** #2047. The first copy/paste program for an installed toolchain is the host builtin from `docs/install.md`:

```
fn main with Stdout { Stdout::write_stream("42\n") }
```

Tutorial README (en/ja) leads with that. Chapter 01 notes that `import @vibe/prelude { stdout_write }` needs the repo or a pin. `scripts/test_vibe_install_hello.sh` compiles that hello in a temp directory without repo `lib/` and asserts stdout `42`.

**Still open (not a short slice):** option 1 — install `@vibe/prelude` with the toolchain so chapter-01 prelude examples work in a fresh empty directory. That is an install/package-layout contract (`~/.vibe/lib`, pins, require-pin diagnostic), not another README edit.

mizchi:
Option 1 leftover is in https://github.com/mizchi/vibe-lang/pull/2056.

`install/install.sh` now materializes `@vibe/prelude` into `$VIBE_HOME/lib` with the other user-facing packages. After that install, this chapter-01 form runs in a temp directory with only the installed toolchain (repo `lib/` not on the search path):

```
import @vibe/prelude {
[...]

PR #2121, opened 2026-08-19
0.1.0 readiness: walked the newcomer path, then audited the references against the compiler

mizchi:
Working the 0.1.0 known-gaps list, then the `lane:stdlib` / `dx` / `refactoring` issues documentation-first. Every claim below was measured on the current compiler rather than read off a changelog. Each commit is independently reviewable.

Three issues close with this branch: **#2100, #2101, #2105** — all landed in this cluster, each closed with the measurement that shows it. Four filed from what turned up: **#2122, #2123, #2124**, plus a measured seven-claim table on #1943.

---

## 1. The documented first program was wrong — and its gate was green

Walked the README's install path on a real install into a scratch prefix. Nobody had measured "someone else's first hour", the bar 0.1.0 exists for.

```console
$ echo 'fn main with Stdout { Stdout::write_stream("42\\n") }' > hello.vibex
$ vibe run hello.vibex        # -> 42
42\n
```

Inside shell single quotes `\\n` stays **two characters**. `docs/install.md` carried the identical line.

**`scripts/test_vibe_install_hello.sh` (#1949) exists to run exactly this program** and passed throughout, because it *restated* the program rather than reading it — with `printf`, where `\\n` becomes a real newline:

| | writes | program the compiler sees |
|---|---|---|
| the gate | `printf '… "42\\n"'` | `"42\n"` → `42` + newline ✅ |
| README / install.md | `echo '… "42\\n"'` | `"42\\n"` → `42\n`, no newline ❌ |

The gate now **derives** the program and its expected output from the documents. Later in the branch it gained a fifth check — see §9.

**Verified on a real install** (exit 0): `vibe version` → `0.1.0-dev`; quickstart → `42`; README sample → clean check, prints `12`; `vibe new` scaffolds and runs.

## 2. Type errors had no source position

`let a: Int = "not an int"` reported with no `line:col` on **both** lanes, and `--single-file --json` answered a synthetic `0:0`. CLAUDE.md blamed the import lane's string diagnostics; measured, the cause is that the bound value is a **literal**, and literals carry no offset slot.

`ascribe_wrap` now puts the binder name's offset on the synthetic ascription call. Value offset still wins where it exists — this only supplies a position where there was none.

| | before | after |
|---|---|---|
| `vibe check` | *(none)* | `line 4:7:` |
| `--json` range | `0:0`, `synthetic:true` | `line 1, char 6–7`, `data:null` |

2988 tests across 251 compiler test files pass; all 88 `fixtures/typecheck` rows keep their verdicts.

## 3. `vibe symbols` returns doc comments

A **fifth** field, present only when the declaration has a doc, so four-field consumers are unaffected. Newline → `\n`, backslash doubled: one declaration, one line, reversibly.

## 4. Every navigation link in the Japanese book was broken

27 of 126 relative links did not resolve — including **every** chapter-to-chapter link in `book/ja/` and all 7 "English version" links, which were **self-links**. `scripts/check_book_links.sh` resolves them now (11 self-test cases). It skips fenced code, because `fn identity[T](x: T) -> T` matches the markdown link grammar.

## 5. The Japanese book rendered nowhere

`SUMMARY.md` listed only the canonical root, and `html_name_for` named pages by **basename** — so listing `book/ja/` would have silently overwritten the English pages one at a time. A translation root is now part of the name; every published English URL is byte-identical.

## 6. The remaining 12 chapters are translated

Chapters 08–19. The Japanese book covers **19 of 20** doctest-checked chapters. Both sides pass:

```
book/ja  59 blocks -- 49 pass, 0 fail, 10 skip
book/en  61 blocks -- 51 pass, 0 fail, 10 skip
```

Site: **42 pages, 23 English + 19 Japanese**, up from 22 and nothing. `book/src` → `book/en` for symmetry with `ja/`, with no URL moving.

## 7. Two documented gc limitations no longer exist

**`bench` runs on wasm-gc** (#1701 landed): linear 6820 ns/op, gc 5803. Really gc — the modules differ by 2311 bytes and report different per-op allocation.

**`mut` struct fields are not gc-only.** Mutation through a function, observed through an alias made before the write: `c=12 alias=12` on **both** lanes. Readers were avoiding a feature that works, and CLAUDE.md offered it as *the reason* to switch to gc.

## 8. `vibe grep`: an unlocated capture emitted `-1` where an offset goes

Measured #1943's seven claims. Three were already fixed (`--help`, `perform` vs `throw`, silent parse-skips). The capture claim was real and worse than stated: `"start":-1` is a **number**, so a consumer slicing `src[start:]` reads from the end of the file. It emits `null` now, and a test asserts `-1` is absent.

A second, previously undocumented finding: a capture's `text` is a **printed form, not a source slice** — `f( a  +  b )` captures as `(a + b)`. So `end` cannot be derived from it, which is *why* captures report no end. The two open claims are split to #2122 with exact diagnostics.

## 9. The language reference taught the legacy tty row

Three of four documents that teach a first program said `fn main with Console`. `docs/cheatsheet.md` said `fn main with Stdout` — in the one document AGENTS.md tells you to read first. Both compile, which is why it drifted.

The hello gate gained a fifth check: every document teaching a `fn main with <Row>` program must name the same row, and a document teaching none also fails. Red-proved.

## 10. Five references listed `Result` as a live type

#1324 removed it; `Ok(1)` is `unknown name: Ok`. The cheatsheet said so in one place and listed `Result` among builtins in four others; `docs/vibe.md` still had "error flow: `Result` composition" and "Result-first policy remains canonical". The distinction the checker already makes and the docs had lost: **`Option` is a builtin, `Result` is a user enum you may declare.**

`ADR-0018` ("migrate library APIs to Result", status `accepted`) is **deleted** per the log's own rule — the direction was later reversed. `ADR-0063/0065/0069` name `index.vibei` as the boundary; those decisions still hold, so they are repointed to `index.vpkg`, not deleted.

## 11. `String::replace` was listed as a builtin

The freeze list says it derives from the cheatsheet's Key Builtins. Probed both: they had drifted in **opposite** directions — the cheatsheet called `replace` a builtin (it is `unknown name` without an import) and the freeze list omitted `byte_at` / `from_byte` (real builtins, outside the SemVer promise). Both fixed; 41 frozen symbols resolve, up from 39.

---

## Two gates measured and deliberately not shipped

Both would have had false-positive rates that teach readers to ignore the checker:

- **A retired-name lint** over doc prose flags 39 lines, many correct — `## Result (N = 1000)` benchmark headings, "before" columns of migration tables, and book chapter 19's `not a Result wrapper`, the sentence that teaches it does not exist.
- **Extending the freeze probe to the cheatsheet** extracts 109 names, mostly nonsense (`Conversion::__add`, `Math::sh`, `Map::perform`) because that section mixes inline headings with prose and tables, so a receiver sticks across sub-sections.

Recorded in #2123 / #2124 with the fix that does work: one canonical statement, others link.

https://claude.ai/code/session_01GbtLER6wNT4jsZEN1Zp6xe


mizchi:
Continued past the 0.1.0 list into the `lane:stdlib` / `dx` / `refactoring` issues, documentation-first. Commits 12–15.

**The same shape kept turning up, three times:** one fact, several hand-written copies, and the gate that was supposed to check it reading its own copy instead of the document. Each is now closed structurally, not just corrected.

| fact | before | now |
|---|---|---|
| the taught first program | 4 documents, one on the legacy row; the gate *restated* the program | derived from the documents and run; all four must name the same entry row |
| the frozen builtin list | freeze list and cheatsheet each declared the other its source, and drifted in **opposite** directions | one probe reads both — 41 frozen + 31 indexed symbols |
| the `Send` allowlist | 4 documents, 3 drifted (`Option`/`Result` listed as if both were builtins) | behaviour pinned by 11 tests; prose canonical in `concurrency.md`, others link |

Closes #2124 and #2123 in addition to #2100/#2101/#2105.

### Two findings that only appear when you write the test

- **A value with an unresolved type parameter is not `Send`**, even when every resolved part is. Unannotated `Result::Ok(1)` gives ``no impl `Send` for `Result[Int, ?t8]` ``. No document said this, and "an enum of Send parts is Send" reads as sufficient.
- **A `vibe grep` capture's `text` is a printed form, not a source slice** — `f( a  +  b )` captures as `(a + b)`. So `end` cannot be derived from it, which is *why* captures report none.

### A correction to my own issue

I filed #2122 claiming `handle` patterns cannot be written at all. Wrong — they work with the arms written out; I had measured only the two failing shapes. The real gap is narrower: no metavariable kind for handler *arms*. #2122 is retitled and its remaining substance is that feature.

Fixed there meanwhile: a declaration pattern was rejected by `fn parameter '__vgm_args__p' requires a type annotation` — the metavariable's internal encoding, for a pattern the user *did* annotate. The good message existed but the parser rejected the declaration before it could run.

### Also deliberately not shipped

Two gates measured and dropped, both with false-positive rates that teach readers to ignore the checker: a retired-name lint (39 lines, many correct — benchmark headings, migration tables, and the sentence teaching that `Result` does not exist), and probing the cheatsheet's whole Key Builtins section (109 names, mostly nonsense). The second is what motivated restructuring that section into bullets so a precise probe *is* possible.

### Verification

`tests/gates/bootstrap/run.sh` green, `stage2==stage3 fixpoint holds`. 2999 compiler tests across 252 files. Both new grep diagnostics confirmed end-to-end through the built stage2.

---

mizchi:
Three more sections since the description was written. Same rule as the rest: measured on the current compiler, not read off a changelog.

## 12. 134 dead expectations became 134 running checks (#1571)

139 fixtures under `fixtures/` carried a `__DATA__` tail with an `error_contains` or `compile_error` string, and **nothing read it** — `scripts/generate_runtime_fixture_tests.mjs:273` skips exactly those two keys. Each was a hand-written expectation that had never been compared to a compiler, and several were stale by years: `raise` for `throw`, `Error` for `Exception`, `let rec`.

134 are now rows in `fixtures/typecheck/expected.tsv` — **223 rows, all green**, 28 known debt — measured against a stage2 and recorded at today's behaviour, with their tails removed. That harness already checked a verdict plus a message substring and already had the `debt-accepts` vocabulary.

Getting there needed two rewrites, and the rewrites are why the expectations had rotted:

- **62 rejected before reaching their own check.** ADR-0069 made a top-level expression illegal and most of these were written as one, so every single one answered *"top-level expressions are not allowed"* no matter what it contained. Wrapping the trailing expression in `fn main` turned 41 into the diagnostic they were named for.
- **10 were declaration-only**, so the compiler had nothing to compile and the check never ran. An empty `fn main` fixes that.

18 compile and are `debt-accepts` — checks this compiler does not make: `x += "hello"` on an `Int`, `apply(add, 42)` where `apply` wants a 1-argument function, `m[42] = 2` on a `Map[String, Int]`, `type T = T`, an inline `[T](x: T: Show)` bound, `let rec` on a non-function, a duplicate effect operation, an unknown ctor in a pattern, and 7 trait rules. 3 are `ok` on purpose: two assert a **runtime** error the type system never promised (their own comments say "(runtime error)").

Two were rewritten twice: #1877 widened the Int literal bound to 2^62-1, so `2305843009213693952` had quietly become legal and the two overflow fixtures were asserting an overflow that no longer happened.

The gate resolves a row name as `fixtures/typecheck/<name>` then `fixtures/<name>`, so no files moved. The last 5 are #2127 — pinning them would record a reason nobody meant.

## 13. A reserved word said nothing useful in five of six binder positions

`fn test() -> Int { 1 }` already named the word and the escape. The others did not — measured, same word, same file:

| written | said |
|---|---|
| `fn test() -> Int { 1 }` | `` `test` `` is reserved; write `` `r#test` `` |
| `let test = 1` | expected name after `'let'` |
| `fn main() { let test = 1 }` | expected identifier after `'let'` |
| `fn main() { let mut test = 1 }` | expected identifier after `'let mut'` |
| `fn main() { let* test = 1 }` | expected identifier after `'let*'` |
| `fn f(test: Int) -> Int { 1 }` | expected parameter name token=test |

Five name neither the problem nor the fix, and at a `let` the message is **actively misdirecting**: the name IS there, so it points at the one thing that was fine. `r#test` works in all six positions and nothing said so.

One helper answers all six now; `kind` is the only difference. `fn` keeps its own sentence because it is true — `r#fn` really is rejected, so offering it would be a lie. Red-proved with 4 new tests. The book chapter documented the old messages and said `r#fn` "is rejected the same way" without ever saying that every *other* reserved word takes `r#`.

## 14. Two compiler bugs found while fixing that, both filed with repros

**#2126 (P1) — a catch-all match arm's bound value is wrong.** These two are semantically identical, and one traps:

```
_     => throw(binder_name_error(tk_name(peek(tokens, pos)), "parameter", pos))  // ok
token => throw(binder_name_error(tk_name(token),             "parameter", pos))  // RuntimeError: unreachable
```

Six variants measured to isolate it: it is not the message, the helper, the block-vs-expression body, or the presence of a binder — it is **using the bound value**. Nothing rejects the broken form at compile time. A synthetic enum does not reproduce it at 4 or at 90 constructors, so the trigger needs something the synthetic case lacks. The six sites in this PR use the `_ =>` spelling with a comment pointing at the issue.

**#2125 (P0) — an unknown type name is a wildcard at argument positions.** `fn takes(x: Intt)` accepts both `takes(1)` and `takes("s")` in the same program, `vibe check` exit 0. Not a blanket wildcard, which is what makes it hard to notice: `x + 1` on an `Intt` **is** rejected, but `String::length(x)` is accepted. Root-caused to two sites that have to be read together — `checker_resolve.vibe:70` turns an unresolved name into `CtNamed(name, [])`, and `checker.vibe:2081` deliberately does not report a unify failure under a `CtNamed` head because that is also how an un-instantiated generic formal is spelled. `unify` itself is correct; the failure is discarded. The fix direction is to reject at resolution so the two stop sharing a representation. Four `debt-accepts` rows in this PR are the ratchet for it.

Also re-measured for #1941: the "literals carry no offset" hole is now **half** closed. A local `let` anchors on the binder (§2); a top-level `let` with a literal RHS is still unlocated, because `SLet` carries no offset and only the local lane goes through `ascribe_wrap`. Recorded there with the exact boundary.

---

`pkf run test` (the full operation gate) passes locally: exit 0, `typecheck fixtures ok (223 fixtures, 28 known debt)`.

---

mizchi:
Four more, continuing the same method — read the document, measure the claim, turn the answer into a gate.

## 15. Three benchmark files were guarding A-vs-B comparisons that nothing ran

`bench_map_vs_immutmap.vibe` says why its `test` blocks exist, right above them:

> `//# ---- agreement: 2 実装が同じ答えを出すことを固定する`
> `//# これが無いと「速いほうが実は違う仕事をしていた」を見逃す。`

Nothing ran them. `scripts/unit_test_runner.sh` discovers `*_test.vibe` under `examples/`/`lib/`/`fixtures/`; these are `bench_*.vibe` in `bench/`, so both filters missed them. 8 tests across 3 files, ~1s each through the runner's own path. A `test` block is the selector rather than the filename, so a bench file that makes no correctness claim costs the lane nothing. Red-proved.

## 16. "Small n favours the assoc list" stopped being true at n=64

Checking #1848's premise. The benchmark picked n=64 as its small-map data point on the stated grounds that **小さい n で assoc list が勝つのは既知**, so the only question worth measuring was real-application size. Measured (VIBE_RC=0, `--bench-count 20 --bench-warmup 3`, median of 3, ns/op):

| n | op | `Map` | `MapHamt` | ratio |
|---:|---|---:|---:|---|
| 8 | build | 1347 | 1393 | 0.97x (noise) |
| 64 | build | 3908 | 1778 | **2.2x** |
| 64 | lookup | 3886 | 2082 | **1.9x** |
| 1000 | build | 514892 | 21122 | **24.4x** |

`MapHamt` already wins at n=64 on both axes; they are indistinguishable at n=8. There is no n where the builtin meaningfully wins, so "the builtin `Map` is for small fixed tables" needs "small" read as single digits. n=8 benches added so the crossover is a measured point rather than a sentence. The n=1000 ratio (24.4x vs the recorded 27.7x) is the same conclusion on a different machine — the claim that changed is the direction at small n.

That matters for #1848 beyond the docs: it removes the main reason to leave a `Map::` site alone after inspecting it.

## 17. Two doctest harnesses answered differently on one document

`docs/cheatsheet.md` is clean under `pkf run doctest`. Handed to `scripts/vibe_md.sh check` — the command AGENTS.md points at — it said "30 pass, **3 fail**". All 3 were the two documented differences: `doctest_extract_run.sh` symlinks `lib` next to the extracted block (so `import ./lib/@vibe/core` resolves) and tolerates a declaration-only block; `vibe_md.sh` does neither.

They are not interchangeable by design — `vibe_md.sh` runs ` ```vibe run ` blocks and checks the paired ` ```output `, which only `*.vibe.md` carries. It now refuses anything else and names the other harness. Guarded for `write`/`fmt`/`fmt-check` too, since `fmt` would otherwise **rewrite** a document it does not understand. 5 self-test cases.

## 18. `docs/source-range-contract.md` had no gate, and its failure mode is silent

ADR-0108 put six editor-query surfaces on byte positions and the document wrote the table down. Nothing checked it, and by the document's own description a drift would be invisible:

> The commands do not error [...] Empty output then reads as "nothing here", which is the CLI's own spelling of clean.

`scripts/check_source_range_contract.sh` (`pkf run check-source-ranges`) probes 8 things against a source with two 4-byte emoji **to the left** of the position under test — byte column 40, codepoint 34, UTF-16 35, three numbers for one position. It pins that `symbols`/`binding-at`/`escapes` report byte offsets, that `type-at`/`binding-at` take a byte column, that a **codepoint** column resolves to nothing (so a surface silently starting to accept them fails this), that `vibe check` text is a 1-based byte column, and that the LSP JSON converts — and that its start is not the byte column, so the conversion cannot be dropped.

On an ASCII fixture all three units agree and every assertion is vacuous, so the check derives the three columns from its own fixture and refuses to run when they are not distinct — before resolving a compiler, since a vacuous fixture is a defect in the check. 4 self-test cases pin that, plus fail-closed on a missing or unusable compiler.

Also corrected: `cli_adapter.vibe` described `vibe symbols` as reporting "char offsets". They are bytes. The contract document said so all along; the comment next to the implementation did not.

## 19. Two more stale claims in AGENTS.md, one of them hiding a P0

**Uppercase imports.** AGENTS.md said `大文字名 (struct / type alias) だけは値環境が判定できず、未検出のまま (#1521 の残り半分)`. Measured: a non-exported struct, a non-exported type alias, and a type name that never existed all report "is not exported by". #1521 closed; its uppercase half closed with it. The gate's own case still carried the name it had when it *was* a gap — `bogus_uppercase_not_reported` — which is presumably how the claim outlived the fix. Renamed, and the two shapes it did not cover (a declared-but-unexported struct, ditto alias, with a `public_upper_ok` control) added.

**`String` `<`.** AGENTS.md said "MoonBit `String <` は length-first", told readers to hand-write a char-by-char comparator, and cited `src/codegen/wasm_codegen_data.mbt` — gone since #594. Measuring it turned up a live P0, **#2128**: `"buf" < "acc_bits"` is **false** from a plain `let` or an `Array[String]` element, and **true** from a tuple destructure, `.0`/`.1`, or an `Array[(String, String)]` element — even an annotated one. Same values, same operator, one program, no diagnostic. The split is type information: passing through `fn take(a: String, b: String)`, re-annotating, or laundering through `String::concat(a, "")` all restore the lexicographic answer.

Two compiler call sites already work around it (`module_diag_lt`, `plan_module_order`, both using `str_lt`) and `module_diag_lt`'s comment repeats the wrong reason, citing AGENTS.md. So the footgun was real and current, the explanation was neither, and nobody could tell which spellings were safe.

`string_lt_type_dependence_test.vibe` pins both halves (6 tests). The two length-first cases are labelled DELETE WHEN FIXED — they fail the day the split closes, which is the signal to promote rather than re-pin. Red-proved.

---

`pkf run test` passes locally on this tree: exit 0, `typecheck fixtures ok (223 fixtures, 28 known debt)`.

---

mizchi:
## Branch update — sections 12–15

Continuing the same method (read the document, measure the implementation, fix as a deterministic gate). Everything below is on top of the body above.

Additionally closed since the body was written: **#2130, #2126, #1941, #2132, #2122, #1571**. Additionally filed: **#2131, #2133, #2134, #2135, #2136**.

---

### 12. A catch-all match binder aliased the scrutinee without a dup (#2126, P0)

`token => tk_name(token)` trapped `unreachable` inside the callee where `_ => tk_name(peek(tokens, pos))` worked. Three theories were wrong before the real one: it is not the Perceus plan (that function has **zero** rows), and not `peek` dropping its container (`idx0=fn idx1=ident idx3=GARBAGE` — only the bound element is corrupt). The arm's binder took the scrutinee's slot without an RC dup, so the value was freed while still referenced. Fixed in `compile_match.vibe`; two regression tests run the recovering parser, which re-reads the token array and is what made the free observable.

### 13. A local annotation mentioning an enclosing type formal was not checked (#2130, P0)

`let y: T = ..` inside `fn f[T]` resolved `T` to a bare `CtNamed("T", [])` — indistinguishable from an unresolved name, which `check_assignable` must tolerate. The mismatch was computed and then discarded. One call at the ascription site (`resolve_tparams_in(resolve_type_expr(...), env)`) fixes it; 5 tests.

### 14. Declaration count capped a valid module, and the number moved three times (#2134)

A valid file stopped compiling once it had enough top-level declarations — the checker recurses once per statement and node's default ~1 MB stack ran out. The diagnostic blamed "expression too deeply nested", which is the one thing it never was: the bodies are `x + i`.

This issue collected **three contradictory ceilings** before the cause was found, and the cause is worth reading on its own: a successful compile writes a persistent module-header row keyed by source content, after which that exact source compiles on the default stack. Every re-probe of a file that had already gotten through reported a ceiling several times too high. Measured cold with unique content per run, it is monotonic and deterministic — 1200 ok, 1400 fail, 3/3 each way — and linear in the stack (`--stack-size=4000` moves it to ~5000).

`scripts/check_declaration_scale.sh` generates unique content **every run**; a fixed corpus would pass forever after its own first green run, fix or no fix. Its Red case re-checks on fresh content that the same size still fails without the raise.

A follow-up commit fixes a bug in the fix: node takes the **last** `--stack-size`, and `generations.sh` passes `--stack-size=131072` for every bootstrap compile — so the unconditional append silently cut a deliberate 128 MB stack to 4 MB. It broke nothing, which is precisely the problem. Guarded, and asserted end to end.

That also answers the issue's open question: nothing scales past the ceiling; that lane simply had a 128 MB stack all along.

### 15. `collect_unknown_type_names` — the detection half of #2125 (P0)

`fn takes(x: Intt) -> Int` accepts both `takes(1)` and `takes("s")`, and `vibe check` exits 0: an unknown type name becomes a rigid `CtNamed`, which `check_assignable` tolerates, so a typo disables checking at that slot.

Two things the issue listed as "to work out first" now have answers. The formals list alone is **not** sufficient — a lambda nested in a generic function resolves its parameters with its own empty `tparams`, so `fn outer[T](..) { (a: T) -> .. }` reports `T`; the env is the authority instead, by the same `CtVar` test `resolve_tparams_in` uses. And a set of compiler-owned nominal heads has to be named explicitly, because they resolve to `CtNamed` with no `TypeDef` and no builtin entry.

**Enforcement is deliberately not wired**, and that is the interesting result: `gen_arr_equals_fn` synthesizes `(a: Array[T], b: Array[T])` with an *empty* type-parameter list, so a compiler-generated declaration type-checks only **because of** the hole. Measured — wiring it builds a stage1 that then fails to compile the stage2 flat source with exactly one diagnostic, on a two-parameter function whose tparams list has length 0. Filed as #2136 with the evidence; wiring becomes a two-call change once that lands.

11 tests pin what must be reported, what must not, and the synthesized shape that blocks enforcement.

---

Operation gate green on each commit.

---

PR #2294, opened 2026-08-25
Fix declaration-order type ownership (#2290, #2292)

### LINKED ISSUE #2290: Map key: a compiler-owned name shadowed by a LATER alias is still falsely rejected
## Status: half closed by `a3e9f07`

`a3e9f07` ("Revalidate forward map aliases after type collection") added `check_completed_alias_map_keys`, a second walk over alias bodies once the full `TypeDef` table exists. Measured on the stage2 built from `cf8746a`, that **closed the missed-diagnostic half**:

| program | before | now |
|---|---|---|
| `type Bad = Map[Key, Int]` / `type Key = Int`, `Bad` **never used** | accept | **reject** ✅ |
| same, `Bad` used | reject | reject |
| same with `type Key = String`, `Bad` used | accept | accept |

## What remains: the false rejection

The other direction is **not** fixed, and the reason is structural rather than an oversight:

| program | verdict | wanted |
|---|---|---|
| `type M[T] = Map[Future[T], Int]` then `type Future[T] = String` | **reject** | accept |
| the two declarations swapped | accept | accept |
| `type M[T] = Map[Future[T], Int]` then `type Future[T] = Int` | reject | reject |

```
the builtin map is String-keyed, so a `Future[T]` key has no implementation in
the body of type alias `M` -- spell the String-keyed map `StringMap[V]`, or use
`MutMap[K, V]` from `@vibe/core` for other keys (generic `Map` keys are #2263)
```

A second pass can only **add** errors, never retract one. The declaration-time walk still runs against an incomplete `defs`, sees `Future` undeclared, falls to the compiler-owned arm and pushes the diagnostic into `frozen_errors`; `check_completed_alias_map_keys` then re-walks with the complete table, correctly concludes the key is `String`, and has no way to withdraw what the first walk already recorded.

So the remaining fix is not another pass but a **suppression at the declaration-time walk**: an unresolved head that a later declaration might own must not be judged there at all, and the completed pass — which now exists — becomes the sole authority for those heads. Concretely, the compiler-owned arm in `map_key_type_is_supported_at`'s `None` branch should defer when the walk is running at declaration time, since `check_completed_alias_map_keys` will re-ask with the full table.

That deferral is safe now in a way it was not when this issue was first filed: the earlier plan ("defer on the body, catch it at the use") was ruled out by the unused-alias case, and `a3e9f07` is exactly the pass that removes that objection.

## Cost of leaving it

A loud false rejection, fixed by reordering two declarations. Not silent-wrong, so it does not block the 0.1.0 freeze — but it does contradict `map_shadowed_wellknown_key_ok`, which asserts that a source declaration of a compiler-owned name wins.

## History

Filed from Codex review round 17 on #2276, extended by round 20's unused-alias case (which ruled out the first proposed fix), half-closed by `a3e9f07`. Rounds 12–17 of that review were all `map_key_type_is_supported_at`, two of them regressions from earlier fixes in the same function, which is why the remainder was moved out of that PR.

### LINKED ISSUE #2292: Checker: local suberror must shadow same-named imported typedef
## What happens

`local_type_def_names` does not include `SSuberror`. When an imported typedef and a local suberror have the same name, import projection can therefore win even though the local declaration should own the spelling.

This was measured while fixing #2276 (see the reply to discussion `r3852489182`): structs/enums are represented in the local-name ownership scan, but suberrors are omitted even though `preclaim_local_type_defs` registers `SSuberror` as a local `TDEnum`.

## Expected

A local `suberror Name(...)` owns `Name` consistently with local structs and enums. An imported type with the same local spelling must not replace it.

## Scope

Checker/import ownership only. The WIT `StringMap` ownership fix in #2276 already handles `SSuberror`; this is the separate checker-side collision noted during that review.

## Regression test

Pin both declaration/import orders and a control where the imported type has a distinct alias. Verify the resolved local name is the suberror/`TDEnum`, not the imported typedef.

mizchi:
## Summary

- validate type-alias Map keys and builtin arity after the current scope has collected all type declarations, but before value/signature diagnostics
- preserve alias-boundary diagnostics for unused and used aliases without making valid forward aliases declaration-order dependent
- include local suberrors in imported-typedef shadow ownership
- cover both suberror/import orders and a distinct import-alias control

Fixes #2290.
Fixes #2292.

## Tests

- `bash scripts/generations.sh build --out-dir _build/remainder-green2`
- `TYPECHECK_FIXTURES_STAGE2=_build/remainder-green2/stage2.wasm bash scripts/check_typecheck_fixtures.sh` — 286 rows pass
- targeted `type_db_cross_module_test.vibe` through `scripts/unit_test_runner.sh` — pass
- `pkf run test-affected` expanded to the full 1021-file battery: 1007 pass; 14 unrelated fresh-worktree generated-bundle/contract tests fail because generated implementations such as `cli_adapter_sources` are absent, plus the pre-existing loader cache assertion. The changed target files and fixture gate pass independently as listed above.


lib/@vibe/compiler/checker/checker_stmt.vibe
```diff
@@ -3792,6 +3807,101 @@ fn check_completed_alias_map_keys(stmts: Array[Stmt], defs: Array[TypeDef], erro
...
+      STypeAlias(exported, name, params, _) => {
+        let idx = completed_alias_index(defs, name, from, taken)
+        if idx >= 0 {
+          match Array::get(defs, idx) {
+            TDAlias(_, _, body) => {
+              current = EnvTypeDefs([TDAlias(name, params, body)], if exported {
+                [name]
+              } else {
+                array_empty()
+              }, env_bind(current, name, body))
```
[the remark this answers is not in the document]
mizchi: Fixed in 96763fa0b. Instead of prepending refreshed aliases, the checker now replaces each local alias payload in its existing EnvTypeDefs node and updates the adjacent alias binding while preserving the original environment-chain topology. Reverse declaration matching preserves duplicate-name ordering, and the walk stops before the imported/outer tail. The regression also asserts that the final Pair lookup is resolved, and the final stage2/typecheck/cross-module gates pass.

lib/@vibe/compiler/checker/checker_stmt.vibe
```diff
@@ -2423,8 +2424,10 @@ fn check_stmts_with_binders(raw_stmts: Array[Stmt], i: Int, env: TypeEnv, defs:
         }, bind_fn_return(env5, name, CtStruct(name))), s, c)
       },
       STypeAlias(exported, name, params, te) => {
-        // #1512: `params` shadow `defs` -- see resolve_type_expr_shadowed.
-        report_map_key_types(te, defs, params, String::concat(String::concat("the body of type alias `", name), "`"), errors)
+        // #2290: map-key and builtin-arity validation is deferred to
+        // check_completed_alias_map_keys. At this point a later alias is not
+        // in `defs`, so judging ownership here made valid programs depend on
+        // declaration order. Resolution still records the body as before.
         let t = resolve_type_expr_shadowed(te, defs, params)
```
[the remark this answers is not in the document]
mizchi: Fixed in 1b067a72c and refined in 96763fa0b. Alias bodies are now re-resolved to a bounded fixpoint after the local declaration table is complete, then the stored TypeDef and environment payloads are refreshed. Added a rejecting local fixture for `let bad: Pair = 1`, a positive tuple assignment, and a cross-module projection regression. The final stage2 build, all 292 typecheck rows, and the targeted cross-module unit test pass.

mizchi:
@codex review

lib/@vibe/compiler/checker/checker_stmt.vibe
mizchi: **[bug]** `publish_completed_alias_defs` walks every local alias after the type-declaration phase and does `env_bind(current, name, body)` on top of the completed environment. `env_lookup` is first-match and skips `EnvTypeDefs`, so those binds hide any later constructor / struct-ctor / suberror bind that reused the alias spelling. `type Make = Int; enum E { Make(Int) }; let x = Make(1)` used to resolve `Make` as the constructor (the enum ran after the alias). After this pass it resolves as `CtInt` and is not callable. Same collision for `type Point = (Int, Int); struct Point { ... }` and `type Label = Int; suberror Label(String)`. Same-file annotation lookup through `defs` is fine; this is specifically the value-environment restack.

**Suggestion:** Do not stack alias value binds on the completed env. Either republish only `EnvTypeDefs` (projection reads that; same-file annotations already re-resolve from `defs`), or rebuild the type-declaration environment in original walk order so a later constructor still wins. Pin with `type Make = Int; enum E { Make(Int) }; let x = Make(1)` plus the reverse order.

lib/@vibe/compiler/checker/checker_stmt.vibe
```diff
@@ -3792,6 +3807,101 @@ fn check_completed_alias_map_keys(stmts: Array[Stmt], defs: Array[TypeDef], erro
...
+    let mut i = 0
+    while i < Array::length(stmts) {
+      match Array::get(stmts, i) {
+        STypeAlias(_, name, params, te) => {
+          let idx = completed_alias_index(defs, name, from, taken)
+          if idx >= 0 {
+            let resolved = resolve_type_expr_shadowed(te, defs, params)
+            match Array::get(defs, idx) {
+              TDAlias(_, _, old_body) => if !types_equal(old_body, resolved) {
+                Array::set(defs, idx, TDAlias(name, params, resolved))
```
mizchi: **[suggestion]** `refresh_completed_alias_defs` only `Array::set`s `TDAlias` slots. `SStruct` / `SEnum` / `SSuberror` / `SEffectDef` still resolve field and payload type expressions during the type walk and store the expanded `Type`. For `type Pair = StringMap[Int, String]; struct Box { value: Pair }; type StringMap[A, B] = (A, B)`, `Pair` is refreshed to `(Int, String)` and `let bad: Pair = 1` now rejects, but `Box.value` keeps `CtNamed("StringMap", ...)`. `type_is_ground` treats `CtNamed` as not ground, so `check_assignable` still tolerates `Box::{ value: 1 }`. That is the same silent-wrong the P1 follow-up closed for a direct alias annotation.

**Suggestion:** After the alias fixpoint, re-resolve stored field/payload types whose expansion mentioned a refreshed alias, or store a nominal alias reference and expand at use. At least pin `struct Box { value: Pair }` next to `map_forward_alias_body_enforced_reject.vibe` so the hole cannot widen quietly.

mizchi:
@codex review\n\nPlease re-review 8c85ff678.\n\n- Review comment 3853214941: replaced the statement-count fixpoint with a cycle-aware reverse single pass; added a bounded-depth recursive-alias regression.\n- Review comment 3853214950: refreshes and revalidates the outer local-def range after nested modules; added a raw-AST regression.\n- Updated the stale forward-applied-alias expectation exposed by CI.\n\nLocal validation: rebuilt stage2, typecheck fixtures, targeted regressions, and exact unit shard 2 (257/257).

lib/@vibe/compiler/checker/checker_stmt.vibe
```diff
@@ -3792,6 +3808,168 @@ fn check_completed_alias_map_keys(stmts: Array[Stmt], defs: Array[TypeDef], erro
...
+fn refresh_completed_alias_defs(stmts: Array[Stmt], defs: Array[TypeDef], from: Int) -> Unit {
+  let mut round = 0
+  let mut changed = true
+  while changed && round <= Array::length(stmts) {
+    changed = false
+    let taken = array_empty()
+    let mut i = 0
+    while i < Array::length(stmts) {
+      match Array::get(stmts, i) {
+        STypeAlias(_, name, params, te) => {
```
mizchi: **[bug]** `refresh_completed_alias_defs` only re-resolves `STypeAlias` slots, and `refresh_completed_alias_env` copies every non-alias `TypeDef` through unchanged (`other => Array::push(refreshed, other)`). `resolve_type_expr_core_shadowed` inlines a `TDAlias` body (`TyName` returns `t`, `TyApp` substitutes into `body`), so a later type declaration that mentions the alias stores a snapshot, not a named hop.

Concrete program:

```
type Pair = StringMap[Int, String]
struct Box { value: Pair }
type StringMap[A, B] = (A, B)
let bad = Box::{ value: 1 }
```

At `Box`, `StringMap` is not yet in `defs`, the 2-arg application misses the 1-arg builtin special case, and `Pair`'s body is `CtNamed("StringMap", [CtInt, CtString])`. The field (and `Box`'s `CtFn` env bind) inlines that. Refresh then rewrites `Pair` to `CtTuple([CtInt, CtString])` — which is why `let bad: Pair = 1` now reports `binding type mismatch for bad` — but `Box.value` stays the `CtNamed`. `check_assignable` swallows `CtNamed("StringMap", [Int, String])` vs `CtInt`: `type_is_ground` is false, `head_kind` is 0, `nominal_head_conflict` needs names on both sides, and `detect_narrowed_generic_mismatch` only reports a same-name `CtNamed` pair. The value lane therefore compiles. The same snapshot is in enum / suberror constructor arguments and effect operation types (`Make(Pair)`, `suberror E(Pair)`). Import projection publishes those `EnvTypeDefs` payloads, so an imported `Box` carries the hole too.

**Suggestion:** At the same type-decl phase boundary, re-resolve the TypeExpr of every local struct field, enum / suberror payload, and effect operation against the completed `defs`, then write the new types back into `TDStruct` / `TDEnum` / `TDEffect`, the matching `EnvTypeDefs` node, and the constructor / struct-ctor `EnvBind` (`CtFn` argument list). Pin `Box::{ value: 1 }` and `Make(1)` as rejects next to `map_forward_alias_body_enforced_reject.vibe`, and a `Box::{ value: (1, "ok") }` accept next to `map_forward_string_map_alias_arity_ok.vibe`.

mizchi:
@codex review\n\nPlease re-review 26d8151bd.\n\n- Review comment 3853500204: 58882632c resolves acyclic zig-zag alias dependencies with an alias-count-bounded pass over raw source expressions while skipping recursive aliases.\n- 26d8151bd also closes the remaining alias-snapshot risk: after aliases complete, struct/enum/suberror/effect payload definitions and their existing environment/companion bindings are refreshed without restacking. Alias bindings are replaced with their completed bodies directly, so recursive aliases are not repeatedly expanded.\n- Added regressions for zig-zag order, bounded recursive depth, struct construction and field projection, enum payloads, effect operations, nested modules, and cross-module exported struct projection.\n\nLocal validation: rebuilt stage2, typecheck fixtures (292 rows), targeted regression files, and exact unit shard 2 (257/257).

mizchi:
@codex review\n\nPlease re-review a8ea564b3.\n\n- Review comment 3853739459: after each processed module, aliases visible across the outer prefix and nested module bodies are flattened into one declaration graph and refreshed over the combined completed definition range. Added a raw-AST regression asserting the alternating outer/nested chain resolves A to Int.\n- Review comment 3853739468: reachability now shares one visited set per target and recursive alias names are computed once outside the refresh rounds. Added a 16-level diamond-suffix regression that completes without increasing the unit file runtime materially.\n- The payload snapshot fixes from 26d8151bd remain covered for struct construction/field projection, enum payloads, effect operations, and cross-module projection.\n\nLocal validation: rebuilt stage2, typecheck fixtures (292 rows), targeted regressions, and exact unit shard 2 (257/257).

mizchi:
@codex review\n\nPlease re-review 7cd957d66.\n\nReview comment 3853889180 is fixed by rebuilding each local suberror constructor binding and its hidden function-return binding from the completed TDEnum payloads. This is done in-place in the existing environment and has a regression for a suberror payload that snapshots a forward-applied alias.\n\nLocal validation: rebuilt stage2, typecheck fixtures (292 rows), targeted regressions, and exact unit shard 2 (257/257).

lib/@vibe/compiler/checker/checker_stmt.vibe
```diff
@@ -3776,20 +3806,422 @@ fn collect_type_param_shadowing_errors(stmts: Array[Stmt], errors: Array[String]
...
+
+/// Refreshes the completed local declaration payloads and their companion
+/// value signatures without restacking the environment. The exact local type
+/// count provides the boundary that keeps imported and outer nodes untouched.
+fn refresh_completed_type_env(env: TypeEnv, defs: Array[TypeDef], from: Int, to: Int, taken: Array[Int], local_count: Int) -> TypeEnv {
+  if Array::length(taken) >= local_count {
+    env
+  } else match env {
+    EnvEmpty => EnvEmpty,
+    EnvBind(name, binding, rest) => EnvBind(name, binding, refresh_completed_type_env(rest, defs, from, to, taken, local_count)),
```
mizchi: **[bug]** `refresh_completed_payload_defs` re-resolves suberror payloads in `defs`, but construction does not read those payloads. `suberror Boom(Pair)` binds `Boom` as `EnvBind("Boom", CtFn([resolved_payload], CtEnum("Boom")), _)` and never wraps that chain in `EnvTypeDefs`. `refresh_completed_type_env` then leaves `EnvBind` types untouched (`EnvBind(name, binding, refresh_completed_type_env(rest, ...))`). Only a non-alias `EnvTypeDefs` node sets `refresh_prefix` and walks companion `EnvBind`s via `refresh_type_env_prefix`; an alias node only patches the immediate alias binding. So this program still type-checks:

```
type Pair = StringMap[Int, String]
suberror Boom(Pair)
type StringMap[A, B] = (A, B)
let bad = Boom(1)
```

At constructor bind time `Pair` expands to unresolved `CtNamed("StringMap", [CtInt, CtString])`. `Boom(1)` goes through `unify_call_args` → `check_assignable`, `type_is_ground` is false for any `CtNamed`, and the Int argument is accepted. `match` against the refreshed `TDEnum` would then see a tuple payload — construction and elimination disagree. The new tests cover `struct` / `enum` / `effect` only; enum constructors happen to sit under `EnvTypeDefs([TDEnum...])` and get the prefix rewrite, which is why `Pick(1)` is rejected and `Boom(1)` is not. A later struct/enum/effect does not reliably save this either: `refresh_type_env_prefix` stops at the next `EnvTypeDefs`, so constructors sitting between two alias nodes stay stale. The same `local_count` includes the preclaimed suberror `TDEnum` that no `EnvTypeDefs` will ever claim, so the "do not touch imported nodes" boundary does not hold whenever a local suberror exists.

**Suggestion:** Publish `SSuberror` as `EnvTypeDefs([TDEnum(name, [], resolved_variants)], selectable, env_eq)` the way `SEnum` already does, so constructor `EnvBind`s become that node's prefix and `local_count` matches claimed nodes. Then add the missing sibling of the enum test: `type Pair = StringMap[Int, String]\nsuberror Boom(Pair)\ntype StringMap[A, B] = (A, B)\nlet bad = Boom(1)` must be `check_source_err`.

mizchi:
@codex review\n\nPlease re-review adea2f869.\n\nReview comment 3853990023 is fixed by collecting all visible type declarations, not only aliases, from processed outer/nested scopes. Alias refresh, payload rebuild, and Map-key validation now run over the same combined declaration graph and completed definition range. Added raw-AST regressions for a later sibling alias invalidating an earlier nested Map key and completing an earlier nested struct payload.\n\nLocal validation: rebuilt stage2, typecheck fixtures (292 rows), targeted regressions, and exact unit shard 2 (257/257).

mizchi:
@codex review\n\nPlease re-review c57c18127.\n\nReview comment 3854033444 is fixed by locating the suberror's unique <Name>::equals marker first and refreshing constructor/hidden-return bindings only in the marker's owned suffix. Later same-named struct constructors above that marker remain untouched. Added the exact suberror-variant/later-struct collision regression.\n\nLocal validation: rebuilt stage2, typecheck fixtures (292 rows), targeted regressions, and exact unit shard 2 (257/257).

mizchi:
Addressed the two P1 findings in 61172f561:

- Recheck module value statements against completed shared type definitions, using each module’s original source-order environment and copied definition tables.
- Revalidate Map keys in completed struct, enum, suberror, and effect payloads in addition to aliases.
- Added regressions for an earlier `Box("x")` use and a completed suberror `Map[Key, Int]` payload.

Local validation: fresh selfhost stage2 generation; 3/3 focused unit files; 292 typecheck fixture rows; unit shard 2/4 at 257/257.

@codex review

mizchi:
Addressed the additional P1 in fe1ff2827.

Completed declaration validation now runs once at the root after the full nested type graph is known. Struct/enum payloads still report unknown names immediately, but Map-key and structural-head arity checks are deferred alongside aliases, suberrors, and effects. Completed declaration diagnostics are inserted before later use-site diagnostics to preserve the existing diagnostic contract.

Added regressions for a nested alias shadowing the builtin `StringMap` head across aliases and struct/enum/suberror/effect payloads, plus diagnostic precedence.

Local validation: fresh selfhost stage2; focused tests 3/3; typecheck fixtures 292/292; unit shard 2/4 at 257/257.

@codex review

mizchi:
Addressed the outer-statement P1 in ab54d39d9.

When a nested module extends the shared type-definition graph, the root statement list is now rechecked once against copied completed definitions. This covers both outer and nested values from their first lookup and rebuilds ordinary final value bindings so completed alias types propagate to later statements. Observation modes retain the first-pass env/substitution pairing while still consuming second-pass diagnostics, preserving typed-artifact IDs.

Added regressions for an invalid outer `Box("x")` call and for completed alias types propagating through an earlier binding into a later mismatch.

Local validation: fresh selfhost stage2; focused tests 4/4; typecheck fixtures 292/292; unit shard 2/4 at 257/257.

@codex review

mizchi:
Addressed both new P1 findings in b1b3db7e5.

- When nested types require the completed root recheck, first-pass statement diagnostics are now replaced (not appended to) by completed declaration diagnostics plus second-pass statement diagnostics. This removes stale builtin-head errors from valid programs.
- Suberror binding refresh now locates its declaration-owned `Name::equals` marker by source-order ordinal across same-name enum/struct/suberror declarations, rather than taking the first identical marker.

Added regressions for a valid earlier module annotation whose head is claimed later, and for a same-name suberror/struct under observation-mode environment retention.

Local validation: fresh selfhost stage2; focused tests 4/4; typecheck fixtures 292/292; unit shard 2/4 at 257/257.

@codex review

mizchi:
The latest unknown-name finding is already covered by b1b3db7e5: when `nested_defs_extended` is true, the checker truncates all first-pass statement diagnostics at `completed_decl_error_insert_at` and replaces them with completed declaration plus second-pass diagnostics.

I added the exact raw-AST regression in 8bfce5f2f: an outer struct and enum directly reference `Pair`, a later nested module defines `type Pair = Int`, checking succeeds, and the final `Box.value` payload is asserted as `CtInt`. The focused suite passes 3/3 against the fresh b1b3db7e5 stage2.

@codex review

mizchi:
Addressed both P1 findings in 208e51798.

- Completed root rechecks now restore a full observation checkpoint (legacy type table, statement capture entries, role/provenance/direct entries, path frames, depth, and completeness), then rerun with the original capture configuration. Final env/substitution and retained observations now come from the same completed pass.
- Recursive-alias suppression is tracked per statement/declaration rather than by spelling, so a recursive alias in one visible scope no longer freezes an acyclic same-name alias in another.

Added exact regressions for completed root-type observations and for an outer acyclic `A = B` alongside nested recursive `A = A`.

Local validation: fresh selfhost stage2; focused tests 4/4; typecheck fixtures 292/292; unit shard 2/4 at 257/257.

@codex review

mizchi:
@codex review

Addressed the P1 exponential alias-DAG expansion in 0ed43d809.

- Preflight resolved tree size before both initial alias registration and completion refresh.
- Saturate at 262,144 nodes without materializing the candidate tree.
- Propagate capped dependencies through the refresh and emit an actionable diagnostic recommending a nominal struct or enum.
- Preserve the existing 16-level diamond success case.
- Add 30-level regressions for both forward-reference and dependency-first declaration orders; the former timed out before the fix and now terminates with the diagnostic.

Validation from a fresh final stage2:
- focused unit files: 4/4
- typecheck fixtures: 292/292 (24 known debt)
- unit shard 2/4: 257/257
- formatter checks and git diff --check
- staged review-regressions AST lint: clean

mizchi:
@codex review

mizchi:
Addressed the eager-unused-argument P1 in 486d2751d.

Alias applications now cap the combined peak of eagerly materialized argument trees and the substituted result. Once expansion is rejected, completed Map-key diagnostics no longer replay the same eager expansion while collecting secondary errors. Added the exact `Ignore[Dup[...]]` depth-30 regression; it failed before the fix and now returns the actionable expansion diagnostic.

Fresh final-stage2 validation: focused 4/4, typecheck fixtures 292/292, unit shard 2/4 257/257, formatter/diff checks, and staged review-regressions AST lint.

mizchi:
@codex review

mizchi:
Addressed the completed-recheck path P1 in 4ef82f90e.

The completed semantic pass now re-enters through the original `raw_stmts` and original start index instead of treating the already partitioned statement list as source order. Added a `[value, type, module]` regression that requires the rebuilt `Int` function type while preserving the original statement path `[0]`.

Fresh final-stage2 validation: focused 4/4, typecheck fixtures 292/292, unit shard 2/4 257/257, formatter/diff checks, and staged review-regressions AST lint.

mizchi:
@codex review

PR #2299, opened 2026-08-25
Expand aliases before structural `==` (#2281)

### LINKED ISSUE #2281: A parameterized alias wrapping a structural container loses structural `==` (silently)
Measured 2026-08-24. **Pre-existing** — reachable without any of `StringMap`, and not confined to maps.

## The defect

Wrap a structural container in a parameterized type alias, and `==` silently falls back to identity comparison. Two separately allocated, element-wise equal values answer `false`.

```vibe
type AL[V] = Array[V]

fn mk() -> AL[Int] {
  [1, 2, 3]
}

fn cmp(x: AL[Int], y: AL[Int]) -> Bool {
  x == y
}

fn main() -> Unit with Stdout {
  println(if cmp(mk(), mk()) { "true" } else { "false" })   // prints: false
}
```

Measured, same shape each time, one build:

| annotation | `==` on two equal values |
|---|---|
| `Array[Int]` directly | `true` |
| `type AL[V] = Array[V]`, used as `AL[Int]` | **`false`** |
| `StringMap[Int]` directly | `true` |
| `type SM[V] = StringMap[V]`, used as `SM[Int]` | **`false`** |
| `type SM[V] = Map[String, V]`, used as `SM[Int]` | **`false`** |

The `Map[String, V]` row is the one that dates it: that spelling predates `StringMap` entirely, so this is not new surface. The `Array` row is the one that sizes it — this has nothing to do with maps.

## Why

`ty_to_eq_shape` and `ty_needs_typed_eq` (`normalize/desugar_trait_dict.vibe`) classify the **syntactic** `TypeExpr`. An alias application is `TyApp("AL", [Int])`, which matches no structural arm, so `ty_needs_typed_eq` returns false and the lowering keeps `EBinOp("==", ...)` — raw identity. The alias is never expanded before classification.

## Why it is P0 rather than P1

This is the failure mode `CLAUDE.md`'s design policy ranks above crashing: the program runs, produces an answer, and the answer is wrong. Nothing is reported, and the value's own type spelling is what decides — so the same data compares equal or unequal depending on whether the author wrote the alias.

It also makes a documented guarantee false. `CLAUDE.md`'s equality section states, of ADR-0097 as shipped, "**measured 2026-08-19, no silent reference equality is left**" and enumerates the contexts that are structural. Alias-wrapped containers are a context that is not, and the note should say so once this is either fixed or scoped. I have not edited that section — the correction belongs with whoever decides the fix.

## Fix

Expand type aliases before shape classification, so `AL[Int]` reaches the same arm `Array[Int]` does. Both `ty_to_eq_shape` and `ty_needs_typed_eq` need it, since they are consulted separately.

Watch the interaction with **#2279**: the obvious way to expand a *parameterized* alias is `resolve_type_alias_params`, and consuming that function's result corrupts the heap. Either fix #2279 first, or expand without materializing a substituted tree (what `collect_unsupported_map_keys_applied` in `checker.vibe` does for the map-key check).

## Red test

`fixtures/structural_eq_contexts_test.vibe` is where the shipped contract is pinned; the alias rows belong there. Three cases, since they fail for the same reason but would be fixed by different arms: `type AL[V] = Array[V]`, `type SM[V] = StringMap[V]`, and a non-parameterized `type A2 = Array[Int]` (not yet measured — worth checking whether the unparameterized form already works, since it needs no substitution).

Reported by Codex on #2276 as the `StringMap`-through-alias case; the `Array` and `Map[String, V]` rows above are from following it out.

Axes: **bug** / **P0** (silently wrong answer) / not a blocker for the 0.1.0 tag — it predates the tag work, and #2276 neither causes nor worsens it.

mizchi:
## Summary

- expand alias bodies before structural equality classification
- preserve in-scope type formals before alias expansion
- seed equality shapes for nullary structural aliases used as parameter types
- cover parameterized and nullary Array/StringMap aliases, derived fields, enum payloads, and same-spelled formals with inspect snapshots

Fixes #2281.

## Why P0

The program runs and answers. An alias-wrapped structural container compared by identity while the direct container spelling compared by content.

## Validation

- Red: nullary alias parameters compared equal arrays as false
- Green: structural_eq_contexts_test.vibe 43/43 with a fresh stage2
- changed-file formatter checks
- review-regressions AST lint
- selfcompile heap: 1281905728 before the final fix -> 1282251440 (+345712, +0.027%); baseline updated because the branch was already 3137 bytes below the old +10% ceiling

## Collision

#2296 also edits desugar_trait_dict.vibe, but only the assert_eq lowering. This change is in equality-shape classification and parameter seeding; git merge-tree reports no conflict.

### DIFF
--- a/lib/@vibe/compiler/normalize/desugar_trait_dict.vibe
+++ b/lib/@vibe/compiler/normalize/desugar_trait_dict.vibe
@@ -4251,7 +4251,22 @@ fn seed_var_types(params: Array[(String, Option[TypeExpr])], tparams: Array[Stri
   while i < Array::length(params) {
     let (pname, ann) = Array::get(params, i)
     match ann {
-      Some(TyName(tn)) => Array::push(out, (pname, tn)),
+      Some(TyName(tn)) => {
+        Array::push(out, (pname, tn))
+        // A nullary alias can hide a structural parameter type just as a
+        // TyApp annotation can. Preserve a same-spelled in-scope formal before
+        // expanding through the global alias table.
+        if !ty_mentions_name_in(TyName(tn), tparams) {
+          let esh = ty_to_eq_shape(TyName(tn))
+          if shape_is_composite(esh) {
+            Array::push(out, (eq_shape_key(pname), esh))
+          } else {
+            ()
+          }
+        } else {
+          ()
+        }
+      },
       // A generic struct annotation `x: LazyIter[T]` -> track the head name so
       // `for x in <param>` over a generic iterator type resolves (#636).
       // #1199 bug 1: for `x: Array[T]` specifically, ALSO record T under the
--- a/lib/@vibe/compiler/normalize/desugar_trait_dict.vibe
+++ b/lib/@vibe/compiler/normalize/desugar_trait_dict.vibe
@@ -8390,7 +8390,7 @@ fn eq_unsafe_types_fill(stmts: Array[Stmt]) -> Unit {
 /// a `Map` arm. It is admitted here for the same reason `Array` is: there is
 /// now a generated comparator that descends into it.
 fn eq_ty_field_is_content_comparable(t: TypeExpr, tparams: Array[String]) -> Bool {
-  match t {
+  match eq_expand_alias_ty(t) {
     TyName(n) => if str_array_contains(tparams, n) {
       // The owner's own type parameter is resolved by the concrete specialized
       // comparator, so it is not a declaration-level failure.
@@ -8847,7 +8847,7 @@ fn eq_for_typed(fty0: TypeExpr, lhs: Expr, rhs: Expr, known_types: Array[String]
   // seed_var_types / collect_fn_eq_return_shape), so today this subsumes the
   // Array-arm re-spelling; at generation time the scope stack is empty and
   // this is the identity.
-  let fty = subst_scope_formals_ty(fty0)
+  let fty = eq_expand_alias_ty(subst_scope_formals_ty(fty0))
   match fty {
     TyName(n) => if n == "__EqAliasCycle" {
       ESeq(ECall(EIdent("assert_true", -1), Array::slice([EBool(false)], 0, 1), -1), EBool(false))
@@ -9712,9 +9712,7 @@ fn ty_mentions_name_in(t: TypeExpr, names: Array[String]) -> Bool {
 /// (`shape_key`'s: Name / `(s1,..)` / `[s]` / "" unknown). Only shapes this
 /// pass can compare are produced; anything else collapses to "" (unknown).
 fn ty_to_eq_shape(t: TypeExpr) -> String {
-  // #2281: classify the alias body, not the alias head. `type AL[V] = Array[V]`
-  // used as `AL[Int]` is `TyApp("AL", ...)` and matched no structural arm, so
-  // `==` stayed raw identity. Expand first; a non-alias is a no-op.
+  // #2281: classify the alias body so AL[Int] is Array, not TyApp("AL", ...)
   match eq_expand_alias_ty(t) {
     TyName(n) => n,
     TyUnit => "{u}",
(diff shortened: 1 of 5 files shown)


lib/@vibe/compiler/normalize/desugar_trait_dict.vibe
```diff
@@ -9712,7 +9712,10 @@ fn ty_mentions_name_in(t: TypeExpr, names: Array[String]) -> Bool {
 /// (`shape_key`'s: Name / `(s1,..)` / `[s]` / "" unknown). Only shapes this
 /// pass can compare are produced; anything else collapses to "" (unknown).
 fn ty_to_eq_shape(t: TypeExpr) -> String {
-  match t {
+  // #2281: classify the alias body, not the alias head. `type AL[V] = Array[V]`
+  // used as `AL[Int]` is `TyApp("AL", ...)` and matched no structural arm, so
+  // `==` stayed raw identity. Expand first; a non-alias is a no-op.
+  match eq_expand_alias_ty(t) {
```
mizchi: **[bug]** Expansion was added only in `ty_to_eq_shape` / `ty_needs_typed_eq`. That is enough for a function result, a parameter annotation, or a `let` ascription, because those record a shape and `shape_to_eq_ty` rebuilds `Array`/`Map` for `eq_for_typed`. It is not enough for a declared field or enum payload of alias type. `gen_eq_fn` / `gen_enum_eq_fn` call `eq_for_typed` on the field type as written (`TyApp("AL", [Int])` / `TyName("A2")`). That match has no alias arm; `AL` is not a generic struct and is not in `dtd_eq_known_types` (only `SStruct`/`SEnum`), so the field compare is a raw `==`. `s1 == s2` dispatches to `Holder::equals` and never revisits `ty_to_eq_shape`. Same process already expands field types in `gen_specialized_eq_fn` (`Box[AL[Int]]` is structural; `struct Holder { xs: AL[Int] }` is not). `eq_ty_field_is_content_comparable` also does not expand, so `Holder` is marked unsafe (empty-array adoption traps) while constructed values still identity-compare. AGENTS.md now lists parameterized aliases among the contexts that are "all structural"; that is false for this path. The new tests only compare function returns, so they cannot catch it.

**Suggestion:** Expand once at the start of `eq_for_typed` (`let fty = eq_expand_alias_ty(subst_scope_formals_ty(fty0))`), which is the same table already populated when `desugar_derives` runs. Expand in `eq_ty_field_is_content_comparable` too so an alias of a content-comparable target does not mark its owner unsafe. Pin it with a struct field and an enum payload of `AL[Int]` / `A2` in `fixtures/structural_eq_contexts_test.vibe` (values built separately so identity and content disagree). If dispatch is left unexpanded, narrow the AGENTS.md sentence to the classification/annotation/return-type path this PR actually fixes.

lib/@vibe/compiler/normalize/desugar_trait_dict.vibe
```diff
@@ -8390,7 +8390,7 @@ fn eq_unsafe_types_fill(stmts: Array[Stmt]) -> Unit {
 /// a `Map` arm. It is admitted here for the same reason `Array` is: there is
 /// now a generated comparator that descends into it.
 fn eq_ty_field_is_content_comparable(t: TypeExpr, tparams: Array[String]) -> Bool {
-  match t {
+  match eq_expand_alias_ty(t) {
```
mizchi: **[bug]** `eq_ty_field_is_content_comparable` now expands before the `TyName` arm that treats `tparams` as content-comparable. A field written `T` on `struct Box[T]` is the parameter, not the alias. If a `type T = ...` exists in the same program, expansion rewrites that field to the alias body *before* `str_array_contains(tparams, n)` can fire. When the body is not comparable, `Box` is marked unsafe and an unannotated `let xs = []` that pushes `Box[Int]` traps.

**Suggestion:** Keep the tparams short-circuit on the *unexpanded* `TyName`, then expand. Cover with an incomparable alias next to `struct Cell[F] { value: F } derive (Eq)` and an unannotated empty binding that pushes `Cell[Int]`.

mizchi:
@codex review

PR #2308, opened 2026-08-25
Rewrite re-export aliases in the merged program (#2287)

### LINKED ISSUE #2287: A re-export alias never reaches the merged program: `export ./dep.vibe { cmp as compare }` dies at codegen
Found while checking a #2277 review finding about `assert_eq` shadowing. The finding pointed at a real symptom, but the cause is not shadowing — it is that a re-export **alias** binds in the checker and then vanishes in the merge.

## Reproduction

```vibe
// dep.vibe
export fn cmp(a: Int, b: Int) -> Int {
  a + b
}
```

```vibe
// main_test.vibe
export ./dep.vibe { cmp as compare }

test "re-exported alias" {
  assert_true(compare(1, 2) == 3)
}
```

```
FAIL (compile) main_test.vibe
       internal compiler error: `compare` (local, @call) reached code generation
       unresolved. The type checker should have bound or rejected this name, so
       this is a bug in the compiler and not in your program -- please report
       the source that triggers it.
```

The message is right about where the fault is. `VIBE_EMIT_MERGED_SOURCE=1` shows why:

```vibe
export let rec cmp = (a: Int, b: Int) -> Int { (a + b) }
test "re-exported alias" { assert_true((compare(1, 2) == 3)) }
```

The merge inlines `cmp` under its original name and drops the re-export, but the call site still says `compare`. Nothing binds it, so the name reaches codegen unresolved. `checker_stmt.vibe` routes `SReExport` through the same `bind_stmt_imports` as `SImport`, which is why the checker accepts the program that codegen then cannot compile.

The plain (non-aliasing) form works, because the dependency's own definition arrives under the name the caller uses:

```vibe
export ./dep.vibe { cmp }   // calling `cmp` is fine
```

## Why it went unnoticed

The one alias spelling that does *not* crash is `cmp as assert_eq` — the assert lowering claims the call by name and runs the builtin assertion, so the program "works" and silently computes an assertion instead of calling `cmp`. That is how this was found: a review finding about `assert_eq` shadowing, whose real content is this bug wearing a disguise. Every other alias name gets the internal error.

## Where it likely lives

Two candidates, and the fix belongs in whichever owns alias resolution for re-exports:

- `core/import_alias_rewrite.vibe` already has `SReExport(raw_path, items)` handling; whatever rewrites `alias -> original` at call sites is either not running for re-exports or is running on the wrong side of the merge.
- The merge itself, which emits the dependency's definitions under their original names — an alias needs either a call-site rewrite or a generated `let <alias> = <original>` binding.

## Regression to add

`tests/gates/late/run.sh`: the two-file shape above with a neutral alias, asserting the test passes; the `assert_eq` spelling belongs with it, asserting the builtin assertion does **not** claim the call.

mizchi:
## Summary

- `collect_import_aliases` now includes `SReExport` items, so `compare` is rewritten to `cmp` like an import alias.
- `stmts_have_import_deep` treats `SReExport` as an import. A file that only re-exports otherwise skipped the rewrite path entirely.

Fixes #2287.

## Reproduction

```vibe
export ./dep.vibe { cmp as compare }
test { assert_true(compare(1, 2) == 3) }
```

Before: internal compiler error, `compare` unresolved at codegen.
After: the test passes. A re-export aliased to `assert_eq` runs the exported function, not the builtin.

## Tests

Red then green through this checkout's stage2:

```
VIBE_TEST_CLI_WASM=_build/selfhost/generations/2293b/stage2.wasm \
  bash scripts/vibe_test.sh fixtures/reexport_alias_test.vibe
# ICE: compare unresolved

VIBE_TEST_CLI_WASM=_build/selfhost/generations/2287b/stage2.wasm \
  bash scripts/vibe_test.sh fixtures/reexport_alias_test.vibe
  bash scripts/vibe_test.sh fixtures/reexport_alias_assert_eq_test.vibe
# ok
```

### DIFF
--- a/lib/@vibe/compiler/core/import_alias_rewrite.vibe
+++ b/lib/@vibe/compiler/core/import_alias_rewrite.vibe
@@ -696,26 +696,29 @@ fn rewrite_alias_expr_ix(expr: Expr, am: AliasIdx) -> Expr {
   }
 }
 
+fn collect_item_aliases(items: Array[ImportItem], aliases: Array[(String, String)]) -> Unit {
+  let mut j = 0
+  while j < Array::length(items) {
+    let item = Array::get(items, j)
+    let original = item.name
+    let alias_opt = item.alias
+    match alias_opt {
+      Some(alias) => if alias != original {
+        Array::push(aliases, (alias, original))
+      },
+      None => ()
+    }
+    j = j + 1
+  }
+}
+
 export fn collect_import_aliases(stmts: Array[Stmt]) -> Array[(String, String)] {
   let aliases = empty_aliases()
   let mut i = 0
   while i < Array::length(stmts) {
     match Array::get(stmts, i) {
-      SImport(_, items) => {
-        let mut j = 0
-        while j < Array::length(items) {
-          let item = Array::get(items, j)
-          let original = item.name
-          let alias_opt = item.alias
-          match alias_opt {
-            Some(alias) => if alias != original {
-              Array::push(aliases, (alias, original))
-            },
-            None => ()
-          }
-          j = j + 1
-        }
-      },
+      SImport(_, items) => collect_item_aliases(items, aliases),
+      SReExport(_, items) => collect_item_aliases(items, aliases),
       _ => ()
     }
     i = i + 1
@@ -847,6 +850,7 @@ export fn stmts_have_import_deep(stmts: Array[Stmt]) -> Bool {
   let rec stmt_has_import_deep: (Stmt) -> Bool = (stmt) -> {
     match stmt {
       SImport(_, _) => true,
+      SReExport(_, _) => true,
       SModule(_, _, inner_stmts) => stmts_have_import_deep(inner_stmts),
       _ => false
     }
--- a/lib/@vibe/compiler/core/import_alias_rewrite.vibe
+++ b/lib/@vibe/compiler/core/import_alias_rewrite.vibe
@@ -175,6 +175,25 @@ fn alias_map_without_names(aliases: Array[(String, String)], names: Array[String
   }
 }
 
+fn alias_map_put_last(aliases: Array[(String, String)], local: String, target: String) -> Unit {
+  let mut i = 0
+  let mut found = false
+  while i < Array::length(aliases) && !found {
+    let (alias, _) = Array::get(aliases, i)
+    if alias == local {
+      Array::set(aliases, i, (local, target))
+      found = true
+    } else {
+      i = i + 1
+    }
+  }
+  if !found {
+    Array::push(aliases, (local, target))
+  } else {
+    ()
+  }
+}
+
 fn pat_binds_name(pat: Pat, name: String) -> Bool {
   match pat {
     PBind(bound) => bound == name,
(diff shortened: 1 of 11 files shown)


lib/@vibe/compiler/core/import_alias_rewrite.vibe
```diff
@@ -696,26 +696,29 @@ fn rewrite_alias_expr_ix(expr: Expr, am: AliasIdx) -> Expr {
...
 
+fn collect_item_aliases(items: Array[ImportItem], aliases: Array[(String, String)]) -> Unit {
+  let mut j = 0
+  while j < Array::length(items) {
+    let item = Array::get(items, j)
+    let original = item.name
+    let alias_opt = item.alias
+    match alias_opt {
+      Some(alias) => if alias != original {
+        Array::push(aliases, (alias, original))
```
mizchi: **[bug]** `collect_item_aliases` records one hop `(alias, item.name)` and skips unaliased items. That is enough for `export ./dep { cmp as compare }` because `cmp` is the def. It is not enough once the immediate original is itself a re-export public name.

```
// a.vibe
export fn foo(a: Int, b: Int) -> Int { a + b }
// b.vibe
export ./a.vibe { foo as bar }
// entry
export ./b.vibe { bar as baz }
test { assert_true(baz(1, 2) == 3) }
```

Typecheck accepts `baz`. Merge rewrites `baz` → `bar` and codegen ICE's on unresolved `bar`. The unaliased sibling `export ./b.vibe { bar }` then `bar(1, 2)` ICE's the same way. Importing the chain from another file succeeds because `build_export_rename_plan` already walks these edges to depth 16.

**Suggestion:** Resolve each re-export local to the origin def using the edge graph `collect_reexport_edges_into` already builds. Rewrite the entry's uses to that origin, including unaliased `export ./mid { public }`. Pin both programs next to `fixtures/reexport_alias_test.vibe`.

lib/@vibe/compiler/core/import_alias_rewrite.vibe
```diff
@@ -696,26 +696,29 @@ fn rewrite_alias_expr_ix(expr: Expr, am: AliasIdx) -> Expr {
...
-            Some(alias) => if alias != original {
-              Array::push(aliases, (alias, original))
-            },
-            None => ()
-          }
-          j = j + 1
-        }
-      },
+      SImport(_, items) => collect_item_aliases(items, aliases),
+      SReExport(_, items) => collect_item_aliases(items, aliases),
```
mizchi: **[bug]** `SReExport` aliases now share the per-file alias list with `SImport`, but `collect_import_value_renames` still walks only `SImport`. A later re-export of the same local is what typecheck last-binds (`env_bind` prepends; `env_lookup` is first-match), while rewrite drops the re-export mapping because that local is in `extra_renames` and follows the import.

```
import ./a.vibe { x as foo }   // x = a - b
export ./b.vibe { y as foo }   // y = a + b
test { assert_true(foo(1, 2) == 3) }
```

Compiles. `foo(1, 2) == 3` traps; `foo(1, 2) == -1` passes. The program is silently the import, not the re-export the checker just bound.

**Suggestion:** Either reject a local bound by both an import and a re-export (fail closed), or make rewrite follow the same last-bind order as the checker. Pin `x` vs `y` with a non-commutative operator.

mizchi:
Addressed the two review bugs on `cf47fb54e`:

- Nested `bar as baz` / unaliased `export ./mid { bar }` now walk the rename-table closure to the origin def (identity-seeded when the origin stays unmangled). Pins: `fixtures/reexport_nested_alias_test.vibe`, `fixtures/reexport_unaliased_nested_test.vibe`.
- Import + re-export of the same local last-binds, matching the checker. Pin: `fixtures/reexport_import_collision_test.vibe` (`a-b` vs `a+b`).
- Codex non-entry facade (`cmp as compare` used in-file): `collect_import_value_renames` now walks `SReExport`. Pin: `fixtures/reexport_facade_use_test.vibe`.

Could not reply in the pending review threads (GitHub 422: one pending review per user). Dismiss or submit that PENDING review on the Files tab.

mizchi:
Not merging. #2287 already closed via #2307 (`884541600`) and #2323: an aliased re-export is publication-only and does not bind a local. This PR rewrote the alias into local scope (`compare` calls `cmp`), which is the opposite rule.

Codex on `cf47fb54e` is the same point (`Keep aliased re-exports out of local scope`). Rebasing would only delete the change. Closing as superseded.

Issue #2305, opened 2026-08-25
The ADR-0068 closure gate does not exist in the split CLI (lib/@vibe/cli), and its cached lane cannot ask the question the adapter asks


mizchi:
Follow-up to #2284 / PR #2302 (review thread [r3855637717](https://github.com/mizchi/vibe-lang/pull/2302#discussion_r3855637717)).

## Not a live bypass today

The gate added in #2302 lives in `lib/@vibe/compiler/cli_adapter.vibe`. The split CLI (`lib/@vibe/cli/entry.vibe` → `dispatch.vibe`) has no equivalent, so its `check` / `compile` / `build` / `serve` handlers would accept a program whose *dependency* imports `@vibe/concurrent` without `VIBE_UNSTABLE=1`.

That is not reachable through the shipped compiler. Measured on `266b74cd`:

- `dist/cli/vibe-cli.wasm` is `generations.sh` stage2 (`scripts/build_cli_wasm.sh:75`), and stage2 is compiled from the flat adapter source, not from `lib/@vibe/cli/`. That source contains **zero** `lib__vibe_cli_*` symbols — `lib/@vibe/cli/` is entirely absent from the artifact. (The 36 `*_dispatch_vibe` symbols in it are all `lib/@vibe/parser/parser_expr_dispatch.vibe`.)
- `scripts/generations.sh:784` refuses to build that entry at all while the seed is `cli_support.vibe`: `split CLI generation requires a bootstrap bump first`.
- The split CLI is buildable only via `scripts/build_cli_core.sh` / `scripts/vibe_cli.sh` (`lib/@vibe/cli/main.vibex`, the #1137 stage-1 lane).

So the gap is latent, not live — but at the #1137 cutover the gate would disappear with nothing to catch it, which is the silent-wrong failure class the gate exists to prevent.

## The part that is not a 30-line port

`selfhost_cli_build_args` / `selfhost_cli_compile_args` / `selfhost_cli_serve_args` compile through `compile_release_file_mode` → `compile_file_fs_mode_cached`, and that function returns early on a persistent-artifact hit:

```
compile_file_fs_mode_cached_observed (file_compile.vibe:863)
  load_persistent_source_group_fingerprint_if_valid(entry_path)
    Some(fp) -> load_persistent_artifact(kind, entry_name, fp)
                  Some(bytes) -> (bytes, db, ...)      // returns WITHOUT loading
```

[...]

Issue #2300, opened 2026-08-25
The `assert_eq` shadowing guard is linear-only: the wasm-gc backend still hijacks a user's own definition


mizchi:
Raised in the #2296 review, confirmed by measurement, and left unfixed there on purpose — the naive fix is known to break the lane.

## The asymmetry

#2296 makes a user-defined `assert_eq` win over the builtin lowering. It only works on the linear lane. Same file, both lanes:

```vibe
fn assert_eq(a: Int, b: Int) -> Int {
  a + b
}

test "own assert_eq" {
  assert_true(assert_eq(1, 2) == 3)
}
```

| lane | result |
|---|---|
| `vibe test` (linear) | `1 passed` — the user's function runs |
| `VIBE_TEST_BACKEND=gc vibe test` | `trap: wasm 'unreachable' instruction executed` |

No diagnostic either way on gc. The desugar guard stands down correctly (that part is backend-independent), but `codegen/gc/backend_call.vibe`'s `assert` / `assert_true` / `assert_eq` arms are unconditional, so the call is compiled as the builtin comparison and traps.

Not a regression: before #2296 both lanes hijacked the call. It is a partial fix that leaves the lanes inconsistent.

## Why the obvious fix does not work

The linear lane guards with `fn_idx_in_list < 0` — a name in the function table is a real top-level definition. Applying the same test to the gc arms was measured to break the lane outright:

```
$ VIBE_BACKEND=gc … fixtures/gc_backend_smoke_test.vibe
CompileError: WebAssembly.instantiate(): Compiling function #62 failed:
expected 1 elements on the stack for fallthru, found 3
```

On that lane the func table **contains the generated builtin names** — `resolve_func(ctx.func_table, "Array::push")` is how its own arms find them — so the guard fires for every ordinary `assert_eq`, not just a shadowed one, and the arm stops emitting its result.

`FuncTable` (`codegen/common_base/common_base.vibe`) carries `names` / `indices` / `params` / `returns` / `needs_env` and no user-vs-generated marker, so the distinction is not available where the guard needs it.

## Candidate fixes

[...]

Issue #2289, opened 2026-08-25
An inherited unstable import consumes the sibling's physical offset, so the diagnostic names the wrong edit


mizchi:
Raised in the #2277 review, against the location cursor added in `7dab582b`.

## The shape

A package's `index.vpkg` shares `import @vibe/concurrent` with its siblings, **and** a sibling also spells its own import of that package. Ingestion prepends the inherited import, so the ingested statement list holds two `SImport`s for the package: the synthetic one first, the physical one second.

The offset scan only ever sees the **physical** file. Walking in statement order, the inherited import is visited first and consumes the physical import's offset. Result:

- the first diagnostic points at the sibling's own `import` line, instead of saying the dependency is inherited;
- delete that line as instructed, and a *second* error appears for the contract import that was never mentioned.

A diagnostic has to name the edit that fixes it (CLAUDE.md's CLI policy). This one names an edit that only reveals another error.

## Why it is not already covered

The gate case for repeated imports (section 108) uses two *physical* imports, where the cursor is exactly right. The inherited case needs a `.vpkg` in the fixture, which is why it slipped past.

Note the configuration is already flagged as a redundant import by `detect_redundant_namespace_imports` — so the reader sees a second, correct warning about it. That is why this is P2 rather than P1: the state is reachable and the message is misleading, but it is not silent.

## Proposed fix: assign by count, not by order

The synthetic prefix always comes first, and physical declarations always come last, so the assignment is determined:

- for each unstable package `p`, let `N_p` be the number of `SImport`s naming it in the ingested statements, and `P_p` the number of matching `import` token pairs in the physical file;
- the first `N_p - P_p` occurrences are inherited → report with no location and the "inherits from its directory's index.vpkg" note;
- the remaining `P_p` occurrences take the physical offsets in order, exactly as the cursor does today.

That needs a pre-pass to count per package, which is why it was not folded into the cursor change: the cursor is correct on its own for the all-physical case, and this is a separate piece of bookkeeping.

An alternative that avoids counting is to walk the statement list in reverse, matching physical offsets from the end and letting the leftovers at the front fall out as inherited, then emit the collected diagnostics in source order. Same result, different cost.

## Regression to add

`tests/gates/late/run.sh` section 108, with a `.vpkg` that shares the import plus a sibling that repeats it: the physical line must be named exactly once, and the inherited one must be reported as inherited rather than pointing at a line.

Issue #2288, opened 2026-08-25
`assert_eq` inside a container expression loses its `path:line`


mizchi:
Raised in the #2277 review against `core/assert_stamp.vibe`.

The stamp walk descends only the expression forms that carry statements — `ESeq`, `EIf`, `ELet`, `EMatch`, `EFn`, the assignment continuations, and so on. Everything else falls to a catch-all that returns the subtree untouched, so an `assert_eq` nested inside a container never gets stamped:

```vibe
let result = (assert_eq(actual, expected), 0)
```

reports `assert_eq failed` with no location, while the same call written as a statement reports `assert_eq failed at path:line`.

Affected forms include tuple and array literals, record/map fields, binary and unary operands, index expressions, and loop initializers.

## Not a safety limit

The file header used to justify the restriction as degrading "rather than borrowing someone else's line". That reasoning is wrong, and the comment has been corrected: a nested `ECall` carries **its own** offset, so a stamp taken there is that call's own position — there is nothing to borrow. `is_two_arg_assert_eq` already declines when the offset is `-1`, which is the real safety property and is independent of which forms the walk visits.

So this is purely missing coverage.

## Shape of the fix

Rebuild each remaining `Expr` variant with its children stamped, the way the `ECall` / `EIf` / `ESeq` arms already do. `core/expr_children.vibe` enumerates the same set for read-only traversal and is the reference for what must be covered — but it cannot be reused directly here, because stamping rebuilds the tree rather than just visiting it.

Worth doing carefully rather than quickly: every arm has to re-emit its constructor's fields in the right order, and a transposed pair is silent. An earlier arm in this same file shipped with `EAssignOp`'s binders reversed and was caught by the `review-regressions` lint, not by behaviour.

## Regression to add

`tests/gates/late/run.sh` section 110 already checks the located form for a statement call and after an assignment. Add the container shapes — at minimum a tuple element and a binary operand — asserting the failure names the line.

mizchi:
Does not reproduce. Closing on measurement.

Both files this issue describes are absent from the tree: there is no `lib/@vibe/compiler/core/assert_stamp.vibe` and no `is_two_arg_assert_eq` anywhere outside the generated bundles. The walk it describes is not what ships — the location is taken in `normalize/desugar_trait_dict.vibe`, from the `ECall`'s own offset, at the point the `assert_eq` is expanded. There is no catch-all to fall through.

[...]

PR #2312, opened 2026-08-25
Four surfaces that answered about a different program than the one on disk

### LINKED ISSUE #2309: A binding named `eq` is silently replaced by the builtin when either argument is scalar
Found while measuring #2285. Measured on a stage2 from `b1d42642`:

```vibe
fn main() -> Unit with Stdout {
  let eq = (a: Int, b: Int) -> Int { a + b }
  println(Int::to_string(eq(1, 2)))
}
```

| expected | actual |
|---|---|
| `3` | `0` |

No diagnostic, no trap. The checker types the call against the user's `eq` (`Int`), codegen emits the builtin `i64.eq`, and `1 == 2` prints as `0`. Wrong value, silently — P0 by the triage rule.

## Cause

`compile_call.vibe:1631`:

```vibe
else if fname == "eq" && Array::length(args) == 2
  && (!prefer_bound_call || cc_arg_is_scalarish(Array::get(args, 0)) || cc_arg_is_scalarish(Array::get(args, 1))) {
```

Every other bare-name scalar builtin (`not`, `add`, `sub`, `mul`, `lt`) is guarded by `!prefer_bound_call` alone and shadows correctly — measured: `let not = (a: Int) -> Int { a + 41 }; not(1)` gives `42`. `eq` is the only one with a disjunct that lets the builtin win **over** a binding.

The disjunct is deliberate, not an oversight: the in-tree comment records that dropping it regressed `leb128_test.vibe` because `prefer_bound_call` is a whole-program name match rather than a file-scoped one, and it points at #711 for the root fix. So the scalarish escape hatch exists to work around an over-broad `prefer_bound_call`, and it pays for that with a silent wrong answer whenever someone binds the bare name `eq` and calls it on scalars.

## Shape of a fix

The two readings of `prefer_bound_call` have to be separated before this arm can be tightened — `fn_idx_in_list` (a real definition, possibly in another file: the #711 over-broad half) versus `local_idx_hint` (a lexical binding here, which is unambiguous). A local binding can win today without touching the #711 problem at all:

```vibe
else if fname == "eq" && Array::length(args) == 2 && local_idx_hint < 0
  && (!prefer_bound_call || cc_arg_is_scalarish(..) || cc_arg_is_scalarish(..)) {
```

That is the same split #2285 relies on, and reading `local_idx_hint` is only sound for names the capture scan skips — `eq` is unconditionally registered in the func table for every compile (per the `is_inlined_scalar_builtin` doc comment), so it never becomes the #1095 capture artifact. Wants measurement against `leb128_test.vibe` before landing.

mizchi:
Closes #2309, #2300, #2280, #2289.

Four surfaces that each answered about a different program than the one on disk. Two grew during review — the sections below describe what actually landed, not what was opened.

## #2309 (P0) — a binding named `eq` is silently replaced by the builtin

No diagnostic, no trap. The checker types the call against the user's `eq` and agrees the result is an `Int`; codegen emits `i64.eq` anyway. Filed for a lexical binding; measurement and two Codex review rounds found it in four more places.

| shape | before | after |
|---|---:|---:|
| `let eq = (a: Int, b: Int) -> Int { a + b }` | `0` | `3` |
| `let eq = (a: Double, b: Double) -> Double {..}`, **RC** | `0` | `4` |
| the same on the non-RC lane | `4` | `4` |
| **top-level** `fn eq(a: Int, b: Int) -> Int` | `0` | `3` |
| gc lane, local and top-level `eq` | trap | pass |
| gc lane, local `not` | trap | pass |
| builtin `eq(1,2) eq(7,7) eq(0.5,0.5) eq(0.5,0.25)` | `n y y n` | `n y y n` |

The RC float arm had **no shadowing guard at all**, and misfires only under RC — the production lane. A non-RC probe would have passed while every shipped build was wrong.

### Why the disjunct stays

The fix does not remove `cc_arg_is_scalarish`. That is the deliberate #710/#711 workaround: `prefer_bound_call` is a whole-program name match rather than a file-scoped one, and dropping it regressed `leb128_test.vibe`.

Instead the flag's two readings are separated, and the missing third distinction is supplied:

- `local_idx_hint` — a lexical binding in this closure. Unambiguous, so it wins. Sound for the same reason as #2285: `eq` is in both func tables unconditionally, so the capture scan skips it and no lambda records the #1095 phantom capture.
- `fn_idx_in_list` — consulted exactly as before. It cannot decide the top-level case at all: `eq` is registered unconditionally, so that index is `>= 0` for **every** call whether or not anyone defined one. That is *why* the disjunct had to exist.
- **`func_table_user_index_of`** (new, `common_base`) — "a definition belonging to the current program", which is what #711 asks for. Both backends fill `fn_names_list` from the program and only then append `builtin_defs`, so user entries are exactly the prefix `[0, num_user_funcs)`.

It scans that prefix rather than going through `func_table_index_of`. When a user shadows a builtin the table holds **both** entries under one name, and that lookup answers with whichever branch it takes — the gc `FuncTable` literals leave `names_sorted` empty so it is the first-match branch today, but resting a correctness answer on that is resting it on sort stability.

## #2300 (P1) — the same guard was linear-only

`fn assert_eq(a: Int, b: Int) -> Int { a + b }` ran the user's function under `vibe test` and trapped `unreachable` under `VIBE_TEST_BACKEND=gc`. One language, two answers, no diagnostic on either.

Porting linear's guard verbatim breaks the lane, exactly as the issue records: `assert*` are in the **gc** func table (`in_gc_table = true`, `in_linear_table = false`), so `fn_idx_in_list < 0` fires for every call and the arms stop emitting their result (`expected 1 elements on the stack for fallthru, found 3`). The gc predicate uses `func_table_user_index_of` too, so both lanes now answer the same question the same way.

Both directions are pinned, because a green shadow probe next to a broken builtin probe is precisely the state the naive fix produces.

## #2280 (P1) — `vibe check` could not check the file that *is* the boundary

```console
$ vibe check lib/@vibe/random/index.vpkg
line 15:1: expected { but got fn
```

A committed contract the build consumes on every run; line 15 is an ordinary bodyless declaration. The message does not merely fail to help — it asserts something untrue about the file.

Every compile lane reads a contract through `ingest_source_text_fs`, which substitutes the generated facade before anything parses it. The check lane read raw bytes. It now ingests too — **contracts only**, because the other ingestion branch can *prepend* imports to an ordinary `.vibe` and shift every offset this pre-parse exists to make exact.

After: the contract checks clean, and a package declaring a name no sibling implements is still rejected, naming it.

## #2289 (P2) — an inherited unstable import consumed the sibling's line

Ingestion prepends a directory's shared `index.vpkg` imports, so a sibling that also spells its own has two `SImport`s for one package: synthetic first, physical second. The offset cursor only sees the physical file, so the synthetic one consumed the physical one's offset — the reader was told to delete their own `import` line, and deleting it revealed a second error nobody had mentioned.

Assigned **by count, not by order** now. The token count is driven by `unstable_import_offset` itself rather than a second copy of its matcher: the count decides which occurrences are inherited, so two matchers disagreeing about a spelling would silently mislabel one.

### Two things found only by measuring this

- **Only the first diagnostic was ever emitted** — `Array::get(warns, 0)` at **six** call sites. The misassignment was not mislabelling one of two reports, it was choosing which single report you saw. All six now emit every diagnostic, per the rule #1567 settled for parse errors.
- **Fixing that exposed a flood.** The loader ingests a module several times per compile, and a shared import reaches every sibling: measured **ten** reports for one `import` line in a 3-member package. Deduped on two axes — an inherited report keys by `(directory, diagnostic)` since the file differs per sibling but the edit does not; everything else keys by `(file, diagnostic)`. The duplicate-ingestion count is not even stable between runs, so this is what makes an exact assertion meaningful rather than flaky.

## Verification

All 28 CI checks green. Locally, through a stage2 built from this checkout:

- **`cmp stage2 stage3` → `FIXPOINT_OK`**
- Every row of the #2309 table above, both lanes
- `leb128_test.vibe` — the #710/#711 regression — 16/16
- A program importing `@vibe/semver`, which exports its own top-level `eq`: still `y n`. If the guard were a whole-program name match it would stand the builtin down there, which is #711's mistake in a new place.
- Gate 113 **red-tested under real lane conditions**: fails `reported 6 unstable imports, want exactly 2` on the pre-fix compiler, passes after.

Perf: selfcompile heap **+0.04%**, execution fuel/heap ±0, output checks 9/9. The new prefix scan runs per `eq` call site and costs nothing measurable.

Gate sections 110 (extended to both lanes), 111, 112, 113.

## Also measured, not changed here

- **#2288 does not reproduce and is closed.** It describes a stamp walk in `core/assert_stamp.vibe` with an `is_two_arg_assert_eq` predicate; neither exists in the tree. The location comes from the `ECall`'s own offset, so a tuple element, array literal, call argument and struct field are each stamped with their own position.
- **#2305's cache concern is latent, not live.** Warming the artifact cache under `VIBE_UNSTABLE=1` and then compiling without it still refuses, and still names the sibling. Table posted on that issue as a fixed reference for the split-CLI port.
- **`print`/`println` are equally bare and unguarded on the gc lane.** Left alone deliberately: a guard on a name the prelude may itself define is the shape that breaks a lane outright, and I have not measured it.

https://claude.ai/code/session_01GbtLER6wNT4jsZEN1Zp6xe

### DIFF
--- a/lib/@vibe/compiler/cli_adapter.vibe
+++ b/lib/@vibe/compiler/cli_adapter.vibe
@@ -769,8 +769,16 @@ fn adapter_serve_component(input_path: String, output_path: String) -> Int with 
     } else {
       check_unstable_surface_warnings_from_path(input_path)
     }
+    // #2289: report EVERY unstable import, not `Array::get(.., 0)`. A file can
+    // depend on the package twice -- once inherited from its directory's
+    // index.vpkg, once spelled physically -- and each needs a separate edit.
+    // Emitting only the first told the reader to make one of the two changes,
+    // and making it revealed the other. This is the same rule #1567 settled
+    // for parse errors: the collection already holds them all, so dropping
+    // every one but the first costs an edit-and-rerun cycle per extra
+    // diagnostic. Five sibling call sites below do the same.
     if Array::length(serve_unstable) > 0 {
-      return emit_compile_diag(output_path, Array::get(serve_unstable, 0))
+      return emit_compile_diag(output_path, String::join(serve_unstable, "\n"))
     } else {
       ()
     }
@@ -798,7 +806,7 @@ fn adapter_serve_component(input_path: String, output_path: String) -> Int with 
       check_unstable_surface_warnings_from_closure()
     }
     if Array::length(serve_closure_unstable) > 0 {
-      return emit_compile_diag(output_path, Array::get(serve_closure_unstable, 0))
+      return emit_compile_diag(output_path, String::join(serve_closure_unstable, "\n"))
     } else {
       ()
     }
@@ -1342,7 +1350,7 @@ export fn cli_main() -> Int with Exception + Fs + Env + Stdin + Stdout {
         check_unstable_surface_warnings_from_closure()
       }
       if Array::length(unstable_warns) > 0 {
-        return emit_compile_diag(output_path, Array::get(unstable_warns, 0))
+        return emit_compile_diag(output_path, String::join(unstable_warns, "\n"))
       } else {
         ()
       }
@@ -1383,7 +1391,7 @@ export fn cli_main() -> Int with Exception + Fs + Env + Stdin + Stdout {
         check_unstable_surface_warnings_from_path(input_path)
       }
       if Array::length(fs_unstable) > 0 {
-        return emit_compile_diag(output_path, Array::get(fs_unstable, 0))
+        return emit_compile_diag(output_path, String::join(fs_unstable, "\n"))
       } else {
         ()
       }
@@ -1483,7 +1491,7 @@ export fn cli_main() -> Int with Exception + Fs + Env + Stdin + Stdout {
         check_unstable_surface_warnings_from_closure()
       }
       if Array::length(closure_unstable) > 0 {
-        return emit_compile_diag(output_path, Array::get(closure_unstable, 0))
+        return emit_compile_diag(output_path, String::join(closure_unstable, "\n"))
       } else {
         ()
       }
@@ -1942,7 +1950,7 @@ export fn cli_main() -> Int with Exception + Fs + Env + Stdin + Stdout {
     unstable_surface_diagnostics(input_path, source)
   }
   if Array::length(bare_unstable) > 0 {
-    return emit_compile_diag(output_path, Array::get(bare_unstable, 0))
+    return emit_compile_diag(output_path, String::join(bare_unstable, "\n"))
   } else {
     ()
   }
--- a/lib/@vibe/compiler/cli_support.vibe
+++ b/lib/@vibe/compiler/cli_support.vibe
@@ -392,7 +392,27 @@ export fn check_linked_file_source_groups(input_path: String) -> Array[(String, 
   // made `vibe check` report a parse error for a file the build compiles.
   // extract_require_pins blanks to spaces (byte length preserved), so every
   // diagnostic offset below stays exact.
-  let (_, entry_source) = extract_require_pins(Fs::read_file(input_path))
+  //
+  // #2280: a CONTRACT (`.vpkg`) is not statement grammar -- it is a list of
+  // bodyless declarations, and every compile lane reads it through
+  // `ingest_source_text_fs`, which replaces it with the generated facade
+  // before anything parses it. This lane read the raw bytes and so answered
+  // `expected { but got fn` about `lib/@vibe/random/index.vpkg`, a committed
+  // contract the build consumes on every run: a diagnostic that asserts
+  // something untrue about the file. Ingest contracts here too, so `check`
+  // answers about the package the contract describes.
+  //
+  // Contracts ONLY. `ingest_source_text_fs`'s other branch
+  // (`apply_vpkg_directory_shared_imports_fs`) can PREPEND import lines to an
+  // ordinary `.vibe` file, which would shift every offset below and mislocate
+  // the diagnostics this pre-parse exists to locate exactly.
+  let raw_entry_source = Fs::read_file(input_path)
+  let entry_source = if is_contract_ext(input_path) {
+    ingest_source_text_fs(input_path, raw_entry_source)
+  } else {
+    let (_, pin_blanked) = extract_require_pins(raw_entry_source)
+    pin_blanked
+  }
   let (tokens, starts, _ends) = lex_with_offsets(entry_source)
   ingestion_pipeline_telemetry_add(14, 1)
   let (entry_stmts, parse_diags) = parse_program_recovering(tokens, starts, entry_source)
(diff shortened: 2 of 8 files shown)


mizchi:
Carrying over the Codex P1 from #2311, since that PR merged before it could be acted on there: *"Guard local assertions in the GC backend"* — `let assert_true = (a: Int) -> Int { a + 41 }; assert_true(1)` returning `0` instead of `42`, and local `assert` / `assert_eq` likewise hijacked.

That is #2300, which this PR closes. The guard's first test is `find_local_slot`, so the local case was already covered by the fix as pushed; measured on this branch's stage2, both local shapes pass on the gc lane.

What was missing is the second half of the review's ask — *"add the local-binding probe to a GC gate"*. Section 110's gc probes only exercised a top-level `fn assert_eq`, so the local half of the guard was unpinned. `e89b4b17` adds both, red-tested: each fails on a stage2 carrying #2309 but not #2300, and passes after.

| gc probe | pre-#2300 stage2 | this branch |
|---|---|---|
| `let assert_true = (a: Int) -> Int { a + 41 }` | 0 passed, 1 failed | 1 passed |
| `let assert_eq = make_cmp()` | 0 passed, 1 failed | 1 passed |

---

mizchi:
Confirmed, and the first fix I pushed for it was not enough.

I found the duplication independently while verifying `e9f8d54b` and deduped by exact report text in `4a12328c`. That collapses one file ingested several times per compile — but not this, because each sibling has a different path, so the reports are not identical. Measured on a 3-member package with the shared import declared in `index.vpkg`, checking the contract:

```
10 reports for one `import` line
   2 × pkg/a.vibe: ... this file inherits `@vibe/concurrent` ...
   2 × pkg/b.vibe: ... this file inherits `@vibe/concurrent` ...
   2 × pkg/c.vibe: ... this file inherits `@vibe/concurrent` ...
   4 × pkg/index.vpkg: line 9:1: ...        <- the physical one, correctly located
```

Text dedupe alone takes that to 4, which is still three reports for one edit.

So the key is now per-axis, which is your "by their owning contract, retaining separately written physical imports":

- an **inherited** report keys by `(directory, diagnostic)` — the file differs per sibling, the edit does not;
- everything else keys by `(file, diagnostic)`, so two siblings that each physically spell the import still report separately, and a single file's inherited and physical reports stay distinct (they differ in the note).

That gives 1 inherited + 1 physical at `index.vpkg: line 9`.

The inherit marker is now a shared constant rather than a re-typed substring in the dedupe. Keying off a second copy would silently stop collapsing the day the wording changed, and the failure mode there is a flood, not an error — nothing would catch it.

Gate 113 grew a probe for exactly this shape: a 3-member package, asserting exactly one inherited report and exactly one for the contract's own line, at line 9.

Two notes on scope. The physical report already names `index.vpkg: line 9`, which *is* the edit, so the surviving inherited report is arguably redundant when the contract is in the closure — I left it, because when a sibling is checked on its own the contract is not loaded and that report is the only signal. And the inherited report still names one arbitrary sibling rather than the `.vpkg`; re-pointing it would mean rewording the note, which is more than this fix needs.

---

PR #2323, opened 2026-08-26
Keep re-exported declaration aliases out of local scope

### LINKED ISSUE #2310: Keep re-exported type and trait aliases out of local declaration scope
## Problem

PR #2307 separates aliased value re-exports from local value lookup, but declaration projections still have no equivalent publication-only origin.

Measured on 2026-08-26 with a fresh stage2 from `142a1aa30e`:

```vibe
export ./dep.vibe { type Original as PublicT }
let local: PublicT = 1
```

and:

```vibe
export ./dep.vibe { trait Marker as PublicMarker }
fn local[T: PublicMarker](value: T) -> Int { 1 }
```

Both check clean in the middleman even though an aliased re-export introduces no local binder. A downstream consumer importing `PublicT` / `PublicMarker` through the middleman also checks clean, and that consumer behavior must remain valid.

## Required contract

- aliased re-exported TypeDef / trait authority remains available to `checked_module_public_env`
- local annotation and bound resolution in the re-exporting module cannot select the alias
- a downstream explicit import can select the published alias normally
- plain non-aliased re-exports retain their current compatibility behavior
- declaration closure needed to interpret a published signature remains available without granting local spelling authority

This requires a declaration-level equivalent of `BindingOriginReExportSurface`, not another value-binding special case.

Follow-up from #2307 and review discussion https://github.com/mizchi/vibe-lang/pull/2307#discussion_r3856227391.

mizchi:
## Summary

- separate aliased re-export TypeDef and trait projections from the environment used to check the re-exporting module
- reattach declaration closure only while constructing the published module environment
- preserve that publication-only projection in both filesystem and in-memory TypeDb lanes
- diagnose a publication-only trait alias at its local generic-bound declaration site
- retain downstream imports, declaration closure, and plain re-export compatibility

## TDD

Red:
- `export ./dep.vibe { type Original as PublicT }` allowed `PublicT` in a local annotation
- `export ./dep.vibe { trait Marker as PublicMarker }` allowed `PublicMarker` in a local generic bound
- the in-memory TypeDb lost aliased type and trait declarations before a downstream consumer imported them

Green:
- local type use reports `unknown type PublicT`
- local trait-bound use reports `unknown trait: PublicMarker`
- downstream consumers can still import both aliases through filesystem and in-memory TypeDb lanes
- private TypeDef closure remains available to interpret the published alias
- plain type and trait re-exports retain their local compatibility behavior

## Validation

- focused `type_db_cross_module_test.vibe`: 99/99 tests passed
- `pkf run test-affected`: 173/173 affected files passed
- `pkf run pre-commit`: passed
- fresh stage0 -> stage3 generation: passed
- compiler gate: all 106 checks passed
- `pkf run full-gate` reached the final gate self-tests; macOS system Bash 3 lacks `mapfile`, so the two Bash-4 self-tests could not run under the task wrapper

Closes #2310

### DIFF
--- a/lib/@vibe/compiler/checker/checker.vibe
+++ b/lib/@vibe/compiler/checker/checker.vibe
@@ -9507,6 +9507,26 @@ fn check_expr_capture_impl(expr: Expr, env: TypeEnv, defs: Array[TypeDef], s: Su
       }
     },
     EFn(tparams, tbounds, params, ret_ty, effect, body) => {
+      // An aliased re-export publishes a trait spelling but does not bind it
+      // in this module (#2310). The import projection retains an origin marker
+      // so this declaration site can report the invalid bound immediately,
+      // instead of accepting the generic function until a concrete call later
+      // fails with a misleading missing-impl error.
+      let mut rti = 0
+      while rti < Array::length(tbounds) {
+        let (_, declared_bounds) = Array::get(tbounds, rti)
+        let mut rbi = 0
+        while rbi < Array::length(declared_bounds) {
+          let bound = Array::get(declared_bounds, rbi)
+          if env_has_reexport_surface_binding(env, bound) {
+            Array::push(errors, String::concat("unknown trait: ", bound))
+          } else {
+            ()
+          }
+          rbi = rbi + 1
+        }
+        rti = rti + 1
+      }
       let tplen = Array::length(tparams)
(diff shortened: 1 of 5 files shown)


mizchi:
@codex review

mizchi:
@codex review

PR #2326, opened 2026-08-26
Classify unkinded import aliases by source name

### LINKED ISSUE #2325: Unkinded aliased type imports are misclassified as value collisions
## Symptom

The duplicate value-import check merged in #2306 classifies every aliased,
unkinded import as a value. That rejects two imports of the same aliased type
head:

```vibe
import @vibe/core { HashMap as Shared }
import @vibe/core { HashMap as Shared }
fn main { 0 }
```

Current diagnostic:

```text
duplicate declaration: Shared
```

This violates #2293's requirement to skip PascalCase type heads. The ambiguity
must be resolved from the source name, not the local alias: `HashMap as Shared`
is type-shaped, while `abs as Abs` is still a value import and two such aliases
must be diagnosed.

## Acceptance criteria

- Two unkinded PascalCase source imports aliased to the same local name do not
  produce a duplicate value declaration diagnostic.
- Two lower-case value imports aliased to the same PascalCase local name still
  produce `duplicate declaration: <name>`.
- The projected cross-module collision regression from #2306 stays green.

Follow-up to #2293 / #2306.

mizchi:
## Summary

- classify an unkinded import from its source name before applying a local alias
- keep `HashMap as Shared` out of the duplicate value-import set
- keep lower-case value imports such as `abs as Abs` in that set

## Why

#2306 correctly rejected duplicate value imports, but its alias branch treated
every aliased unkinded import as a value. That made two imports of an aliased
PascalCase type head fail with `duplicate declaration`.

## Validation

- Red: the new aliased type-head regression trapped against the #2306 stage2
- Green: focused compiler tests passed (2 files, 110 tests)
- `pkf run test-affected` promoted to the full battery and passed 1027/1027 files
- `pkf run ensure-generated`
- pre-commit structural gates

Closes #2325.

### DIFF
--- a/lib/@vibe/compiler/checker/checker_stmt.vibe
+++ b/lib/@vibe/compiler/checker/checker_stmt.vibe
@@ -776,15 +776,12 @@ fn duplicate_import_value_note(seen: Array[String], item: ImportItem, errors: Ar
     ()
   } else {
     let local = import_bound_name(item)
-    // Unaliased PascalCase is the unkinded type-head form (`import { HashMap }`).
-    // An alias is a value local even when it is PascalCase (`abs as Abs`).
-    match item.alias {
-      Some(_) => duplicate_value_decl_note(seen, local, errors),
-      None => if name_is_type_head(local) {
-        ()
-      } else {
-        duplicate_value_decl_note(seen, local, errors)
-      }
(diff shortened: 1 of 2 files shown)


mizchi:
@codex review

Issue #2199, opened 2026-08-22
OOB aborts identify operation/index/length but lack source provenance


mizchi:
## Current problem

Out-of-bounds aborts now identify the operation, index, and length, and the
compiler-developer heap dump is opt-in. A normal user-facing abort still lacks
the editable source location of the offending access.

Current default shape:

```text
Array::get: index 10 out of bounds for length 3
RuntimeError: unreachable
    at wasm://wasm/...:wasm-function[...]:0x...
```

The message explains the mistake but not where to edit it. Trapping remains the
correct behavior; this issue is only about provenance.

## Already complete

- #2224: shared OOB aborts name `Array::get` / `Array::set` / `Bytes::get` /
  `Bytes::set` / `String::byte_at`, plus index and length.
- `8cb00072`: `[crash debug]` heap and memory dumps are gated behind
  `VIBE_CRASH_DEBUG=1` (and the existing explicit debug aliases).

## This does not ride #1987

#1987 records the runtime *freeing site* in the shadow-RC lane so a later
dup/drop can name both ends. An OOB access already has its source offset at
compile time. The mechanisms share a diagnostic goal, not an implementation.

The closer precedent is the compile-time `path:line` stamping used by
`assert_eq`. A literal per-site string is not acceptable as the default here:
the compiler source contains roughly 9,000 guarded accesses, and the measured
estimate was about 350 KiB of distinct strings before code overhead against a
roughly 2.4 MiB stage2.

## Preferred direction

Extend the compact `vibe.linemap` shape so the runner can map the trap's wasm
function/code offset back to a source location without a distinct string at
every access. Today that section is coupled to `VIBE_DEBUG_BREAK=1` and break
probe sites; ordinary OOB sites are not recorded. Decoupling and compacting the
source-location table is the implementation work here.

This should be designed as general trap provenance, even though OOB aborts are
the acceptance case. Out-of-band assert provenance and removal of seed-era text
recognizers remain separate in #2219.

## Acceptance criteria

- a normal production-style build reports an editable `path:line` for an OOB
  access without a debug flag;
- the mapping covers at least the shared OOB abort sites listed above;
- no per-callsite source string is emitted into ordinary code;
- stage2 size impact is measured and kept within the existing size gate;
- missing or stripped mapping data degrades to the current self-consistent wasm
  frame, never to a wrong source location;
- linear and wasm-gc lanes either agree or state a gated backend limitation.

mizchi:
[...]

Issue #1987, opened 2026-08-16
Shadow RC: report the FREEING site's provenance at the offending dup/drop, not just the trap location


mizchi:
`VIBE_RC=shadow` was the decisive tool in the #1964 investigation: where the plain RC lane surfaced free-list corruption several calls downstream (`__rt_rc_alloc` OOB in an unrelated function), the shadow lane trapped deterministically at the first guarded `rc_dup` of an already-freed value, with a usable stack.

What it did NOT answer is **who freed the value** — the trap names the reader, not the earlier over-release. In #1964 that gap is exactly why the general-merge under-provision was never root-caused: we know `railway_rw`'s EFn-rebuild dup'd a freed node, but the mis-accounted drop that freed it (which binding, which plan action, which emission site) had to be guessed at, and the safe response was to narrow the fix rather than diagnose. The #715 investigation had the same need and hand-rolled a one-off ("drop-site-stamped shadow diagnostic" per the comments in compile_expr_tail.vibe).

## Proposal

In the shadow lane only (it already trades speed for diagnosis), have `rc_drop` record a small provenance stamp for the freeing site — at minimum the function index of the dropper, ideally a site counter distinguishing multiple drops within one function — into the freed block's header or a side table. When a later dup/drop touches a freed block, the trap message reports both ends: `dup of freed value at <site A>; freed at <site B>`. The funcmap already maps indices back to names.

With that, an under-provision diagnosis becomes one shadow run instead of wat archaeology, and attempts at #1980 (per-use pairing emission) get a feedback loop measured in minutes.

mizchi:
This **is** the remaining provenance channel (FREEING site at the offending dup/drop). #2199 ask 3 rides it. Not a local diagnostic; leaving open as the recorded blocker.

PR #2329, opened 2026-08-26
Stamp the freeing seq on shadow RC UAF traps (#1987)

mizchi:
## Summary

- Shadow RC (`VIBE_RC=shadow`) records a per-op sequence at each dup/drop and stores the freeing seq beside the freed-block mark.
- A later dup/drop of that block prints `dup|drop of freed value at site N; freed at site M` before trapping, instead of a bare `unreachable`.
- Reuses `gen_oob_abort_body` with a different middle fragment (`__rt_shadow_uaf_abort`). Registered only when `rc_shadow` is on (ADR-0062: normal RC stays byte-identical).

#2199 remaining source provenance for OOB rides this print-before-trap path; the crash-debug dump is already opt-in.

## Test Plan

- [x] `builtin_bodies_test.vibe` — shadow abort form strictly larger than bare unreachable
- [x] `fixtures/rc_shadow_regression_test.vibe` compiled with this stage2 under `VIBE_RC=shadow` still prints `25297489`

### DIFF
--- a/lib/@vibe/compiler/codegen/builtin_bodies/bodies_core_a1a2.vibe
+++ b/lib/@vibe/compiler/codegen/builtin_bodies/bodies_core_a1a2.vibe
@@ -36,6 +36,9 @@ import @vibe/compiler/codegen/common_base {
   emit_i64_load,
   emit_i64_store,
   emit_rc_dup_inline,
+  emit_shadow_bump_seq,
+  emit_shadow_mark_freed,
+  emit_shadow_uaf_trap,
   emit_str_ref,
   expr_tag,
   get_arm_body,
@@ -672,13 +675,13 @@ fn emit_rc_free_push(b: Bytes, ptr_local: Int, bins_base: Int) -> Unit {
 /// disagree about what a dup does, including the VIBE_RC=shadow
 /// freed-block trap, which is baked in here per module instead of at every
 /// site.
-export fn gen_rc_dup_body(rc_shadow: Bool) -> Bytes {
+export fn gen_rc_dup_body(rc_shadow: Bool, uaf_abort_idx: Int, prefix_fat: Int) -> Bytes {
   let b = bytebuf_new()
-  emit_rc_dup_inline(b, 0, rc_shadow)
+  emit_rc_dup_inline(b, 0, rc_shadow, uaf_abort_idx, prefix_fat)
   b
 }
 
-export fn gen_rc_drop_body(self_idx: Int, rc_shadow: Bool, bins_base: Int) -> Bytes {
+export fn gen_rc_drop_body(self_idx: Int, rc_shadow: Bool, bins_base: Int, uaf_abort_idx: Int, prefix_fat: Int) -> Bytes {
   let b = bytebuf_new()
   // not a heap pointer (even) → nothing to do
   emit_local_get(b, 0)
@@ -710,14 +713,18 @@ export fn gen_rc_drop_body(self_idx: Int, rc_shadow: Bool, bins_base: Int) -> By
   // shadow byte says this block is already freed. For a freed block, vptr-4
   // is the free-list NEXT link, so a drop-of-freed would silently corrupt the
   // list (decrement the link) and usually return without reaching zero --
-  // the entry check is the only reliable place to catch it.
+  // the entry check is the only reliable place to catch it. #1987: name the
+  // current seq and the freeing seq instead of a bare unreachable.
   if rc_shadow {
+    emit_shadow_bump_seq(b)
     emit_local_get(b, 1)
     emit_i32_const(b, rc_shadow_base())
     emit_i32_add(b)
     emit_i32_load8_u(b, 0, 0)
     emit_if_void(b)
-    bytebuf_push(b, 0)
+    emit_shadow_uaf_trap(b, () -> {
+      emit_local_get(b, 1)
+    }, uaf_abort_idx, prefix_fat)
     emit_end(b)
   } else {
     ()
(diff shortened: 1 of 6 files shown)


mizchi:
Worked through the review findings here. I can't push to this branch (permission-blocked on my side), so this is a patch handed to @mizchi rather than a commit — writing the substance out so it can be reviewed as text. Threaded replies were rejected by a pending review on the PR, hence one comment.

## Both stamping findings confirmed, and fixed

Confirmed by reading the emitters, not inferred from the description.

**The stamp order.** `emit_shadow_mark_freed` loads `rc_shadow_seq_addr()` at the point it is emitted. In `gen_rc_drop_body` the bump is at entry, while both marks come after every recursive `call self_idx` — so the parent is stamped with its last descendant's seq. The grown-buffer stamp has the same shape for the same reason.

Fixed by snapshotting: the entry seq goes into an i32 local right after the bump, and both mark sites store that. New `emit_shadow_mark_freed_with_seq(buf, push_vptr, push_seq)`; the live-counter `emit_shadow_mark_freed` stays for sites that free without recursing, where reading the counter is still correct.

**The reuse path.** Both `gen_rc_alloc_body` reuse sites cleared only the mark byte, leaving the previous lifetime's stamp at `vptr+4`. New `emit_shadow_clear_freed` clears the i32 together with the byte.

## One correction to the suggested fix

"meta_i32 5 → 6" is not sufficient on its own. `push_builtin(gen_rc_drop_body(...), 5, 1, 1)` is 5 i32 **plus 1 i64**, so local 6 was already the i64 local — used 13 times in this function as the recursive call's argument. Adding an i32 local shifts that i64 to 7, so those 13 uses have to be renumbered with it. Applied literally, the new i32 store would have been aimed at an i64 local.

## The test request, met only partly

Asking for a fixture that actually traps and matches the message is the right instinct, and I could not fully meet it: a use-after-free cannot be written from vibe source — correct RC accounting is the point — so there is nothing to compile that reaches the trap. That is why `rc_shadow_regression_test.vibe` is a happy path.

What is pinned instead is the mark **form**: the emitted shadow drop body must contain the snapshot-reading mark exactly once and the live-counter mark not at all. Both needles are produced by calling the real emitters, so an encoding change moves body and needle together rather than rotting the assertion. Verified red by reverting the two mark sites, and the two needles are asserted to differ so the comparison cannot pass vacuously.

Worth recording, since it is the same failure the review is about: my first version of that test counted loads of the seq address and expected 2. Measured against the built compiler the right count is 4 — entry bump, the snapshot, and the two UAF traps, which each correctly name the *current* op — so it failed against a correct compiler. I had reasoned the number out instead of measuring it.

## Not addressed

The other `#715` sites. `gen_arr_push_body`, `gen_arr_push_all_body` and `gen_arr_slice_body` still use a bare `unreachable` and the one-byte mark; threading `uaf_abort_idx`/`prefix_fat` through those emitters is a separate change and is not in this patch. That finding stands.

## Verification

- `stage2 == stage3` (FIXPOINT_OK) from a build containing the change
- `builtin_bodies_test` 12/12 on that stage2, and red when the two mark sites are reverted
- `fixtures/rc_shadow_regression_test.vibe` compiles and runs to `25297489` under `VIBE_RC=shadow`
- `ensure_generated.sh` structural probe clean; `vibe fmt --check` on every changed file
- I did **not** run the full gate lane

---

mizchi:
Follow-up to my previous comment: the third finding is now addressed too, so all three are covered by one patch (handed to @mizchi — I remain permission-blocked from pushing to this branch).

## The array paths now name the site

`gen_arr_push_body`, `gen_arr_push_all_body` (grown-buffer double-free, and the appended-element dup) and `gen_arr_slice_body` (element dup) each still emitted a bare `unreachable`, and where they freed, only the one-byte mark. All six sites now bump the seq, trap through `emit_shadow_uaf_trap`, and stamp via `emit_shadow_mark_freed` where they free. These free without recursing into children, so the live counter is still their own op — the snapshot form is only needed in `gen_rc_drop_body`.

`uaf_dup_fat` / `uaf_drop_fat` move above the array registrations, which is where they are now first needed.

Two things the change surfaced that the finding didn't mention:

- **A fourth caller.** `gen_arr_push_body` is also called by the wasm-gc backend (`backend_body.vibe`). Shadow RC is off on that lane, so it passes `-1` for the abort index, which keeps the historical bare `unreachable` there.
- The structural probe caught the arity mismatch before any build, which is worth knowing if anyone extends these signatures again.

## Regressions

Two, each verified **red against its own defect** and landing on its own message:

- the drop stamp asserts the mark **form** — the shadow drop body must contain the snapshot-reading mark exactly once and the live-counter mark not at all;
- the array paths assert the abort form is strictly larger than the bare-trap form, the same shape this file already uses for `rc_dup`/`rc_drop`. Before the change they ignored the abort index, so each pair was byte-identical.

Plus a control asserting that with shadow **off** the array bodies stay byte-identical, so none of this leaks into the production RC lane (ADR-0062).

## Verification

- `stage2 == stage3` (FIXPOINT_OK)
- `builtin_bodies_test` 13/13 on that stage2; red on each pre-fix form, checked separately
- `fixtures/rc_shadow_regression_test.vibe` → `25297489` under **both** `VIBE_RC=shadow` and `VIBE_RC=1`
- structural probe clean; `vibe fmt --check` on all 10 changed files
- the full gate lane was **not** run

The caveat from my previous comment stands unchanged: these are structural assertions, not runtime ones. A use-after-free cannot be written from vibe source, so "parent and child print different sites" is still not observed by execution, and the request for an actually-trapping fixture remains only partly met.

---

Issue #2270, opened 2026-08-24
Concrete `Iter[A]` pull type: chainable iterator surface without HKT


mizchi:
Companion to #2269. The iterator-surface question from the 0.1.0 scope thread ("should `Iterator::map` over a trait be the surface instead of `Array::map`?") does not need HKT: the shape MoonBit and Rust both ship is a **concrete** intermediate type, and vibe can do the same with stdlib work only.

## Shape

```vibe
type Iter[A]                                   // pull; one concrete constructor, no F[_]
fn Array::iter[A](xs: Array[A]) -> Iter[A]
fn Iter::map[A, B](it: Iter[A], f: (A) -> B) -> Iter[B]
fn Iter::filter[A](it: Iter[A], p: (A) -> Bool) -> Iter[A]
fn Iter::take[A](it: Iter[A], n: Int) -> Iter[A]
fn Iter::fold[A, B](it: Iter[A], init: B, f: (B, A) -> B) -> B
fn Iter::collect[A](it: Iter[A]) -> Array[A]
```

`xs |> Array::iter |> Iter::filter(p) |> Iter::map(f) |> Iter::collect` chains with pipe-first today — no type-system change, no HKT, no method-call sugar needed.

## Fit with what exists

- ADR-0099's two-layer decision stands: eager `Array::*` stays the frozen surface; `AsyncIter` is the pull layer for async sources. `Iter[A]` is the **sync pull** sibling — the role `lazy_iter_*` had before ADR-0099 folded it toward AsyncIter; whether `Iter[A]` supersedes that folding or wraps sources as `Future::ready` per ADR-0099 (1) should be decided against that ADR rather than beside it.
- `Iterable` (for-in) is unaffected.
- Element-type erasure caveats that killed the old generic combinator layer (`C: Iterable` returning Array) do not apply: every signature above is first-order generic over `A`/`B`, the same shape as `List::map` in `@vibe/core` today.

## Scheduling

Adding this after the 0.1.0 tag is a compatible Minor by the stable-surface SemVer rules, so it does not press on the freeze. Which release takes it is an owner decision (not assigned here).

Related: #2269, ADR-0099, ADR-0044.

mizchi:
Still blocked on the **concrete `Iter[A]` / HKT decision** (#2269). Chainable pull iteration without HKT is a 0.1.0 surface choice, not a local diagnostic.

mizchi:
[...]

Issue #2269, opened 2026-08-24
Higher-kinded types before 0.1.0 (ADR-0110)


mizchi:
From the 0.1.0 freeze thread (2026-08-24). Two things are settled, one is deliberately **not**:

- Settled: HKT's absence must not leak a misleading surface into the freeze — #2268 fail-closes `F[T]` where `F` is a type formal (today it parses and collapses to one opaque type), so 0.1.0 ships without vacuous constructor-application syntax.
- Settled: the cost survey below, so the scheduling decision is made on measured facts.
- **Open — owner decision, to be made before the 0.1.0 tag: which release implements HKT.** This issue does not assign a version. (One input for that decision: the roadmap's 0.2.0 "type system aimed at formalization" redesign touches the same kernel, so implementing HKT inside that redesign avoids paying the type-representation rework twice; implementing it earlier is possible at the cost of reworking it there.)

## Cost survey (measured against the implementation, 2026-08-24)

Codegen is unaffected in principle — the pipeline is uniform representation (tagged i64) with **no monomorphization** (`monoify.vibe` is not wired into any compile entry). Everything lands in the checker stack, and none of the required structure exists yet:

1. **No type application node.** `Type` is `CtNamed(String, Array[Type])` — the head is a `String`, so a variable can never stand in constructor position (`core/types.vibe:24-43`). `unify` has no head decomposition; `F := Array` is unrepresentable. `Array`/`Option` are separate *structural* constructors (`CtArray`/`CtOption`), so expressing `F := Array` also means reconciling those special cases (`type_constructor_name` / `type_args_of` are the current paper-over).
2. **No kinds.** `CtVar` carries no arity; type-parameter binders are `Array[String]` end to end (parser `parse_bounded_type_params`), and `CtForAll`'s bound metadata is `(var, Array[String])` trait names.
3. **Trait bounds are bare name strings.** `[T: Iter[Int]]` and `impl Iter[Int] for X` do not parse; impl targets drop their type args in the parser (the #2262/#1503 root).
4. **First-order parameterized traits are themselves not implemented yet.** `trait Iter[A]`'s params are parsed, threaded as shadow names, then **erased to `CtUnknown` at every dispatch** (`relax_unknown_named`, `checker/checker.vibe:392-425`, #636). Real HKT sits on top of making these real first.
[...]

Issue #2268, opened 2026-08-24
A type variable applied to arguments (`F[T]`) is accepted but has no application semantics — `F[A] -> F[B]` collapses to one opaque type


mizchi:
Found while sizing HKT for the 0.1.0 freeze (stage2 @ 6a12924, 2026-08-24). The surface accepts a type variable in constructor position, but gives it no application structure: `F` unifies with the **whole** applied type, and the bracket argument is decorative.

## Measurements

Accepted, and vacuously so — identity typechecks as a "map":

```vibe
fn fake_map[F, A, B](x: F[A], f: (A) -> B) -> F[B] {
  x                                  // accepted: F[A] and F[B] are the same opaque var
}
```

The collapse is visible in errors: with `fn pairish[F, A](x: F[A], y: F[Int]) -> Bool`, the call `pairish(["a"], [1])` is rejected with `expected Array[String], got Array[Int]` — `F[Int]` resolved to `Array[String]`, i.e. F captured the first argument's entire type and `Int` was ignored. `trait Functor[F] { fmap[A, B](F[A], (A) -> B) -> F[B] }` parses the same way; bound dispatch then fails with the witness-carrier diagnostic ("no parameter of `lift` carries the `[F: Functor]` witness ... Add a parameter of type `F` (or `Array[F]`)"), which is written for a plain type var and cannot see `x: F[A]` as carrying F.

**No value-level hole**: the caller-side result of `fake_map([1, 2], f)` is honestly `Array[Int]`, and annotating it `Array[String]` is rejected. The damage is at the promise level — a signature that reads as a transformation quietly means "returns the same type", misleading both the author and every future reader, and the use-site errors (`expected Array[String], got Array[Int]` for a mismatch the author spelled as `F[Int]`) do not mention that `F[...]`'s argument was ignored.

## Why this matters for 0.1.0

Same freeze logic as #2262: §2.2/§2.5 freeze generics and traits, and this form is today part of what the checker accepts. Giving `F[T]` real application semantics later (HKT) turns currently-accepted vacuous signatures into errors — breaking. Fail-closing it now is free.

## Proposed fix (fail-closed)

Reject a type application whose head is a **type formal in scope** (`F[T]` where `F ∈ [F, ...]`), in fn signatures, trait operation signatures, and aliases, with a diagnostic that says the semantics do not exist yet, e.g.:

[...]

PR #2331, opened 2026-08-26
Record that HKT lands before 0.1.0 (ADR-0110)

mizchi:
## Summary

Locks the owner decision #2269 left open: **higher-kinded types land before the `0.1.0` tag.**

Reason: vibe is pipeline-first (ADR-0020). The expression that convention is for is `xs |> map(f) |> filter(p)` over a constructor parameter `F`, not `xs |> Array::map(_, f)`. Today's `Array::map` is ADR-0099's eager lowering, not the 0.1.0 destination surface.

- New **ADR-0110** (`proposed`). `accepted` when `F[A]` unifies (`F := Array`), a `map[A, B](F[A], (A) -> B) -> F[B]` trait method type-checks, and at least Array and Option pipeline-chain through it.
- #2270's concrete `Iter[A]` **without HKT is not the 0.1.0 substitute**. A concrete pull type may still exist as one instance of the HKT surface.
- #2268's fail-closed `F[T]` stays until HKT is real.
- Cost survey on #2269 is unchanged: this delays 0.1.0 by a kernel project. The delay is accepted.

Docs only. No compiler change.

## Test plan

- [x] `scripts/check_version_ladder.sh` (notes / ladder / freeze version still agree)
- [x] pre-commit doc-path citations


mizchi:
Superseded by the implemented and executable final-state documentation in #2354. ADR-0110 is accepted there together with the constructor-indexed witness implementation.

Issue #2349, opened 2026-08-26
`mapfile` (bash 4) keeps six gate scripts from running on macOS stock bash — and check_gate_portability.sh does not know about it


mizchi:
**Body restated 2026-09-06 against main at `e16855e`.** The doctest third of the original report is fixed; the formatter third is not, and it reaches four more scripts than were named. See the [re-measurement comment](https://github.com/mizchi/vibe-lang/issues/2349#issuecomment-5559833593).

## Problem

`mapfile` is a bash 4 builtin. macOS ships bash 3.2, so every script below aborts before it can validate anything, and the documented local sign-off commands (`pkf run fmt`, `bash scripts/check_vibe_fmt.sh`, `pkf run release-check`) do not run on a supported contributor environment:

| script | line |
|---|---|
| `scripts/vibe_fmt_apply.sh` | 39 |
| `scripts/check_vibe_fmt.sh` | 47 |
| `scripts/cache_clean.sh` | |
| `scripts/check_freeze_surface.sh` | |
| `scripts/coverage_wasm_std.sh` | |
| `scripts/lint_method_style_naming.sh` | |

The first two are what the original report hit. The other four were found by grepping for the builtin rather than by running into them, and they mean fixing only the formatter pair leaves `release-check` broken on the same platform for a different reason. The unit of work is the six.

The compiler and the affected unit / typecheck gates themselves pass on macOS — this is a shell-portability defect in the gate layer, not a compiler one.

## ✅ Already fixed: the doctest worker

`scripts/doctest_extract_run.sh` NUL-delimits each tab-separated work-list record (`awk ... printf "%s%c", $0, 0` into `xargs -0`), so tabs survive BSD `xargs`'s `-I` normalization. No work left on that third.

## The reason it recurred, and the part that stops it

`scripts/check_gate_portability.sh` already enforces this class lexically — it rejects `rg` (including through an absolute path like `/usr/bin/rg`), `sed -i` with no suffix (GNU takes an optional one, BSD requires it, so the bare form aborts there), and `grep -E` with `\t` or `\d`. It has no rule for bash 4 builtins, which is why six uses accumulated under a gate whose whole job is this.

`AGENTS.md` states the failure mode this belongs to: *a gate that fails for a reason unrelated to what it checks gets excused rather than fixed.* A gate that cannot run at all on a supported platform is the limit case.

## Done when

[...]

PR #2352, opened 2026-08-26
Make parameterized traits semantically real

### LINKED ISSUE #2336: Make parameterized traits and bound arguments semantically real
Second implementation slice of #2269 / ADR-0110, after #2335.

## Problem

`trait Iter[A]` parameters currently parse but are relaxed to unknown at dispatch, while trait bounds and impl matching are largely bare-name based. HKT requires first-order parameterized traits to work before constructor parameters can be sound.

## Scope

- Preserve and instantiate trait declaration parameters in operation signatures.
- Parse and represent applied bounds such as `T: Iter[Int]` and applied impl trait heads.
- Match witnesses by trait arguments, not just the trait name.
- Diagnose arity, argument mismatch, missing instances, and overlaps at the bound site.
- Version every checked-state/artifact encoding changed by the richer bound representation.

## Done when

- Two instantiations of the same trait remain distinct in checking and witness selection.
- A method result/argument mentioning the trait parameter substitutes the concrete bound argument.
- Cross-module contracts preserve applied bounds.
- Existing bare first-order traits remain source compatible.
- Fixtures cover local/imported, generic/concrete, positive/missing/ambiguous cases.

Depends on #2335; blocks kinded constructor parameters in #2269.

mizchi:
Closes #2336

## Summary
- preserve applied trait arguments in bounds and impl heads across parsing, checking, imports, and cached checked state
- instantiate trait methods and witness selection from applied arguments, including generic impl matching and overlap checks
- version affected persistent TypeEnv, TDRE, module interface, and checked-state encodings
- add local and cross-module positive, missing, arity, argument-mismatch, and ambiguity coverage

## Validation
- 1,041/1,041 compiler unit tests
- 481 fixture files and 219 runtime expectations
- seed to stage3 bootstrap fixpoint
- generation and incremental trace checks
- production TDRE oracle
- late compiler gate 106/106 after fixing the detected runtime package dependency declaration
- pre-commit lint gates

The aggregate local gate reaches the known macOS Bash 3 mapfile portability failure tracked by #2349; all preceding compiler gates pass.

### DIFF
--- a/fixtures/parameterized_trait_distinct_witnesses_test.vibe
+++ b/fixtures/parameterized_trait_distinct_witnesses_test.vibe
@@ -1,0 +1,51 @@
+trait Value[A] {
+  value(Self) -> A;
+  choose(Self, A) -> A
+}
+
+struct Both {
+  number: Int;
+  text: String
+}
+
+impl Value[Int] for Both {
+  value(self: Both) -> Int {
+    self.number
+  }
+  choose(_self: Both, value: Int) -> Int {
+    value
+  }
+}
+
+impl Value[String] for Both {
+  value(self: Both) -> String {
+    self.text
+  }
+  choose(_self: Both, value: String) -> String {
+    value
+  }
+}
+
+fn int_value[T: Value[Int]](item: T) -> Int {
+  T::value(item)
+}
+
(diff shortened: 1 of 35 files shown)


mizchi:
@codex review

Issue #2387, opened 2026-08-29
Symbol interning: identifiers flow through the whole pipeline as fresh String allocations


mizchi:
Every compiler-internal identity is a `String`: env lookups, trait-dict `var_types`, the builtin registry, and monoify's mangled names (60+ chars, built by concatenating the full module path, e.g. `codegen_body_cache_new_exp_lib__vibe_compiler_codegen_common_base_common_base_vibe`). Each comparison is `__rt_str_eq` touching memory; each occurrence of an identifier is a fresh substring allocation in the lexer. The identity fast path added in #2385 only helps when two sites happen to hold the same object — today they usually don't.

## Plan (phased, spec-invariant — no user-visible change)

**Phase 1 — hash-consed identifier strings.** Intern identifier tokens at the lexer (module-level intern table, content-verified like the `pctx_cache` pattern), and intern the mangled-name constructor's results. Every occurrence of the same name becomes the same string object, so:
- the `__rt_str_eq` identity fast path fires on every successful name comparison (env hit, table verification);
- repeated identifier allocations dedup, cutting bump-heap pressure (a large share of the 1.25GB high water is name strings).

**Phase 2 — `Symbol = Int` / `(module_id, symbol_id)`.** Replace string keys in the hot tables (TypeEnv, dicts, registries) with interned ids; keep `String` only at the diagnostics/output boundary. This is what makes env misses cheap (integer compare) and removes the need for one-off tricks like the `#active_region` negative sentinel and backward `lookup_var_type` scans.

## Why blocker

The TypeEnv complete-cache redesign and the typed-lane affordability work both shrink dramatically once identity is an int. Doing them against string keys means doing them twice.

## Validation

Per the compiler-perf-profiling skill: baseline profile first, bundle regen, stage2 == stage3 fixpoint, output-wasm byte-equality on an unchanged input (interning must be a pure optimization), full unit suite, re-profile.

mizchi:
**Measured negative result for Phase 1 (PR #2393, withdrawn)** — recording it here so the next attempt starts from the data, not the plausible story.

[...]

Issue #2386, opened 2026-08-29
Design-level performance: retire the string-scan / defensive-copy architecture


mizchi:
Index for the design-level bottlenecks identified after #2385 (profiling notes there). The recurring shape across #799, #1848, and #2385 — three separate perf PRs each fixing another instance of "linear String scan + defensive `Array::slice` copy" — is structural, not incidental. This epic tracks removing the structure instead of the instances.

**Where things stand (2026-09-06):** the instance-by-instance lane is finished — about fifty measured slices over #2482–#2549 took the full-closure compile from 10.76 s / 6.48 s (cold / warm) and 1,611 MB to ~6.2 s / ~4.0 s and 820 MB on the same corpus and machine, all byte-identical, and no compiler-source function is above ~1.5% self time on either lane any more (status comments below carry the per-PR tables and the dropped attempts). What remains is structural: the warm inner loop is whole-program work re-done for unchanged modules (#2388, next, measurement first), the cold lane's environment chain walks (#2390), and the runtime intrinsics every program pays (`__rt_arr_push`, `__rt_str_eq`, `__rt_arr_new`; a codegen question, not a compiler-source one).

Sub-issues (in intended order):

1. Symbol interning — identifier identity as interned strings / symbol ids (blocker; phase 1 measured net negative in #2393, phase 2 goes with the TypeEnv redesign)
2. Per-module codegen cache for the compile lane (the check lane already has per-module reuse) — steps 1–3 landed (#2394, #2400, #2402: prefix-stable index spaces, persisted bodies behind `VIBE_CODEGEN_BODY_CACHE`); the passes before codegen are still whole-program
3. Perceus reuse analysis (in-place reuse of unique cells, FBIP)
4. TypeEnv: complete cache nodes instead of accelerator-over-chain
5. Make the typed lowering lane affordable, then delete the shape-recording machinery
6. Spec discussion: second-class borrows; module-level mutable cell

Evidence baseline (post-#2385): full-closure compile wall 8.09s, bump-heap high water 1.25GB, `env_lookup` miss band ~6% CPU, typed whole-program check on the FS lane measured 6.12s → 77.55s (#2158).

mizchi:
[...]

PR #2393, opened 2026-08-29
Intern identifier-shaped strings in the lexer and parser (#2387 Phase 1)

mizchi:
First slice of #2387 (parent epic #2386): hash-cons the identifier-shaped strings the front end produces, so equal spellings share one `String` object.

## Why this is the highest-leverage string fix

A vibe `String` is a packed non-owning `(ptr<<32)|len` view and is never RC-tracked, so two equal names almost never shared a pointer:

- every occurrence of an identifier lexed as its own zero-copy view into the source buffer — equal content, distinct pointer, so `__rt_str_eq`'s identity fast path (#2385) could not fire on the compiler's hottest comparisons (env lookup, dispatch ladders, table scans);
- every occurrence of a qualified name (`Array::push`, `Iterator::map`, ...) was TWO fresh `String::concat` allocations in the parser, per occurrence, in a bump lane that never reclaims mid-compile.

The interner (`lib/@vibe/parser/intern.vibe`) is an open-addressing table with the same fail-safe shape as the perceus `pctx_cache`: every probe re-compares actual bytes before answering, so a collision or a stale slot can only cost an extra probe, never return the wrong string. The lexer interns every `TIdent`; the parser resolves `Ns::member` through `intern_qualified`, which compares piecewise and only builds the concatenation on the first sighting of a spelling.

## Measured

Same corpus and machine as #2385: full-closure compile of `lib/@vibe/compiler/tests/codegen_lexer_test.vibe` via `scripts/profile_compile.sh`.

| metric | before | after | Δ |
|---|---:|---:|---:|
| wall | 8.10s | 5.58s | **−31%** |
| CPU total | 7.67s | 5.22s | −32% |
| bump-heap high water | 1.245GB | 727MB | **−42%** |
| linear memory pages | 19179 | 13060 | −32% |
| `env_lookup_binding` self | 448ms | out of top-20 | |
| `__rt_str_eq` self | 376ms | 210ms | −44% |
| `__rt_str_concat` self | 145ms | 72ms | −50% |
| `__rt_arr_new` self | 408ms | 165ms | −60% |

The allocator-band drops (`arr_new` / `arr_push` / `arr_get`) follow from the heap shrink: ~500MB less bump traffic means fewer `memory.grow` copies and better locality for everything downstream. Cumulative over the two perf PRs, the same compile went 12.19s → 5.58s (−54%) and 1.44GB → 727MB (−50%).

## Validation

- **Pure-optimization proof**: the same input compiled by the old and new stage2 produces **byte-identical output wasm** (interning changes which objects hold a spelling, never what the compiler emits).
- `bash scripts/generations.sh build --stage3`: **stage2 == stage3** byte-identical fixpoint.
- Full unit suite against the new stage2: 1078/1079 — the one failure was `codegen_parser_test.vibe`'s manual parser self-compile list missing the new file; fixed in this PR (same shape as #956's `float_format` row) and green on a direct re-run.
- `bash scripts/compiler_gate.sh` directly: green, all steps. (`pkf run test` currently trips over a nix-glibc `sort` mismatch in this container *before* reaching anything change-related; the same preflight steps pass when the gate runs directly.)
- `scripts/vibe_fmt.sh --check` clean on every touched file.
- New `lib/@vibe/parser/intern_test.vibe` covers round-tripping, the qualified path, and a 40k-name growth/collision sweep that re-verifies every returned string's content after rehash.

## Notes

- `#zero_alloc` is deliberately absent from `intern.vibe`: the committed seed's zero-alloc checker rejects one of the helpers, and the annotation is an assertion, not a requirement.
- New file registered in `compiler_sources_manifest.tsv` (the #740 lane guard caught the missing row — the diagnostic named the fix exactly).

https://claude.ai/code/session_0159xd5111g8KFt2CGyNyZvA

---
_Generated by [Claude Code](https://claude.ai/code/session_0159xd5111g8KFt2CGyNyZvA)_

### DIFF
--- a/lib/@vibe/compiler/compiler_sources_manifest.tsv
+++ b/lib/@vibe/compiler/compiler_sources_manifest.tsv
@@ -5,6 +5,7 @@ vibe_parser	../../../lib/@vibe/parser/index.vpkg
 vibe_parser	../../../lib/@vibe/parser/polyfill.vibe
 vibe_parser	../../../lib/@vibe/parser/token.vibe
 vibe_parser	../../../lib/@vibe/parser/float_format.vibe
+vibe_parser	../../../lib/@vibe/parser/intern.vibe
 vibe_parser	../../../lib/@vibe/parser/lexer.vibe
 vibe_parser	../../../lib/@vibe/parser/parser_base.vibe
 vibe_parser	../../../lib/@vibe/parser/parser_binder_context.vibe
--- a/lib/@vibe/compiler/tests/codegen_parser_test.vibe
+++ b/lib/@vibe/compiler/tests/codegen_parser_test.vibe
@@ -37,6 +37,9 @@ test "wasi: parser.vibe import - compile and run" {
     // parser-private __to_string polyfill — include both.
     "lib/@vibe/parser/polyfill.vibe",
     "lib/@vibe/parser/float_format.vibe",
+    // #2387: the lexer interns identifiers, so its intern table must ride
+    // along in this manual self-compile (same shape as float_format above).
+    "lib/@vibe/parser/intern.vibe",
     "lib/@vibe/parser/lexer.vibe",
     "lib/@vibe/parser/parser_base.vibe",
     "lib/@vibe/parser/parser_binder_context.vibe",
--- a/lib/@vibe/parser/index.vpkg
+++ b/lib/@vibe/parser/index.vpkg
@@ -127,6 +127,14 @@ fn parse_effect_stmt_with_binder_context(tokens: Array[Token], pos: Int, exporte
 fn parse_type_alias_stmt_with_binder_context(tokens: Array[Token], pos: Int, exported: Bool, context: Option[ParserBinderContext]) -> (Stmt, Int) with Exception
 fn parse_struct_stmt_with_binder_context(tokens: Array[Token], pos: Int, exported: Bool, context: Option[ParserBinderContext]) -> (Stmt, Int) with Exception
 
+// #2387: hash-consing for identifier-shaped strings. The lexer interns every
+// TIdent and the parser interns qualified `Ns::member` names, so equal
+// spellings share one String object (str_eq identity fast path) and the
+// per-occurrence qualified-name concat is gone.
+fn intern_slice(s: String, start: Int, end: Int) -> String
+fn intern_str(s: String) -> String
+fn intern_qualified(ns: String, member: String) -> String
+
 // --- parser_expr (+ primary/core/dispatch stages)
 fn parse(tokens: Array[Token]) -> Expr with Exception
 fn parse_ok(tokens: Array[Token]) -> Bool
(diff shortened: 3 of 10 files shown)


mizchi:
Withdrawing this PR: **the headline numbers in the description are a measurement error**, and a fair re-measurement shows the change is a small regression, not a win.

## What went wrong in the measurement

The persistent typing cache is **content-keyed, not compiler-keyed**, so my "after" profile run was served by cache entries written during the "before" run. The comparison was cold-baseline vs warm-candidate. Re-measured with the cache controlled (`VIBE_BUILD_CACHE_DIR` pointed at a fresh directory per run, 3 runs each, full-closure compile of `codegen_lexer_test.vibe`):

| lane | baseline (main) | this PR | qualified-only variant |
|---|---:|---:|---:|
| cold wall | 7.99–8.18s | **8.32–8.43s (+3.5%)** | 7.95–8.46s (neutral) |
| cold heap_ptr | 1.4577 GiB | 1.4569 GiB | 1.4569 GiB (−0.9 MiB) |
| warm wall | 4.72s | 4.78s | — |

This is consistent with what CI's deterministic perf report already said (selfcompile heap +0.07%, parse B/op −0.5..−1.3%, everything else ±0): the per-occurrence qualified-name concat I blamed for the heap was actually ~1 MiB, and hashing every identifier at lex time costs more than identity-canonicalization buys in this lane.

The byte-identical-output and fixpoint results were real — the change is semantically sound — but a sound change that costs 3.5% cold and wins nothing measurable should not merge.

## What survives

- The negative result and the corrected methodology are recorded on #2387 — the win there has to come from *replacing* string comparisons (Phase 2's interned ids as keys), not from adding a hashing layer on top of them.
- A follow-up PR adds the cache-controlled measurement protocol to the profiling skill so the next perf change doesn't fall into the same trap.

---

mizchi:
Two corrections to the withdrawal comment above, for the record — this PR merged at 12:49Z, fifteen minutes **before** the withdrawal was posted, so flagging them clearly:

1. **This change is in main and measures as a ~+3.5% cold-wall regression on the FS full-closure lane** (cache-controlled, N=3 per side; warm is ~neutral, CI's deterministic rows are ±0 with parse B/op −0.5..−1.3%). The description's −31%/−42% claims were a measurement error. It is semantically sound (byte-identical output, fixpoint), so nothing is broken — but if the cost isn't wanted, the clean revert is dropping the `intern_slice` calls in `lexer.vibe` (the hashing per identifier occurrence is the cost center) while keeping or dropping `intern_qualified` (measured neutral). Happy to prepare that revert if you want it; leaving the call to you since the merge may have been a deliberate architecture bet for #2387 Phase 2.

2. **The withdrawal's mechanism explanation was wrong**, as Codex pointed out on #2394: the cache namespace embeds `codegen_fingerprint()` (compiler sources), so a cross-compiler cache hit is impossible. The real contamination was same-compiler warming with asymmetric preparation — the candidate's own stage3 self-compile and the output-equivalence check of the profiling corpus pre-warmed its namespace, while the baseline's build ran without `--stage3`. The corrected diagnosis and protocol are in `.claude/skills/compiler-perf-profiling` §0.5 (PR #2394). The +3.5% number itself is unaffected — it came from the cache-isolated re-measurement.

---

PR #2415, opened 2026-08-30
Prune unreachable functions and renumber (#2359)

### LINKED ISSUE #2359: Builtins are linked unconditionally: 230 of the 245 bytes `fib` has grown are code it never calls
`core/dce.vibe` tree-shakes top-level **user** statements. It does not tree-shake **builtins** — every registry builtin's body, func-table entry and type entry is part of every linked module's fixed footprint, whether or not the program can reach it.

`bench/perf/size_baseline.txt` has been recording the consequence for a year of rebaselines without anyone naming it as its own problem. Reading the entries' own attributions for the smallest sample, `bench/binary_size/fib.vibe`:

| rebaseline | fib | Δ | what the entry attributes it to |
| :--- | ---: | ---: | :--- |
| initial (#1109-4) | 755 | — | |
| #1191 | 780 | **+25** | 5 new `Map` runtime builtins; "none of these five samples touch Map" |
| #1666 | 795 | +15 | main's drift, recorded but not attributed |
| #2199 | 881 | **+86** | OOB abort message strings; "none of them indexes out of range" |
| #2199 ask 2 | 980 | **+99** | `__rt_print_i64_raw` + `__rt_oob_abort`; same shape |
| #2344 | 1000 | **+20** | `Int::popcount` / `ctz` / `clz`; none of them counts a bit |

**+245 B (+32%) total, of which 230 B is attributed to builtins the sample never calls.** Every one of those four entries repeats the same sentence — *"there's no usage-based tree-shaking for builtins today, only for top-level user statements (`core/dce.vibe`)"* — which is a strong signal that the note has become the finding.

The ratchet is doing its job; this issue is not a complaint about it. The point is that each individual rebaseline was correct and proportionate, and the accumulation is nonetheless now the dominant term in a small program's size.

## Why it matters beyond tidiness

`docs/wasm/feature-levels.md` and ADR-0043's `--allow-*` work both aim at *"capabilities fixed at build time drive progressive code generation for the target platform's wasm runtime"* — dropping code for capabilities that were not granted, via const-fold plus DCE. Builtin DCE is the same mechanism one layer down, and without it the floor under every generated module keeps rising no matter what the capability work removes.

It also changes the calculus for adding builtins. Right now, every new one taxes every program forever, so "is this worth 20 bytes for everyone?" is a real question that gets asked at every rebaseline. With DCE it stops being a question.

## Shape of the work

Builtins already have exactly the metadata reachability needs — `core/builtin_registry.vibe` is the single source of truth for names, and both lanes resolve bodies by index from it. The hard parts are the ones to scope first:

- **Builtins that call builtins.** Bodies reference each other by index (`gen_bytes_concat_body(hbo + 18, hbo + 22)`, `gen_string_contains_body(false, hbo + 30)`), so reachability has to close over those edges, not just over user code.
- **Index stability.** Both lanes' index chains are positional and, as #2344 found the hard way, stale-by-construction in two places (linear `max_builtin_func_idx`, gc `num_generated`). Pruning makes indices depend on the program. Assigning them after the reachable set is known is likely cleaner than trying to keep a fixed numbering with holes.
- **The lane that contracts not to prune.** `compile_wasi_module_no_dce_impl` deliberately exempts itself (`enable_late_dce = false`); `size_baseline.txt`'s #1666 note records a gate that was written and then removed for demanding pruning from exactly that lane. Any new gate here must test the lane that ships.

## Definition of done

Per the admission policy in #2340: a committed before/after on `bench/binary_size/*` with the ratchet **ratcheted DOWN** so the gate protects the win, per `size_baseline.txt`'s own header (*"either direction — ratchet DOWN after a size win"*). If pruning turns out not to pay for its complexity, the measurement is the answer and this closes as rejected.

Found while rebaselining for #2344.

mizchi:
Fixes #2359.

| sample | before | after | Δ |
|---|---:|---:|---:|
| fib | 1026 | 648 | **−36.8%** |
| hello_world | 1062 | 684 | **−35.6%** |
| closure_indirect | 2701 | 1887 | −30.1% |
| variant_float | 4530 | 3309 | −27.0% |
| fizzbuzz | 1615 | 1247 | −22.8% |
| stage2.wasm | 2800980 | 2797370 | −0.1% |

## What was actually costing the bytes

The issue's framing — "every builtin's body is part of every module" — is not what happens, and I corrected that [in the issue](https://github.com/mizchi/vibe-lang/issues/2359#issuecomment-5466591809) before implementing. The existing pass already shrinks an unreachable helper's **body** to a 3-byte `unreachable` stub. What it cannot remove is the **index slot**, because every call site references indices fixed far earlier. Measured exactly: 75 of `fib`'s 79 functions are unreachable, 378 B = 36.8% of the file.

So the whole task is the one thing the issue listed as a sub-problem: renumbering.

## Why the existing scan cannot do it, and what replaces it

The current pass finds call targets by scanning body bytes for `call`(0x10) / `ref.func`(0xd2) and decoding whatever follows as a LEB. Its safety argument is explicit in the source: a byte that merely *looks* like an opcode — one sitting inside another instruction's immediate — can only ever **keep** a helper that is already dead. That is harmless when the answer is "stub or not". Rewrite that same byte and the module is corrupt.

This pass decodes instructions properly instead of scanning for them. Every immediate is skipped by its own encoding, so a call opcode is recognized only where an instruction actually starts, and each immediate's offset is known rather than guessed.

**The decode carries its own proof.** A body is accepted only when the walk lands *exactly* on its final byte. A mis-decoded immediate desynchronizes the walk, which then overshoots or stops short — so "my opcode table is incomplete" surfaces as a refusal, never as a wrong rewrite. This is not theoretical: I prototyped the decoder in Python first and it hit exactly this twice (`throw`, `try_table`), failing loudly both times instead of silently mis-parsing.

Anything not understood — unknown opcode, desync, a `ref.func` in a global initializer, a start section, an unexpected element-segment encoding, or a kept body calling something pruned — abandons pruning for the **whole module** and returns the input unchanged. The cost of being wrong about the encoding is the optimization, not the program.

## A second finding: the old scan leaves real dead code

Because it over-approximates, the byte-scan does not merely fail to remove slots — it leaves whole live bodies that nothing can reach:

| sample | dead fns | stubbed by the old pass | **missed** |
|---|---:|---:|---:|
| variant_float | 67 | 58 | **9 = 909 B** |
| closure_indirect | 72 | 67 | **5 = 467 B** |
| fib / hello_world / fizzbuzz | 73–75 | 72–74 | 1 = 6 B |

That is why `variant_float` and `closure_indirect` gain more than their stub-slot count alone predicts.

## The lane that contracts not to prune

Gated to the same builds as the stubbing pass (instrumented builds must see unmodified bodies), **plus `enable_late_dce`** — so `compile_wasi_module_no_dce_impl` keeps its contract rather than being made to prune by a different route. `size_baseline.txt`'s #1666 note records a gate that was written and then removed for demanding pruning from exactly that lane.

## Verification

- **Full `compiler_gate.sh` green** (exit 0, 124/124 sections) including the stage2==stage3 fixpoint. That is the strongest test available here: the compiler compiles *itself* through this lane, so the pass ran over a 5,860-function module, and the resulting stage2 then compiled everything else.
- All five samples produce **identical output** through `run_wasm_vibe_host_runner.sh`.
- Sizes match an **independently written Python prototype to the byte** on all five. Agreement on one number is luck; five exact byte counts from two separate implementations is not.

## Baseline ratcheted DOWN, from measured numbers

Per `size_baseline.txt`'s own header. The "before" column is the **currently measured** size, not the committed rows (fib 1020, hello_world 1056, fizzbuzz 1609, closure_indirect 2681, variant_float 4513). Those were stale: #2403 moved every sample +6…+20 B without rebaselining, allowed because each stayed inside +2%. Ratcheting from them would have quietly credited this change with #2403's drift.

**Red/green on the ratchet itself**, because a gate that has not been shown to fail is not protecting anything:

```
pre-change compiler vs the new baseline  -> FAIL on all five (+29% .. +58%)
this compiler        vs the new baseline -> ok, 0 B on every sample
```

That is also what catches a silent refusal: every bail-out path returns the unpruned module, so the sizes jump straight back over tolerance rather than quietly reverting to the old behavior.

## Expected in the perf report

Every linear-lane scenario loses its dead function slots, so the nine execution scenarios will move substantially too. Those are this change; the small pre-existing drift in the committed baseline rows was #2403's, and the entry in `size_baseline.txt` says so rather than letting the two be conflated again.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FYMVd6D2o82suTfTwoJPMf)_

### DIFF
--- a/bench/perf/size_baseline.txt
+++ b/bench/perf/size_baseline.txt
@@ -174,8 +174,36 @@
 # The wasm-gc lane is the opposite and is NOT gated here: it links builtins
 # unconditionally, so these four cost it +2.2..3.1% on every perf scenario
 # including programs that never search bytes. That is #2359, not this file.
-closure_indirect 2681
-fib 1020
-fizzbuzz 1609
-hello_world 1056
-variant_float 4513
+
+# 2026-08-30 rebaseline (#2359, whole-module function pruning): the linear
+# lane now DROPS unreachable functions and renumbers, instead of shrinking
+# their bodies to a 3-byte `unreachable` stub and keeping the index slot.
+# closure_indirect 2701->1887, fib 1026->648, fizzbuzz 1615->1247,
+# hello_world 1062->684, variant_float 4530->3309. Ratcheted DOWN per this
+# file's header so the gate protects the win.
+#
+# Note the "before" column is the CURRENT measured size, not the row that was
+# committed below (fib 1020, hello_world 1056, fizzbuzz 1609,
+# closure_indirect 2681, variant_float 4513). Those were stale: #2403 (`~`)
+# moved every sample +6..+20 B without rebaselining, which the gate allowed
+# because each stayed inside +2%. Ratcheting from the stale numbers would
+# have quietly credited this change with #2403's drift.
+#
+# The tax this removes is the one the NOTE above predicted. That note tracked
+# fib 755 -> 1000 across five rebaselines and attributed 230 of the 245 bytes
+# to builtins linked unconditionally that the sample never calls. Measured
+# exactly, it was larger than the note estimated: 75 of fib's 79 functions are
+# unreachable, 378 B = 36.8% of the file. Four of the five samples were more
+# than a fifth index slots for functions that can never run.
+#
+# Two of those bytes-per-sample figures are NOT stub slots. The older
+# body-stubbing pass finds call targets by scanning body bytes for 0x10/0xd2,
+# which over-approximates: measured, it leaves 9 dead functions with real
+# bodies in variant_float (909 B) and 5 in closure_indirect (467 B). The new
+# pass decodes instructions properly, so it finds those too.
+
+closure_indirect 1887
+fib 648
+fizzbuzz 1247
+hello_world 684
+variant_float 3309
--- a/lib/@vibe/compiler/codegen/wasi/linked_compile.vibe
+++ b/lib/@vibe/compiler/codegen/wasi/linked_compile.vibe
@@ -3718,6 +3718,712 @@ fn lc_derive_helper_names(stmts: Array[Stmt]) -> Array[String] {
   names
 }
 
+//# ---------------------------------------------------------------------------
+//# #2359: whole-module function pruning (exact decode + renumber)
+//#
+//# The older pass below (search "Binary-size: replace the bodies") shrinks an
+//# unreachable generated helper's BODY to `unreachable` but must leave its
+//# function index in place, because every call site references indices fixed
+//# far earlier. Measured on `bench/binary_size`, that leftover index block is
+//# the dominant term in a small module: 74 of `fib`'s 79 functions are
+//# byte-identical 3-byte stubs, 371 B = 36% of the file.
+//#
+//# Removing the slots means renumbering, and renumbering is exactly what that
+//# pass's safety argument does not support. It finds call targets by scanning
+//# body bytes for `call`(0x10) / `ref.func`(0xd2) and decoding whatever
+//# follows as a LEB. A byte that merely LOOKS like an opcode -- one inside
+//# another instruction's immediate -- can only ever KEEP a helper that is
+//# actually dead, which is harmless when the answer is "stub or not". Rewrite
+//# the same byte and the module is corrupt.
+//#
+//# So this pass decodes instructions properly instead of scanning for them.
+//# Every immediate is skipped by its own encoding, so a call opcode is only
+//# recognized where an instruction actually starts, and the byte offset of
+//# each call immediate is known exactly rather than guessed.
+//#
+//# The decode carries its own proof: a body is only accepted when the walk
+//# lands EXACTLY on its final byte. A mis-decoded immediate desynchronizes the
+//# walk, which then overshoots or stops short -- so "my opcode table is
+//# incomplete" surfaces as a refusal, never as a wrong rewrite. Any unknown
+//# opcode, any desync, any construct below that is not handled (a `ref.func`
+//# in a global initializer, an unexpected element-segment encoding) abandons
+//# pruning for the WHOLE module and returns the input unchanged. The cost of
+//# being wrong about the encoding is the optimization, not the program.
+//#
+//# Gated to the same builds as the older pass, plus `enable_late_dce`: the
+//# `compile_wasi_module_no_dce_impl` lane contracts not to prune, and this
+//# honors that rather than making it prune by a different route.
+//# ---------------------------------------------------------------------------
+
+/// Read a LEB128 at `p`, returning the decoded value; `cell[0]` receives the
+/// position just past it. Unsigned and signed LEBs share a terminator (the
+/// first byte with bit 7 clear), so the SAME walk skips both -- only call and
+/// ref.func immediates are read for their value, and those are unsigned.
+fn lc_pf_leb(b: Bytes, p: Int, cell: Array[Int]) -> Int {
+  let mut q = p
+  let mut v = 0
+  let mut shift = 0
+  let mut more = true
+  let len = Bytes::length(b)
+  while more && q < len {
+    let byte = Bytes::get(b, q)
+    if shift < 35 {
+      v = v|((byte&127)<<shift)
+    } else {
+      ()
+    }
+    q = q + 1
+    if (byte&128) == 0 {
+      more = false
+    } else {
+      shift = shift + 7
+    }
+  }
+  Array::set(cell, 0, q)
+  v
+}
+
+/// Skip a LEB128 at `p`, returning the position just past it.
+fn lc_pf_skip_leb(b: Bytes, p: Int) -> Int {
+  let mut q = p
+  let mut more = true
+  let len = Bytes::length(b)
+  while more && q < len {
+    let byte = Bytes::get(b, q)
+    q = q + 1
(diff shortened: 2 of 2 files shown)


mizchi:
## The `selfcompile heap_ptr_bytes +3.21%` in the perf report is real, and it is this PR's

Flagging it rather than letting it pass as noise, and reporting a fix attempt that **failed**, because the failure is the useful part.

### Reproduced

`scripts/selfcompile_kpi.sh`, same input (`codegen_lexer_test.vibe`), stage2 built from each side:

| compiler | heap_ptr_bytes | vs baseline |
|---|---:|---:|
| pre-change | 1,496,595,432 | — |
| this PR | 1,536,044,400 | **+2.64%** |

(CI measures a different input and reports +3.21%; same effect.)

### Where it comes from

Not the pruning. The pass re-emits the whole module, so every **unchanged** section — types, imports, globals, the string-table data section — is copied out and back. On the compiler's own 2.8 MB module that is the dominant cost, and it is paid on every compile regardless of how much is pruned.

Note the shape of the trade there specifically: on `stage2.wasm` itself the pass saves 3,610 B (−0.13%), because the compiler actually uses nearly all of its functions. Small modules are where the 20–39% lives.

### The obvious fix makes it worse

`lc_pf_copy_range` pushes one byte at a time, so replacing it with `Bytes::slice` + `Bytes::append` looks like a free win. Measured:

| copy strategy | heap_ptr_bytes | vs baseline |
|---|---:|---:|
| byte-at-a-time (this PR) | 1,536,044,400 | +2.64% |
| bulk `Bytes::slice` + append | 1,557,379,376 | **+4.06%** |

`Bytes::slice` allocates a fresh buffer per range, and on the bump allocator that intermediate is never freed — it counts straight into the high-water mark. Copying "efficiently" adds 21 MB of garbage on top. Reverted; the branch is unchanged from the green commit.

I am not landing a speculative second attempt on a PR that is already green. Two directions look real if the regression is judged unacceptable:

1. **Skip the re-emit when the saving is negligible.** The dead fraction is known after reachability and before re-emission — 95% of functions on the samples (75/79 on `fib`), a rounding error on the compiler's own module. A threshold there keeps every win and skips exactly the pathological case. It is a tuning knob, which is why I have not added it unilaterally.
2. **Accept the trade**: +2.6–3.2% compile-time heap for −20…−39% on every emitted program, with fuel, B/op and gc output all unchanged.

My read is that (2) is defensible on its own and (1) is worth doing anyway, but that is a call about this compiler's priorities rather than a fact about the diff, so it is yours. Happy to push (1) if you want it in this PR.

---

mizchi:
Closing the open question from https://github.com/mizchi/vibe-lang/pull/2415#issuecomment-5467732347: I measured option (1) and it does not work. Option (2) — accepting the compile-heap trade — is the right call.

The flaw was in my own reasoning, not the implementation. I generalized from "the compiler's own module is 0.4% dead" to "the selfcompile KPI is the pathological case", but the KPI compiles `codegen_lexer_test.vibe`, whose module is **2.3% dead**. A 2% threshold is a no-op on it (byte-identical output *and* heap) while still costing 10,374 B on the compiler's shipped artifact; a 5% threshold engages and buys 1.92% of transient compile heap for 2.79% of output size. Both sides scale with module size, so that ratio holds at every dead fraction — skipping is only free when there is nothing to drop at all, which the pass already tests exactly.

Full numbers and method in #2417, which lands the measurement as a comment at the tempting spot. No behavior change.

---

mizchi:
Correction to my comment above, since it changes the answer: I wrote "both sides scale with module size, so that ratio holds at every dead fraction". That is wrong, and review caught it.

The re-emit's cost tracks module size, but the saving tracks `dead_fraction × module_size` — so the exchange rate is *not* constant, and skipping gets relatively cheaper as the dead fraction falls. My own data shows it: the compiler's own module at 0.4% dead costs 10,374 output bytes to skip, a twentieth of the KPI's 104,420.

So the accurate answer is narrower than what I gave. At the one point measured (2.3% dead), skipping buys 29,848,504 bytes of transient compile heap for 104,420 bytes of output — a bad trade for a pass whose purpose is output size, and unnecessary while the heap gate has 4.2% of slack. **A threshold is unmeasured at low dead fractions, not refuted.** What is dead-fraction-independent is only the direction: skipping never *saves* output bytes, so "nothing to drop at all" is the only free case.

Option (2) still stands as the call for now, but on "not worth taking at the measured point" rather than the general claim I made. Fixed in #2417 (`eddc0bec1`).

---

mizchi:
Two figures in my comments above are wrong; both are fixed in #2417 (`f771196e4`), and both happen to have argued *against* my own recommendation, so correcting them strengthens it.

- **"4.2% of slack"** was the distance below the committed baseline, not the gate's capacity. `ci.yml:248` sets the ceiling to baseline + 10% (1,781,726,918), so from 1,551,740,640 the gate allows **14.8% of growth**. More headroom means less pressure to buy heap with output bytes.
- **"a twentieth"** — 104,420 / 10,374 = 10.07, so a tenth. The compiler module is cheaper to skip than the KPI, but by 10×, not 20×.

Everything else I quoted here re-derives correctly from its source. Recommendation is unchanged: accept the trade for now, on "not worth taking at the measured point" rather than any general claim.

---

PR #2465, opened 2026-09-02
Run the perf report inside ci.yml on compiler-gate's artifacts instead of rebuilding them

mizchi:
## What

The `perf-metrics` job (the sticky "📊 Perf report" PR comment, and the `bench-data` snapshot on main pushes) moves from its own workflow, `perf.yml`, into `ci.yml` as an advisory job that `needs: [compiler-gate]` and downloads that job's `compiler-stage2` artifact plus a new `compiler-generated` artifact (the five generated files and the stamp) instead of regenerating and rebuilding them itself. `perf.yml` is deleted.

## Why (measured)

On every compiler-source push the job's caches miss, and it rebuilt what `compiler-gate` builds in the same run. PR #2456, run [33632520247](https://github.com/mizchi/vibe-lang/actions/runs/33632520247):

| step | time |
|---|---:|
| Ensure generated compiler artifacts | 3 min 34 s |
| Build stage2 | 2 min 36 s |
| Build viberun (rust-cache hit) | 1 min 26 s |
| Collect metrics | 47 s |
| total | 8 min 45 s |

The metrics themselves are under a minute; 7 of the 8.75 minutes were a second copy of another job's build.

## How

- `compiler-gate` uploads `compiler-generated` (the artifacts `ensure_generated.sh` produces, `include-hidden-files` for the stamp) next to the existing `compiler-stage2` upload.
- `perf-metrics` downloads both, then runs `ensure_generated.sh --check` so a download that does not match this checkout's (seed, compiler source) fingerprint fails instead of silently measuring another commit's bundles. The seed cache stays (the runner-calibration series is compiled by the committed seed). The wasmtime install goes: it served only `generations.sh`'s per-stage validation, which no longer runs in this job.
- Everything after "Collect metrics" (baseline fetch, report, comment upsert, main-only snapshot append, artifact upload) is unchanged; `permissions` moved to the job.
- The job is not in `ci-required`, so waiting on `compiler-gate` does not touch the required path's t=0 layout (#1849 / #2001); it holds the same position `wasi-p3-gate` already does.
- Docs and comments that pointed at `perf.yml` (`bench/perf/README.md`, `bench/exec/README.md`, `bench/perf/tracked_benches.txt`, `scripts/bench_metrics.sh`) now point at the job.

Expected: about 3 minutes for the job (viberun build + metrics) and one fewer stage2 build per push. This PR's own run is the first measurement; I will post the numbers here.

## Not changed here

The same rebuild happens in every other job of the run: 16 jobs each regenerate the bundles (~3.5 min) and build a stage2 (~2.5 min) on a compiler-source push, about 100 runner-minutes per push. Those jobs are on the required path, whose t=0 layout is a recorded decision in `ci.yml`'s header, so trading runner-minutes for a shared build stage there is a separate decision.

https://claude.ai/code/session_01QKZdvBvHVUagwg6Yih8er8

---
_Generated by [Claude Code](https://claude.ai/code/session_01QKZdvBvHVUagwg6Yih8er8)_


mizchi:
First run on this PR's head, [33640571430](https://github.com/mizchi/vibe-lang/actions/runs/33640571430): the `perf-metrics` job ran after `compiler-gate`, downloaded both artifacts, passed `ensure_generated.sh --check`, and posted the perf report above.

| perf-metrics step | before (perf.yml, run 33632520247) | after (this run) |
|---|---:|---:|
| Ensure generated compiler artifacts | 3 min 34 s | — (downloaded) |
| Build stage2 | 2 min 36 s | — (downloaded) |
| Build viberun | 1 min 26 s | 1 min 22 s |
| Collect metrics | 47 s | 48 s |
| job total | 8 min 45 s | 2 min 30 s |

The `compiler-generated` upload added to `compiler-gate` did not register as a step over 5 s. Caveat on the comparison: the "before" run was a compiler-source push (every cache missed), this run touches no compiler source, so its other jobs also hit the caches; the perf job's own figure is not affected by that, since its build steps no longer exist. Its first cache-miss measurement will come from the next compiler-source PR after this merges.

---

Issue #2231, opened 2026-08-22
A call whose callee roots at another call shares the inner call's offset, so its Double result misses float_call_offsets


mizchi:
Split out of the #2226 review (which asked for the codegen half; that half is done).

## What happens

`#2158` gives linear codegen the checker's per-call-site `Double` answer through
`CompileCtx.float_call_offsets`, keyed by the **call node's source offset**. A
call whose callee is not an identifier never gets into that table, so its
`Double` result reaches `__to_string` as raw f64 bits and prints a
plausible-looking integer.

Measured 2026-08-22 on a stage2 from the #2226 branch, single-source lane
(the only lane the offsets are wired into):

| callee shape | printed |
|---|---|
| `f()` — identifier | `2.5` |
| `(f)()` — parenthesised identifier (same AST) | `2.5` |
| `g(1.25)` — alias binding | `2.5` |
| `(mkh().f)()` — **field** callee | **`348`** |
| `(() -> { 2.5 })()` — **immediately-invoked lambda** | **`372`** |

```vibe
fn main with Console {
  println("iife = \{(() -> { 2.5 })()}")   // iife = 372
}
```

## Where it goes wrong

Not in the classifier. `cc_expr_is_floatish` / `expr_is_floatish` now consult
the offset table **before** dispatching on the callee shape (that was the
#2226 review finding, fixed) — but for these shapes the table has no entry to
find.

`check_program_double_call_result_observation_impl` only records an occurrence
when `off >= 0`:

```vibe
if role == "CallResult" && lane == "Primary" {
  primary_calls = primary_calls + 1
  if off >= 0 {
    ...
```

and `tab_call_ty` is gated the same way. So the question is upstream: the
located parser does not give these call nodes an offset (or does not record
them as a primary `CallResult`). The two failing rows differ from the three
passing ones only in the callee's AST shape, and `(f)()` passing while
`(mkh().f)()` fails shows parentheses are not the variable — being an `EIdent`
is.

## Why P1 rather than P0

It is silently wrong output, which is P0 shaped. It is filed P1 because it is
**not a regression**: before #2158 every one of these five rows printed raw
bits, and that change fixed three of them without touching the other two. The
wrong answer is old; only the contrast is new.

## What "fixed" looks like

The located parser assigns an offset to a call node regardless of its callee's
shape, so the existing recording and lookup work unchanged. No codegen change
should be needed — that side is already offset-keyed and callee-agnostic.

[...]

PR #2478, opened 2026-09-03
parser: stop anchoring the throw desugar's two call nodes at one offset (#2427)

### LINKED ISSUE #2427: CLI-core build traps in __rt_str_concat with a corrupt (ptr|len) — NOT a float-call-offset misclassification (measured)
Found while driving PR #2425: adding an ordinary `Int`-returning function to `lib/@vibe/compiler/runtime/typecheck_fs.vibe` and rendering its result with `__to_string(...)` inside a `String::concat` chain made the **selfhost CLI-core build** (`tests/gates/bootstrap/stage2_oracles.sh` → `scripts/test_cli_core.sh`) trap deterministically with `RuntimeError: memory access out of bounds` inside `__rt_str_concat` — the stage2 binary's own compiled body hands concat a corrupt `(ptr|len)`.

## Evidence (all measured, probes on PR #2425 around commit 3ffd6f4)

- **Not memory exhaustion**: a runner instrumented to dump stats at the trap shows `heap_ptr = 986,667,896` (~0.94GB) with 4.26GB pre-grown; raising pre-grow to the wasm32 ceiling changes nothing.
- **Not the values**: shrinking the accumulator mask from `(1<<60)-1` to `(1<<40)-1` (eliminating all wrapping) still traps.
- **The call site, not the loop**: the identical checksum loop invoked as a standalone `let _cs = module_float_offsets_checksum(offsets)` statement passes; the identical call inlined into the innermost argument of a nested `String::concat` header build traps; hoisting into locals still traps once the result flows into `__to_string`.
- **Guard instrumentation pinpointed the corrupt operand**: with length guards on every operand, the build fails legibly with `float-offsets diag: checksum_text corrupt` — i.e. `__to_string(checksum_value)` returned a non-digit garbage string for a value that had just passed a `>= 0` check.
- **A standalone micro-repro of the exact same shapes, compiled by the same trapping stage2, passes** — the corruption is position-dependent, not shape-dependent.
- **KB-scale source diffs flip it**: stage2 built from one commit completes the CLI-core build at a 3.65GB peak; stage2 from the next (a ~KB diff in one file) traps, cold or warm cache, idle machine.

## Reading

The affected binary is stage2, whose body is compiled by stage1 from the **minified flat bundle** through the single-source lane — the lane that feeds `CompileCtx.float_call_offsets` from its whole-program check (#2158/#2226). Everything above is consistent with a **float-offset misattribution keyed by call-site offset in the bundle's offset space**: an unlucky call site gets classified as a Double call result, codegen treats the Int result as floatish, and the first consumer that renders it as a string gets garbage. Which call site loses the lottery shifts with any edit that moves bundle offsets — which is exactly why the same code passes as a statement and fails inlined, and why unrelated KB diffs toggle it.

`unique_call_offsets_in_stmts` was supposed to make collisions degrade instead of misclassify; the #2231 offset-anchoring family (callee-start vs `(`-fallback) is the known way one node's recorded offset can coincide with a different node's `get_ecall_off`. This instance says a hole of that class is still reachable on the bundle lane.

## Impact

P0-class (silently wrong values are possible before it escalates to a trap), currently rare — it needs an unlucky offset coincidence in the bundle. PR #2425 sidestepped it by not adding new integer-rendering call sites (v3 end-marker envelope instead of a rendered checksum, commit b02211b) and left the probe trail in that PR's comments.

## Suggested next steps

1. Reproduce the misattribution directly: run `check_program_double_call_result_located_observation` over the current `compiler_sources_bundle.vibe` and diff the recorded offsets against the actual `get_ecall_off` of every call node — any offset that matches a node the checker did NOT record as its own Double call is a smoking gun.
2. Audit the offset-anchoring rules (#2231) for a remaining shape whose recorded offset can equal a different node's key.
3. Consider making the channel fail closed harder on the bundle lane: drop any offset that is not BOTH recorded and matched to the same node identity, not just numerically unique.

3 axes: **type** = bug / **priority** = P0 (silently wrong) / not a `blocker`.

mizchi:
#2427's step 2 asked for an audit of the #2231 offset-anchoring rules "for a remaining shape whose recorded offset can equal a different node's key". This is the answer, measured over the tree instead of argued from examples, plus the fix and a scan that keeps answering.

## The finding

Exactly one shape in the whole corpus gives two call nodes one `get_ecall_off` key:

```
throw(x)  ->  ECall(EIdent("perform", off), [ECall(EIdent("Exception::Throw", off), [x], off)], off)
```

`parse_throw_primary` anchored both call nodes at the `throw` keyword. That is the hazard `call_site_offset` (#2231) exists to remove — this desugar bypasses it by building `ECall` nodes directly rather than going through `parse_postfix`.

Scanned across every `ECall` offset and every `EDot` field-name offset — both halves of the Double table's key space:

| corpus | keys | colliding offsets |
|---|---:|---:|
| `lib/**`, 997 parseable files | 161,423 | 904 |
| `_cli_adapter_module_source.vibe` (the reprinted flat source the single-source lane compiles) | 80,747 | 782 |

Every one of them this shape. No `EDot` field-name key on a call key, no second desugar, and the obvious candidate — a curried `f()()` — anchors its two calls four bytes apart.

**It has never misclassified anything.** Measured on `fn f(x: Int) -> Double with Exception { if x > 0 { 1.5 } else { throw("no") } }`, the checker records no occurrence at the `throw` offset at all: the `perform` arm of `check_expr` returns without routing either result through `tab_call_ty`. An offset nobody records is an offset nobody looks up. That is an accident of a distant arm, not a property of the shape — `perform`'s result in a `Double`-returning function IS a `Double`, and the day someone types it for hover the offset becomes a claim about both nodes.

## The fix

The inner op call gets no offset. `-1` is the honest key rather than a fallback: the source has no `Exception::Throw(...)` token to point at, and an unlocated call is never recorded and never looked up. The `perform` node and the op's `EIdent` keep the keyword offset, so `op_off` in the checker's perform arm and every located diagnostic are unmoved.

## The scan

`lib/@vibe/compiler/tests/offset_key_collision_test.vibe` walks the whole tree on every run (4.6s) and asserts zero collisions, reporting the SHAPES rather than a list of offsets — a shape is what a reader has to act on, and offsets repeat one finding per site.

Two things it does on purpose:

- **It is a scan, not a set of cases.** What this guards against is a shape nobody thought to write a case for; the `throw` desugar went 904 sites without one.
- **It pins the corpus size** (`paths > 900`, `keys > 100000`). A scan that reads nothing reports "clean", and that failure mode is indistinguishable from a pass. Both floors sit well under today's numbers — they exist to catch a walk that silently returns little, not to be bumped as the tree grows.

It also sizes its duplicate-finder by NODE count, not file length: the flat module source is 7.4MB, and a counting array indexed by byte offset would have been the reason the scan could not run on the one input that matters most.

Red-tested — reverting the parser hunk gives:

```
lib/@vibe/cli/dispatch.vibe: 59 colliding offset(s) @6172{perform/1 Exception::Throw/1 } ...
```

## Correction

The comment on `check_program_double_call_result_observation_impl` (#2476) said "no colliding pair was found in today's bundle while writing this". That measurement was on the wrong file: `compiler_sources_bundle.vibe` holds compiler sources as **string constants**, so parsing it sees a few dozen call nodes and answers "clean" about code it never looked at. The comment now records what the scan found.

## What this does not close

#2427 stays open. The only key collision on the bundle lane is inert and there is no other, so the issue's "call-site-offset lottery" reading is a hypothesis the corpus does not support — whatever handed `__rt_str_concat` a corrupt `(ptr|len)` in #2425's build is still unidentified. Step 3 (node identity in the key) is also untouched, and is still the real fix for an offset-keyed channel.

## Verification

Local stage2 build plus gate lanes `early` / `mid` / `late` and the full unit suite are running; `vibe fmt --check` is clean on all three changed files, and the parser, symbol-span and range-agreement suites (142 tests) pass. I will report the gate results on this PR and push any fix they turn up.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FYMVd6D2o82suTfTwoJPMf)_

### DIFF
--- a/lib/@vibe/compiler/checker/checker_stmt.vibe
+++ b/lib/@vibe/compiler/checker/checker_stmt.vibe
@@ -6689,9 +6689,25 @@ export fn typed_lowering_enc_tag(enc: Int) -> Int {
 /// Double claimant unclassified. Node identity in the key is the fix, and an
 /// offset-keyed consumer cannot express it -- #2427's remaining step.
 ///
-/// No colliding pair was found in today's bundle while writing this, and a
-/// curried `f()()` -- the obvious candidate -- anchors its two calls four
-/// bytes apart. The rule closes the hole whether or not an instance is live.
+/// A curried `f()()` -- the obvious candidate -- anchors its two calls four
+/// bytes apart, and this comment used to stop there, at "no colliding pair was
+/// found in today's bundle". That was measured on the wrong file. The bundle
+/// holds compiler sources as STRING CONSTANTS, so parsing it sees a few dozen
+/// call nodes and answers "clean" about code it never looked at; the flat
+/// source the single-source lane actually compiles is the reprinted
+/// `_cli_adapter_module_source.vibe`.
+///
+/// Scanned properly (998 files, every `ECall` offset and every `EDot`
+/// field-name offset -- both halves of this table's key space), exactly ONE
+/// shape collided: the `throw(x)` desugar, which anchored `perform` and the op
+/// call at the same `throw` keyword. 904 sites in `lib/**`, 782 more in the
+/// flat module source, and nothing else -- no dot key on a call key, no second
+/// shape. It never misclassified anything, because the `perform` arm returns
+/// without routing either result through `tab_call_ty`, so neither node
+/// records an occurrence. `parse_throw_primary` now leaves the op call
+/// unlocated (#2427), and `tests/offset_key_collision_test.vibe` scans for the
+/// next one. The rule here still closes the hole whether or not an instance is
+/// live -- a desugar is one edit away from making one.
 fn check_program_double_call_result_observation_impl(stmts: Array[Stmt]) -> Option[(CheckedProgram, Array[Int], Int, Int, Array[Int], Array[String])] with Exception {
   let capture = reset_double_call_capture()
   let checked = check_program_type_defs_observation_impl(stmts, None, Some(capture)).checked_program
--- a/lib/@vibe/compiler/tests/offset_key_collision_test.vibe
+++ b/lib/@vibe/compiler/tests/offset_key_collision_test.vibe
@@ -1,0 +1,291 @@
+//# #2427: no two call nodes may share one `get_ecall_off` key.
+//
+// The Double typed-lowering table (#2158 tag 0, #2424 tag 4) is keyed by
+// OFFSET, so an offset in it classifies EVERY node anchored there. #2476 made
+// the extraction poison an offset whose claimants disagree, which keeps a
+// collision from being silently wrong. This file is the other half: the
+// invariant that says a collision should not exist in the first place.
+//
+// `call_site_offset` (#2231) establishes it for every call the expression
+// parser builds. The hazard is a DESUGAR that constructs `ECall` nodes
+// directly and picks their offsets by hand -- it bypasses that function
+// entirely. `throw(x)` was such a desugar: it anchored `perform`, the op call,
+// and both callee idents at the `throw` keyword, so two call nodes answered
+// `get_ecall_off` with one number.
+//
+// Measured over the tree at the time of writing: 998 files, 904 colliding
+// sites in `lib/**` and 782 more in the reprinted flat module source
+// (`_cli_adapter_module_source.vibe`, the shape the single-source lane
+// compiles) -- every one of them this desugar, and no other shape anywhere,
+// including zero `EDot` field-name keys colliding with a call key.
+//
+// The scan below is the durable form of that measurement. It is a scan, not an
+// example: the failure mode this guards against is a shape nobody thought to
+// write a case for.
+
+//# Imports
+
+import @vibe/compiler/syntax {
+  lex_with_offsets, parse_program_located
+}
+
+import @vibe/compiler/core {
+  Expr, Stmt, expr_children
+}
+
+//# Walk
+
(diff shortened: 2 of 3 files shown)


mizchi:
Verification, and one note on the perf report above.

**Local, against a stage2 built from this branch** (`_build/gen2427b/stage2.wasm`; both stage1 and stage2 self-validation passed inside the build):

| check | result |
|---|---|
| `compiler_gate.sh` lane `early` | ok |
| `compiler_gate.sh` lane `mid` | ok |
| `compiler_gate.sh` lane `late` | ok |
| `unit_test_runner.sh` | 1110/1110 |
| `vibe fmt --check`, all three files | clean |

CI is green on `2879155` (28 checks, `ci-required` success), and Codex finished with no findings.

### The perf report's baseline is one merge behind this branch

It compares against `4f9cdf099` (the #2476 merge) while this branch is based on `50801e0a9` (the #2471 merge), so **#2471's diff — 12 files, +579/−64 — is inside every delta it shows**. Reading +0.16% selfcompile heap as this PR's would be wrong.

This PR's own contribution to those numbers is 2,516 bytes of **comments** in two sources the bundle embeds verbatim (`parser_expr_primary.vibe` +1,420, `checker_stmt.vibe` +1,096; both are in `compiler_sources_manifest.tsv`), which is roughly half of the +0.08% bundle growth. The code change is one integer literal. Every execution row is ±0 and output checks are 9/9, so there is no behavioural drift from either side of the comparison.

---

Issue #2464, opened 2026-09-02
gc lane: `__to_string` of a Bool-returning BUILTIN call renders `0`/`1` — a regression against the seed


mizchi:
## Symptom

`__to_string(eq(a, b))` renders **`0`** on the wasm-gc lane and **`false`** on the linear lane. Same file, same compiler, no diagnostic.

```vibe
test "eq render" {
  let a = 1
  let b = 2
  assert(__to_string(eq(a, b)) == "false")
}
```

| compiler | linear | gc |
|---|---|---|
| committed seed | `false` ✅ | `false` ✅ |
| stage2 @ `e37c4da8e` (main) | `false` ✅ | **`0`** ❌ |

**This is a regression on main**, not a pre-existing gap: the committed seed answers correctly on both lanes. Found while converting a fixture's expectations to `inspect` in #2463 — `inspect(v, content = ..)` desugars to `__to_string(v)`, so the conversion turned the gc lane red and exposed it.

## Narrowing

Measured on a stage2 built from `e37c4da8e`, one shape per file so the first trap does not mask the rest:

| producer | linear | gc |
|---|---|---|
| a direct comparison `a < b` | `true` | `true` |
| a named `fn` returning Bool | `true` | `true` |
| a `let`-bound Bool | `true` | `true` |
| **the `eq` builtin's result** | `false` | **`0`** |
| the same spelled `__to_string(eq(a, b))` | `false` | **`0`** |

So it is specifically **the result of a Bool-returning builtin CALL**. Every other Bool shape renders correctly on gc.

## Cause

`gc_expr_is_boolish` (`lib/@vibe/compiler/codegen/gc/backend_expr.vibe:465`) has **no `ECall` arm at all**:

```vibe
fn gc_expr_is_boolish(e: Expr, local_names: Array[String], ctx: CompileCtxGc) -> Bool {
  match e {
    EBool(_) => true,
    EIdent(fname, _) => { ... ctx.gc_bool_locals ... },
    EBinOp(op, _, _, _) => op == "==" || op == "!=" || ... ,
    EUnaryOp(op, _) => op == "!",
    _ => false
  }
}
```

Its linear twin `cc_expr_is_boolish` (`codegen/expr/compile_call.vibe:1032`) has a tag-8 arm covering `__arr_equals__*`, `T::equals`, `not(..)`, the ascription marker, and `cc_builtin_call_ret_kind(e) == 1` — which lists **17** Bool-returning builtins:

[...]

Issue #2462, opened 2026-09-02
`__to_string` of a Bool renders `1` when the producer is a lambda bound inside a function body


mizchi:
## Symptom

`__to_string(f())` renders `1` instead of `true` when `f` is a lambda bound by a
`let` **inside a function body**. The same lambda bound at top level renders
`true`, and so does a named `fn`. There is no diagnostic: the program compiles,
runs, and prints the wrong text.

Found while pinning #2423; unrelated to it and not a regression.

```vibe
let top_lam: () -> Bool = () -> Bool {
  1 < 2
}

fn main() -> Unit with Stdout {
  let inner_lam = () -> Bool {
    1 < 2
  }
  println(String::concat("top-level lambda : ", __to_string(top_lam())))
  println(String::concat("inner lambda     : ", __to_string(inner_lam())))
}
```

```
top-level lambda : true
inner lambda     : 1
```

## Narrowing

Measured on the committed seed and on a stage2 built from `e37c4da8e`, on the
linear/RC lane and the gc lane — identical on all four. Not a regression; the
shape has never worked.

| producer | rendered |
|---|---|
| named `fn` returning a Bool literal | `true` |
| named `fn` returning a comparison | `true` |
| named `fn` with a large non-foldable body | `true` |
| `let` at TOP LEVEL bound to a `-> Bool` lambda | `true` |
| `let` INSIDE a function bound to a `-> Bool` lambda | **`1`** |
| the same, lambda body is a comparison | **`1`** |
| the same, lambda captures a Bool | **`1`** |
| the same, result bound to an unannotated `let` first | **`1`** |
| result bound to an **annotated** `let x: Bool = f()` | `true` |
| a plain `let b = true` | `true` |
| a bare comparison `1 < 2` | `true` |

So the defect is exactly: **the call result of a lambda whose binding is not a
top-level statement**. An explicit `Bool` annotation on the consuming binding is
a working workaround (it goes through the ascription-marker arm).

## Cause

Two independent classifiers answer "is this argument a Bool", and neither sees
this shape.

1. `interp_arg_is_bool` (`lib/@vibe/compiler/normalize/desugar_trait_dict.vibe:9660`)
   → `infer_arg_type_name`, which for an `ECall(EIdent(name), ..)` looks `name`
   up in **`fn_returns`**. `collect_fn_returns`
   (`desugar_trait_dict.vibe:7400`) walks the **statement** list and records
[...]

PR #2480, opened 2026-09-03
codegen: say why the Int/Bool render tables must NOT be filtered (#2427)

mizchi:
Continuing #2427 past the falsified offset-collision hypothesis (#2478). This is a **negative result plus a tripwire** — no behaviour change, a corrected comment and a new test.

## The suspect

`linked_compile` narrows the Double table with `unique_typed_lowering_offsets_in_stmts`, deliberately *after* `effect_lowering_prelude` has rewritten `stmts`, and installs the Int (tag 3) and Bool (tag 5) tables raw:

```vibe
let float_call_offsets = unique_typed_lowering_offsets_in_stmts(stmts, checker_float_call_offsets)
let int_render_offsets = checker_int_render_offsets
let bool_render_offsets = checker_bool_render_offsets
```

That asymmetry looks like an oversight, and it is exactly the shape #2427 describes: an offset-keyed classification reaching codegen without the property it depends on being established. So I measured it, and the first result looked like a confirmed P0 — routing the Int table through the same filter dropped rows on ordinary code (`int 6->4`, `int 1->0` on two of the first fixtures scanned).

## What it actually is

The rows were not being dropped because a node was duplicated. They were dropped because **the filter and these tables disagree about what a key is**.

`count_typed_lowering_offsets_expr` counts `ECall` (own offset) and `EDot` (field-name offset), and deliberately not `EIdent` — a call's own offset IS its callee identifier's offset, so counting both would score every ordinary `f(x)` as 2 and drop every call-result row. But tags 3 and 5 key an argument by `typed_lowering_arg_offset`, which **does** answer for `EIdent`. Such a row scores 0, not 1, and the filter drops it.

Measured, one row each through the three key kinds:

| source | key kind | int before → after |
|---|---|---|
| `inspect(v, ..)` | name | 1 → **0** — dropped |
| `inspect(n(), ..)` | call | 1 → 1 |
| `inspect(s.a, ..)` | projection | 1 → 1 |

Rendering a bare NAME through `inspect` is the repo's preferred assertion idiom, so the symmetric-looking hardening would have deleted most of the channel at its most common site and put back the syntactic classifier tags 3 and 5 exist to replace (#2462, #2464). The Double table is immune because `typed_lowering_float_key` answers `-1` for `EIdent` and never files a name-keyed row in the first place.

## What changes

**The comment.** It credited the merge's per-file rebase for keeping the offset spaces disjoint. That is the wrong reason twice over: `unique_typed_lowering_offsets_in_stmts`'s own doc says disjointness must be **established** rather than assumed from how some entry parsed, and the rebase is not what protects these tables. What does is the checker's agreement rule — an offset is filed only when every typed occurrence there resolves to the wanted type, so a call offset shared with its callee's `CtFn` occurrence is dropped on disagreement.

**A tripwire test.** `tests/typed_lowering_filter_key_space_test.vibe` pins all four measurements and says what to do if it fails: if the counter learned about `EIdent`, check that a name-keyed row scores 1 rather than 2 via its enclosing call, then delete the file and filter all three tables. It is deliberately a tripwire, not a specification of desirable behaviour.

Each case also asserts the *before* count, so none of them can pass vacuously — if the checker ever stopped filing the row, the test fails rather than quietly agreeing that 0 → 0.

Red-tested by adding an `EIdent` arm to the counter — the name-keyed case flips `0 -> 1` and the test fails naming the row.

## On #2427

Still open, and this narrows it further rather than closing it. The merge rebase is sound, the alias-collision clone already copies the provider's whole tag-encoded table into a fresh range, and the raw Int/Bool install is a necessity rather than a gap. Together with #2478's corpus scan, the offset-keyed channel is now accounted for on both lanes, and the `__rt_str_concat` corruption from #2425 remains unexplained.

The related improvement — a counter that scores a name-keyed row 1 by skipping a call's own callee ident, which would let all three tables be filtered — is deliberately **not** in this PR. It is a behaviour change to a hot codegen walk, and nothing measured here shows the current install is wrong.

## Verification

All 28 CI checks green on `38abcda`; perf `+0.00%` selfcompile heap with every deterministic and execution row `±0`. `vibe fmt --check` clean on both files.

The test's helper measures the real lane's order, which I checked rather than assumed: `check_program` desugars `inspect` itself (`checker_stmt.vibe:5314`, and `:6974` on the projected-env path) before checking, so the sequence really is desugar → check → prelude → filter, and the prelude's own `desugar_inspect_calls` is a no-op by then.

Note for reviewers: `float_call_offset_fs_lane_test.vibe` fails under `vibe_test.sh`'s default seed compiler and passes against a real stage2 — pre-existing on a clean tree, the documented "which compiler answered" caveat, not this change. Codex's review failed twice on this head with `An unknown error occurred` (a review-service error, not findings), so this PR has no bot review coverage.

### DIFF
--- a/lib/@vibe/compiler/codegen/wasi/linked_compile.vibe
+++ b/lib/@vibe/compiler/codegen/wasi/linked_compile.vibe
@@ -4670,11 +4670,32 @@ export fn compile_wasi_module_linked_impl_with_typed_offsets(stmts: Array[Stmt],
   // rather than against whatever the entry handed the checker. Empty in, empty
   // out, without a walk.
   let float_call_offsets = unique_typed_lowering_offsets_in_stmts(stmts, checker_float_call_offsets)
-  // #2424: the Int render-argument table needs no uniqueness filter of its
-  // own. The checker already refuses an offset whose typed occurrences
-  // disagree, and #2429's per-file rebase (`+ base * 8` on the encoded value)
-  // keeps the merged offset spaces disjoint, so what arrives here is already
-  // the fail-closed set.
+  // #2424: the Int (tag 3) and Bool (tag 5) render-argument tables are
+  // installed RAW, and the filter above must not be extended to them.
+  //
+  // What makes the raw install honest is the checker's agreement rule: an
+  // offset is filed only when EVERY typed occurrence there resolves to the
+  // wanted type, so a call offset shared with the `CtFn` occurrence of its
+  // callee is dropped on disagreement.
+  //
+  // The reason NOT to filter is stronger than "it is unnecessary" -- the
+  // filter and these tables disagree about what a key is, so applying it
+  // deletes rows (#2427). `count_typed_lowering_offsets_expr` counts `ECall`
+  // and `EDot` and deliberately not `EIdent`, since a call's own offset IS its
+  // callee identifier's offset. But tags 3 and 5 key an argument by
+  // `typed_lowering_arg_offset`, which DOES answer for `EIdent`, and a bare
+  // name is the commonest render argument there is. Measured, one row each:
+  //
+  //   inspect(v, ..)    name        1 -> 0   dropped
+  //   inspect(n(), ..)  call        1 -> 1   kept
+  //   inspect(s.a, ..)  projection  1 -> 1   kept
+  //
+  // `inspect(<name>, ..)` is the repo's preferred assertion idiom, so the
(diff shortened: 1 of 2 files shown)


mizchi:
@codex review

(The review on `38abcda` reported **Failed** about a minute after starting — a review-infrastructure failure rather than findings, so there is no coverage on this head yet. Re-triggering once.)

---

Issue #2391, opened 2026-08-29
Make the typed lowering lane affordable, then delete the shape-recording machinery


mizchi:
A family of mechanisms exists only because lowering sometimes does not know the type: the structural `==` shape-recording chain (ADR-0097 / #1526, `dtd_elem_projecting_shape` and friends), the `String <` address-compare class (#2128), traps on untyped empty-literal equality, and per-instantiation comparator emission. Each is a separately maintained levee against silent-wrong, and each carries permanent complexity plus a P0 risk surface.

The real fix — full type information available to codegen — was built in #2158 and measured at **6.12s → 77.55s** on the FS lane, so it ships only where a whole-program check already runs. That 12x is not intrinsic: it is the cost of string-keyed env lookups (interning sub-issue), accelerator-over-chain misses (TypeEnv sub-issue), and re-checking the whole closure per compile (per-module cache sub-issue + #1379/#1959/#1960).

**This issue is the payoff tracker:** once those land, re-measure #2158's lane. When the typed check fits the FS-lane budget, make it the default, then delete the shape machinery and the address-compare fallbacks — retiring the whole silent-wrong workaround class rather than patching its instances. That is perf work buying correctness, which is the most on-policy trade this codebase can make ("never be silently wrong" ranks above everything else).

Acceptance shape: typed lane on by default in `vibe test`/`vibe run` with an agreed overhead bound (proposal: ≤1.3x wall vs today's untyped lane on the full-closure corpus), then removal PRs for the shape recorder and its fixtures-turned-obsolete.

mizchi:
Premise re-measured on current main, and the first slice is up as PR #2425.

**The whole-program re-check is still unaffordable, but less so than recorded.** Full-compiler closure (`codegen_lexer_test`), interleaved runs, isolated caches: adding `check_program_double_call_result_located_observation(merged_stmts)` to the FS/RC lane costs **5.8s → 32.6s warm (~5.6x)** and **10.4s → 39.7s cold (~3.8x)** — down from the recorded 12.7x (the #2385 sentinel and the head-query fix did help), but nowhere near the ≤1.3x budget. Two additional disqualifiers surfaced: the recursive whole-program checker **exhausts the default host stack** on a compiler-sized merged program (needs the 128MB stack the bootstrap compiles use), and the check **mutates the merged AST** enough to change the emitted wasm.

[...]

Issue #2388, opened 2026-08-29
Per-module codegen cache: the compile lane's cache granularity is whole-closure while the check lane is per-module


mizchi:
The check lane has per-module typing reuse (TDRE9 transport in `runtime/typecheck_fs.vibe`, validated; #1379 / #1959 / #1960 track persisting it further). The compile lane does not: the persistent wasm artifact cache (`entry/compiler/file_compile/file_compile.vibe`) is keyed on the **whole-closure** source-group fingerprint, so a one-file edit is all-or-nothing — a full ~8s recompile of the 1000+ file closure.

The building blocks already exist:

- `CodegenBodyCache` (`codegen/common_base/common_base.vibe`, used by `linked_compile.vibe` for in-run parallel slicing) stores per-function bodies plus relocation metadata (`fn_meta_i32/i64/v128`, lambda captures, table slots). That is effectively a relocatable object format — it is just never persisted.
- The `.vpkg` bodyless contract is a stable module interface; `vibe deps --direct` gives the invalidation graph at header-cache granularity.
- The stage2 == stage3 fixpoint discipline means the compiler is deterministic, so content-addressed caching is unusually safe here.

## Proposed key

`(module implementation fingerprint, dep interface fingerprints, capability/allow set, backend, codegen flags)`. The whole-program steps that remain at link time — monoify instantiation set, structural-eq comparator emission, capability const-fold + DCE, function index assignment — stay whole-program but operate on cached bodies; per-instantiation caching keyed by `(generic def fp, type args)` can come later.

## Expected upside

- One-file edit inner loop: recompile the changed module + replay + link, plausibly 8s → ~1s warm.
- The unit suite (1079 files) almost all imports the same compiler closure; a shared per-module cache compiles those modules once per suite instead of per entry. This is the largest CI cost today.

## Safety rule

Cache invalidation bugs here are silent-wrong (P0 class). Follow the established pattern: content-verify or validated transport (TDRE9, `pctx_cache`), and fall back to a fresh compile on any mismatch — a stale entry may only cause a miss, never a wrong body.

mizchi:
**Measured design study: what actually blocks per-module codegen reuse** (first implementation slice in PR #2394).

## Correcting this issue's premise

[...]

PR #2578, opened 2026-09-07
Fix three gates that were red on main, and wire them into CI (#2577)

### LINKED ISSUE #2577: check_task_inputs.sh is red on main, and it is wired into no CI job
Found while running the meta-gates against a new gate (#2562 / #2576). It is not caused by that change — verified by stashing and re-running on a clean `origin/main`.

## The failure

```console
$ bash scripts/check_task_inputs.sh
check-task-inputs: FAIL: task `bench-compile-hotspots` runs scripts/profile_compile.sh, which reaches the compiler,
  but it does not include a complete compiler input set and does not set `cache = false`. A compiler identity
  change would replay a cached pass.
  Add `...compilerProbeInputs` to the list -- spelling inputs out should ADD to the
  defaults, not replace them.
```

`bench-compile-hotspots` entered the tree in #2473. The gate has been red since.

## Why it stayed red

`check_task_inputs.sh` is a pkfire task and **appears in no `.github/workflows/*.yml`**, so nothing has been running it. Measured:

```console
$ grep -rn "check_task_inputs" .github/workflows/*.yml Taskfile.pkl
Taskfile.pkl:612:  cmd = "bash scripts/check_task_inputs.sh"
Taskfile.pkl:619:  scriptTask("test-check-task-inputs", "scripts/check_task_inputs_test.sh")
```

Same shape as #2538 (a test red on main and wired into no gate) and as the case `AGENTS.md` already records — *"`release-check` has no CI job, which is how `lint-tracked-experiment-names` went unrun for weeks."* Three instances of one pattern now: **a gate that exists, is correct, and runs nowhere.**

## Why the finding matters more than the fix

The fix itself is one line — add `...compilerProbeInputs` to the task, or `cache = false`. What the gate is protecting against is the rule `AGENTS.md` states at length: *a gate that asks the compiler a question has to be told WHICH compiler*, and a task keyed on the wrong inputs replays a cached green about a compiler that does not contain the change. `bench-compile-hotspots` profiles a real compile, so a stale cached pass there is a measurement of the wrong binary — the exact failure `.claude/skills/compiler-perf-profiling` and the #2393 cache-control rule exist to prevent.

## Done when

- `bench-compile-hotspots` carries `...compilerProbeInputs` (preferred — it adds to the defaults) or `cache = false`, and `check_task_inputs.sh` is green.
- **`check-task-inputs` and its self-test run in CI**, in the `structural-lint` job beside the other structural gates. Without that half, the next instance is invisible again.
- While there: audit whether any other pkfire-only gate has the same exposure. `check_task_inputs.sh`, `check_gate_portability.sh`, `check_gate_self_tests.sh` and `check_doc_commands.sh` are all reachable from `pkf` — check which of them a CI job actually invokes, and wire the ones that do not.

## Severity

P2 by the stated symptom: nothing that works today is broken, and the gate is correct — it is the wiring that is missing. It earns attention anyway because an unwired gate is indistinguishable from a passing one, which is the property `AGENTS.md` says to stop tolerating.

mizchi:
Closes #2577. **Stacked on #2576** — that PR already edits `Taskfile.pkl` and `ci.yml`, and `docs/issue-triage.md`'s own rule says not to start a branch on a file an open PR occupies. Base retargets to `main` automatically when #2576 merges.

#2577 reported **one** unwired red gate. Working its third done-when item — *"audit whether any other pkfire-only gate has the same exposure"* — found **two more**, and one of them is the gate `AGENTS.md` already names as the cautionary example.

## 1. `check_task_inputs.sh`

`bench-compile-hotspots` reached the compiler with `inputs { ...vibeSources; ...scriptSources }`: no `bootstrap/seed.json`, no `cache = false`.

Fixed with **`cache = false`**, not the `...compilerProbeInputs` the gate's message suggests. No input list can key this task correctly: the stage2 to profile arrives as an **argument**, so two runs against different compilers share a cache key and the second replays the first's numbers — the "which compiler answered" trap, in the one task whose entire output is a measurement. Its three sibling measurement tasks that take arguments (`kpi-selfcompile`, `measure-fs-heap`, `kpi-edit-cycle`) all use `cache = false`; this was the outlier, and the gate accepts either.

## 2. `check_portable_boundary.sh` — newly found

It asserted a **two-generations-old** shape of the boundary:

```
export let compile_source: (String) -> Bytes with { Error }   # in entry/source_compile/index.vibe
```

ADR-0070 moved the contract to `index.vpkg`, and ADR-0085 replaced `Error` with `Exception`. So the gate reported *"missing expected pure boundary"* for **files that do not exist**, while every boundary it names was intact in the contract beside them:

```
fn compile_source(source: String) -> Bytes with Exception     # in entry/source_compile/index.vpkg
```

Repointing it revealed a second defect the first had been hiding: `forbid_pattern` matched `compile_file_fs_mode` inside a **doc comment** (`-- see \`compile_file_fs_mode\`'s comment, #2391`). These files exist to explain how they differ from the FS lane, so prose naming a native entry point is the writing we want. Comment **lines** are now stripped before matching — whole lines only, never a trailing comment after code, since a stripped tail could hide a real call sharing the line.

Repairing a gate that has been dark is exactly when *"it says ok now"* is equally consistent with **"it was fixed"** and **"it was neutered"**, so it gets the self-test it never had:

| case | expected |
|---|---|
| control, unmutated tree | passes (runs first — three cases assert failure, so a gate that rejected everything would make them vacuously pass) |
| `perform Fs::read_file` in **code** | rejected |
| the same names in a **comment** | accepted — this is what keeps the comment strip from being a blanket exemption |
| a boundary declaring `with Fs` | rejected |

Removed from `gate_self_test_allowlist.txt`; that ratchet allows removals.

## 3. `lint_tracked_experiment_names.sh` — newly found

`scripts/prefix_stability_probe.py` is a tracked probe with no allowlist entry — real, referenced by nothing, and measuring exactly the question #2388 is open about. Allowlisted as `manual-experiment` with that reason rather than deleted.

**This is the gate `AGENTS.md` cites**: *"`release-check` has no CI job, which is how `lint-tracked-experiment-names` went unrun for weeks."* It is still unrun, and it has been red.

## The wiring, which is the actual fix

All three, plus `check-doc-commands`, now run in CI's `structural-lint` job, each with its self-test where one exists. Three instances of one pattern — a gate that exists, is correct, and runs nowhere — is not three accidents.

## The audit, reported rather than acted on

**19 of 40 gate scripts have no path from any workflow** (traced transitively through `pkf run` tasks named in workflows, their aggregate members, `gates_shard.sh`, `compiler_gate.sh` and the four `tests/gates/*/run.sh` lanes). `release-check` appears in no workflow at all, which is the common cause.

Of the nine cheap ones I could measure directly, six were green, three were red — the three fixed here. The rest need a stage2 and cannot live in `structural-lint`; wiring those is a scheduling decision about which CI job pays for a compiler build, so it is left on #2577 with the data rather than guessed at here.

One methodological note, since it bit me mid-audit: my first pass classified the unwired gates as "needs a compiler" by grepping each script for `stage2|runtime/vibe|ensure_seed`. That matched `check_task_inputs.sh` and `check_doc_commands.sh`, both of which merely *talk about* the compiler and run in about a second. Measuring beat grepping — the same proxy-versus-property substitution these gates exist to stop.

## Verification

Every touched gate and self-test, run directly:

- gates: `check_gate_portability`, `check_gate_self_tests`, `check_task_inputs`, `check_portable_boundary`, `lint_tracked_experiment_names`, `check_doc_commands`, `check_doc_classification` — all green
- self-tests: `check_portable_boundary_test`, `check_doc_classification_test`, `check_task_inputs_test`, `lint_tracked_experiment_names_test` — all green
- `check_gate_self_tests.sh` still passes with one fewer exemption; `gate_self_test_failing.txt` stays empty
- `pkf run pre-commit` — passed

https://claude.ai/code/session_0194y71aVEMcwoem5oAxFJPX

---
_Generated by [Claude Code](https://claude.ai/code/session_0194y71aVEMcwoem5oAxFJPX)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -734,6 +734,20 @@ jobs:
         run: bash scripts/check_doc_classification_test.sh
       - name: doc classification lint
         run: bash scripts/check_doc_classification.sh
+      - name: task-inputs gate self-test
+        run: bash scripts/check_task_inputs_test.sh
+      - name: task-inputs gate
+        run: bash scripts/check_task_inputs.sh
+      - name: doc-commands gate
+        run: bash scripts/check_doc_commands.sh
+      - name: portable-boundary gate self-test
+        run: bash scripts/check_portable_boundary_test.sh
+      - name: portable-boundary gate
+        run: bash scripts/check_portable_boundary.sh
+      - name: tracked-experiment-names lint self-test
+        run: bash scripts/lint_tracked_experiment_names_test.sh
+      - name: tracked-experiment-names lint
+        run: bash scripts/lint_tracked_experiment_names.sh
       # The #1189 / ADR-0081 naming ratchet, reaching CI for the first time. It
       # is a release-check dep, but release-check has no CI job either, so its
       # one enforcement point was a local sign-off nobody re-runs on a green
--- a/Taskfile.pkl
+++ b/Taskfile.pkl
@@ -440,6 +440,13 @@ local testLintTrackedExperimentNames =
 // same artifacts purely to diff them against the tracked copies.)
 local checkSelfhostPortableBoundary =
   scriptTask("check-portable-boundary", "scripts/check_portable_boundary.sh")
+// #2577: the gate had rotted to an `index.vibe` / `with { Error }` shape two
+// ADRs old and reported a missing boundary that was intact in the contract
+// beside it. The self-test pins that it can still fail -- and that a doc
+// comment naming a native entry point is not a leak, which is what the repair
+// had to distinguish.
+local testSelfhostPortableBoundary =
+  scriptTask("test-portable-boundary", "scripts/check_portable_boundary_test.sh")
 local checkHostRuntimeContract = new Task {
   name = "check-host-runtime-contract"
   description = "fail-closed conformance check for the documented vibe.* core host ABI"
@@ -1248,7 +1255,16 @@ local benchSelfhostCompileHotspots = new Task {
   name = "bench-compile-hotspots"
   cmd = "bash scripts/profile_compile.sh $@"
   acceptsArgs = true
-  inputs { ...vibeSources; ...scriptSources }
+  // `cache = false`, like every other measurement task that takes an argument
+  // (kpi-selfcompile, measure-fs-heap, kpi-edit-cycle). No input list can key
+  // this one correctly: the stage2 to profile arrives as an ARGUMENT, so two
+  // runs against different compilers would share a cache key and the second
+  // would replay the first's numbers -- the "which compiler answered" trap
+  // (AGENTS.md), in the one task whose entire output is a measurement. It ran
+  // with `inputs { ...vibeSources; ...scriptSources }`, which also omitted
+  // bootstrap/seed.json, and check_task_inputs.sh has been red on main since
+  // (#2577).
+  cache = false
 }
 
 local selfcompileKpi = new Task {
@@ -1874,6 +1890,7 @@ tasks {
   checkTreesitterWasmCorpus
   testCheckTreesitterWasmCorpus
   checkSelfhostPortableBoundary
+  testSelfhostPortableBoundary
   checkHostRuntimeContract
   checkVibeFmt
   checkTaskfileFmt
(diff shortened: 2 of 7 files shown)


mizchi:
This PR's commits never reached `main` — it merged into `claude/doc-classification-gate-8eytq3` *after* that branch had already landed via #2576, so its head sits on a branch nothing merges from. All of its content reached `main` through [#2579](https://github.com/mizchi/vibe-lang/pull/2579), merged at `cd9eb4f`. Nothing here needs redoing; this note exists so the orphaned history is not mistaken for missing work.

---

Issue #2109, opened 2026-08-18
A handled body may not call a row-carrying function, so `handle { println(..) }` does not compile


mizchi:
## Symptom

Inside a `handle ... with E { .. }`, the handled body may not call a function that carries an effect row — any row, including one that has nothing to do with `E`.

```vibe
effect Ask {
  Get -> Int
}

let f = () -> Int with Ask::Get {
  perform Ask::Get
}

let side = (n: Int) -> Int with Stdout {
  println("x")
  n
}

test "t" {
  handle {
    side(f())
  } with Ask {
    Get => resume(42)
  }
}
```

```console
handle of effect 'Ask' cannot be compiled here. Every perform this handle covers has to be
statically visible to it, so the handled body may only: perform directly, call a named top-level
`fn`, or call a closure literal that carries an effect row annotation. A call through a local
binding or a closure parameter hides the perform and is what this rejects -- move the `handle`
into the function that performs, or replace the indirect call with a direct one. (ADR-0076
evidence-passing migration; the replay engine that used to accept this shape was removed.)
```

`side` **is** a named top-level binding and the call **is** direct, so the diagnostic does not describe what was rejected.

## Measured (2026-08-18, wasm-gc lane, stage2 from `main`)

The variable is whether the callee carries a row. Nothing else:

| callee in the handled body | result |
|---|---|
| `let side = (n: Int) -> Int { n }` | compiles |
| `let side = [T](n: T) -> T { n }` (generic, rowless) | compiles |
| `let side = (n: Int) -> Int with Stdout { .. }` | **rejected** |

The same rejection covers the shapes that reach a row through a desugar, which is how it is most likely to be met in practice:

```vibe
test "t" { handle { println("x") f() } with Ask { Get => resume(42) } }   // rejected
test "t" { handle { assert_eq(f(), 42) } with Ask { Get => resume(42) } } // rejected
test "t" { handle { inspect(f(), "42") } with Ask { Get => resume(42) } } // rejected
```

All three are rejected on `main` **today** (measured against the `24beb69` stage2), before #2107 gave the print builtins their row. What #2107 changed is only how easy the shape is to hit: `inspect` imported from `@vibe/core` now carries `Stdout`, so four fixtures that wrapped `inspect(run_x(), "42")` in a `handle` had to hoist the call out (#2108).

## Why this is wrong, not just strict

[...]

Issue #1979, opened 2026-08-16
Single source of truth for the borrow-returning builtin list (three copies already drifted once)


mizchi:
The set of borrow-returning builtins (`Array::get`, `MutList::get`, `Map::get`, `__index`, `__rec_field`) is hardcoded in at least three places:

- `is_borrow_ret_fname` (perceus/perceus.vibe)
- the `known` seed of `compute_borrow_returning_names` (same file)
- the `is_borrow_ret` chain in `classify_let_value_heap` (codegen/expr/compile_expr_tail.vibe)

They already drifted once: `MutList::get` was added to `is_borrow_ret_fname` (and the codegen chain) when MutList landed, but not to the classifier seed — caught only by Codex review on #1932. The failure mode of this drift class is the bad one: a function wrapping the forgotten builtin classifies as owned-returning, RC callers drop an interior view, and the owning container's reclamation double-frees it — silent memory corruption, not a diagnostic.

## Proposed fix

One exported function (or constant list) in the perceus package, e.g. `borrow_ret_builtin_names() -> Array[String]` / a membership predicate, consumed by all three sites. The codegen site currently also short-circuits on the same strings inline for speed; a shared predicate keeps that (string comparisons inline fine) while making "add a borrow-returning builtin" a one-place change.

Optionally add a `scripts/review_lint.vibex` pattern (per the re-occurrence-prevention policy in CLAUDE.md) rejecting new hardcoded occurrences of these names outside the shared definition.

Context: #1932 review thread, #1915.

Issue #1955, opened 2026-08-16
Define a source-level HostStream next protocol and lifecycle


mizchi:
## Decision needed

HostStream already supports `host_stream_next` as a one-byte Async operation with EOS `-1`, but the source-level protocol is undecided. Candidates include `Option[Int] with Async`, `Future[Option[Int]] with Async`, or an AsyncIter-style `(Int, Self)` result. The handle is stateful and unchanged, so carrying Self may be meaningless.

## Scope

- Choose one source spelling consistent with the language policy that authority travels in the row
- Define EOS, close, idempotence, early stop, and ownership
- Implement it as a prelude wrapper over existing host primitives where possible
- Do not overload the name while eager `Stream::next` still exists

## Acceptance criteria

- Buffered and parking producers, repeated reads, EOS, early close, and concurrent streams
- No representation mixing with legacy Stream or generic AsyncIter
- WIT/component mapping remains nominal and exact
- WASI P3 and named-host-stream gates pass

Extracted from #1538.

mizchi:
Completed by #2151.  now exposes ,  is idempotent, EOS/repeated reads/early close/concurrent streams are covered, nominal WIT mapping remains exact, and the named-host-stream plus pinned Wasmtime 47 WASI P3 gates pass. Legacy  and generic AsyncIter representations remain separate.

mizchi:
Completion details for #2151: HostStream next now exposes Option of Int with Async, HostStream close is idempotent, EOS, repeated reads, early close, and concurrent streams are covered, nominal WIT mapping remains exact, and the named-host-stream plus pinned Wasmtime 47 WASI P3 gates pass. Legacy Stream and generic AsyncIter representations remain separate.

PR #2583, opened 2026-09-09
codegen: one pure-builtin allowlist, not two copies (#2373)

### LINKED ISSUE #2373: edp_pure_builtin_names does not mirror idp_pure_builtin_names, and nothing checks that it does
`edp_pure_builtin_names`' doc comment in `lib/@vibe/compiler/codegen/common_base/inline_direct_perform.vibe` says:

> Mirrors inline_direct_perform.vibe's idp_pure_builtin_names allowlist exactly (same hand-audited, individually-verified set -- see that file's doc comment for the auditing rule).

Measured on `015dc72`, it does not. Three names are in `idp` and absent from `edp`:

| name | why it is benign today |
|---|---|
| `Array::concat` | `edp_has_unsafe_construct` reaches it through the registry first-order fallback added in #2109 |
| `HostStream::next` | `edp_is_pure_builtin_call` reaches it through `idp_is_host_stream_named_raw` |
| `HostStream::close` | same |

`edp` ⊆ `idp` holds; only the other direction is broken. That is the harmless direction, because `edp`'s consumer has fallbacks and `idp`'s (`scps_named_call_ok`, suspend-class lowering) does not — which is exactly why the divergence went unnoticed.

## Why this is worth an issue rather than a drive-by fix

Two things, and only the second is really about these lists.

**The doc comment is false.** A comment asserting an invariant nothing enforces is the kind of statement that gets trusted by the next person adding a builtin. Either make it true (add the three names, redundant but harmless) or rewrite it to say what actually holds — `edp` is a subset, and `edp`'s other predicates cover the rest.

**The recurring defect is not mirror drift; it is that adding a builtin means editing several hand-maintained classification lists, and nothing points at the ones that were missed.** Twice in two commits on #2372's branch:

- #2344: `Bytes::index_of_bytes` was missing from `builtin_is_non_allocating`, so `builtin_allocates` defaulted it to allocating and a correct `#zero_alloc` function was rejected.
- #2372: it was missing from `idp_pure_builtin_names` / `edp_pure_builtin_names`, so a suspend-class handle body calling it was refused as an opaque callee while `String::index_of` in the same position was accepted.

Both were found by review, not by a gate. A strict mirror gate would not have caught either: in the second case the name was absent from *both* lists, so the two agreed.

## What a useful check would have to be

Not "every registry builtin is on the list" — the list is narrow on purpose (entries are audited individually against the checker's builtin type table, and `builtins_misc.vibe` mixes pure entries with `print` / `println` / `throw` in the same file, where `throw` is a perform in disguise).

Candidates worth thinking about:

1. **A registry column instead of a list.** The audit result is a property of the builtin, so it could live in the registry row next to the type and the existing three booleans, with the lists derived. That collapses N hand-maintained lists into one place a new builtin cannot be added without touching.
2. **A "new registry name is unclassified" gate.** A registry name absent from every classification list is not necessarily wrong, but it is always unreviewed. A ratchet (existing names allowlisted, deletions only) would make the omission visible at the point it is introduced.
3. **The mirror check on its own**, which is cheap and lexical but, as above, catches the narrower half of the class.

Per the gate discipline in `CLAUDE.md`, whichever is chosen needs a `scripts/<name>_test.sh` that mutates a real input and asserts the gate FAILS — and the red test has to be shown to actually land, not merely to pass.

## Triage

`refactoring` / P2: nothing that works today is broken. The two defects it would have prevented are already fixed.

mizchi:
Closes #2373.

## The divergence is still there

`edp_pure_builtin_names`' doc comment says it "mirrors `idp_pure_builtin_names` **exactly**". Measured on this tree, it does not — 72 names vs 69:

| missing from `edp` | why it never failed |
|---|---|
| `Array::concat` | `edp_has_unsafe_construct` reaches it through the registry first-order fallback (#2109) |
| `HostStream::next` | reached through the raw-half prefix predicates |
| `HostStream::close` | same |

The issue's framing is the important part: the two copies gate consumers whose **failure directions are opposite**. The idp side gates suspend-class lowering — `scps_named_call_ok` *refuses* a program naming a callee it does not recognize. The edp side has fallbacks, so a name missing there only sinks eligibility silently. Neither copy is the safe one to forget, which is exactly how the divergence outlived a comment asserting it could not exist.

## Delete the copy rather than gate it

`edp_is_pure_builtin_call` now reads `idp_pure_builtin_names_table`. The second list and its own module-level memo are gone (−115 lines of duplicated data).

This follows the repo's own #1979 consolidation in `core/builtin_registry.vibe` — *"there is one spelling of the names, not two that must be kept equal"* — and it is why this lands **without** a mirror gate. A script asserting two lists are equal is a proxy for the property; `AGENTS.md`'s gate discipline says to remove the structure that needs a proxy rather than to check the proxy. There is no second list for it to stand in for.

The three names becoming visible to the evidence pass is a widening in the direction its own fallbacks already reached them by. The audit criterion — *"can neither perform a user effect nor invoke a function-typed argument"* — is one criterion for both consumers, so a name belongs to both lists or to neither.

## What is pinned

`verify_pure_builtin_name_agreement` (exported, declared in `index.vpkg`) + `lib/@vibe/compiler/tests/pure_builtin_name_agreement_test.vibe`:

- every listed name is accepted by **both** predicates;
- `Array::push` and `println` by **neither** — the negative control is the half that makes the positive half mean anything, since a predicate answering `true` unconditionally passes the loop alone.

**Red-tested both ways**, with mutations that are type-correct so that only the assertion catches them:

| mutation | result |
|---|---|
| re-fork `edp_is_pure_builtin_call` onto a two-name table | FAIL |
| `fn edp_is_pure_builtin_call(nm) { String::length(nm) >= 0 }` | FAIL |
| restored | ok, 1 passed |

## Also

Repoints the four fixture comments and the `tests/gates/mid/run.sh` comment that named the now-deleted function.

## Verification

All against a stage2 built from this checkout (`_build/selfhost/generations/cfg-spans-emission-2026-09-04_cd9eb4f/stage2.wasm`), passed explicitly via `VIBE_TEST_CLI_WASM` / `VIBE_STAGE2_WASM` rather than letting the seed answer:

- `pure_builtin_name_agreement_test.vibe` — red → green as above
- `fixtures/effect_suspend_bytes_index_of_bytes_test.vibe` — 5 tests ok
- `fixtures/async_future_boundary_resolved_test.vibe` — compiles ok
- `scripts/vibe_fmt.sh --check` — clean on all three changed `lib/` files
- `scripts/compiler_gate.sh early mid late` — all lanes ok
- `pkf run pre-commit` — review-regressions lint, architecture-debt lint, doc-path citations all ok

## Not in scope

The issue's second, broader defect stands: *adding a builtin means editing several hand-maintained classification lists, and nothing points at the ones that were missed* (`builtin_is_non_allocating`, `scps_is_safe_mut_builtin`, `borrow_ret_builtin_name_table`, …). That is the registry-column or unclassified-name-ratchet work, a different change. This PR removes one whole list from that set and closes the mirror half; the census half is worth its own issue.

https://claude.ai/code/session_01Q9zbiDqkfF7xYQ3v8aCQ3b

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q9zbiDqkfF7xYQ3v8aCQ3b)_

### DIFF
--- a/fixtures/async_future_boundary_resolved_test.vibe
+++ b/fixtures/async_future_boundary_resolved_test.vibe
@@ -1,8 +1,9 @@
 /// ADR-0089 Decision 2 (#1218): the Future cell primitives are inert
 /// callees for the evidence migration (`Future::pending`/`Future::ready`
 /// allocate the [state, payload] cell, `Future::resolve` is two stores;
-/// none can perform). Before they joined the pure-builtin allowlists
-/// (idp/edp_pure_builtin_names), this program was REJECTED outright: the
+/// none can perform). Before they joined the pure-builtin allowlist
+/// (idp_pure_builtin_names; #2373 folded the evidence pass's second copy of
+/// it into that one), this program was REJECTED outright: the
 /// entry boundary wraps the whole Async-row entry body in a
 /// `handle ... with Async`, and one opaque callee name sank the entire
 /// body to "not eligible for evidence-passing migration".
--- a/fixtures/effect_handle_call_evidence_pure_helper.vibe
+++ b/fixtures/effect_handle_call_evidence_pure_helper.vibe
@@ -2,7 +2,8 @@
 // user-defined PURE helper function (no `with` clause at all, so the
 // checker has already proven it performs no effect) that is neither a
 // needing function itself nor one of the hand-audited
-// `edp_pure_builtin_names`.
+// `idp_pure_builtin_names` (#2373 folded the evidence pass's second copy of
+// that list into the one the direct-perform pass reads).
 //
 // edp_has_unsafe_construct's ECall case used to unconditionally reject any
 // call whose callee wasn't "perform"/"resume", a hand-listed pure
--- a/fixtures/effect_suspend_bytes_index_of_bytes_test.vibe
+++ b/fixtures/effect_suspend_bytes_index_of_bytes_test.vibe
@@ -1,6 +1,6 @@
 // The `Bytes` search builtins (#2345) have to be on the hand-audited
-// pure-builtin allowlist (idp_pure_builtin_names / edp_pure_builtin_names,
-// codegen/common_base/inline_direct_perform.vibe) -- the same list
+// pure-builtin allowlist (idp_pure_builtin_names, the one list both passes
+// read since #2373 -- codegen/common_base/inline_direct_perform.vibe) -- the same list
 // `Array::concat` joined in #1536, and the one `String::index_of` has been on
 // all along.
 //
--- a/fixtures/gc_backend_effect_pure_builtin_index.vibe
+++ b/fixtures/gc_backend_effect_pure_builtin_index.vibe
@@ -1,6 +1,6 @@
 // ADR-0076 (#817) gc-backend follow-up: `edp_has_unsafe_construct`'s
-// pure-builtin allowlist (idp_pure_builtin_names/edp_pure_builtin_names,
-// codegen/common_base/inline_direct_perform.vibe) was missing `__index`
+// pure-builtin allowlist (idp_pure_builtin_names, the one list both passes
+// read since #2373 -- codegen/common_base/inline_direct_perform.vibe) was missing `__index`
 // (what `obj[i]` desugars to), `Array::get`/`Map::get`/`Bytes::get`, and
 // `Array::length`/`Map::size`/`Bytes::length` -- all seven individually
 // checker-verified pure (no possible perform, no function-typed param).
--- a/lib/@vibe/compiler/codegen/common_base/index.vpkg
+++ b/lib/@vibe/compiler/codegen/common_base/index.vpkg
@@ -279,6 +279,11 @@ fn lower_optional_performs(stmts: Array[Stmt], resolution: Array[(String, String
 // --- evidence_dict_pass
 fn evidence_dict_pass(stmts: Array[Stmt], entry_name: String) -> Array[String]
 fn edp_bind_census(stmts: Array[Stmt], names: Array[String]) -> Array[Bool]
+// #2373: the direct-perform and evidence passes read ONE pure-builtin
+// allowlist; this throws if a listed name is rejected by either predicate, or
+// if a non-member is accepted by one. Exported for
+// tests/pure_builtin_name_agreement_test.vibe.
+fn verify_pure_builtin_name_agreement() -> Unit with Exception
 
 // --- forin_discard_pass (#1916): wrap every discard-position `for` as
 // `ESeq(for .., EUnit)` so the linear ESeq lowering compiles it without the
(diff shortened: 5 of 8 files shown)


mizchi:
## 実測: この widening は出力を変えない (differential, 280 fixtures)

PR 本文はこの変更が安全な理由を audit criterion からの**議論**で述べていました。レビューで真っ先に出る問いは「edp が 3 名前を新たに受理することで、コンパイル結果が変わるのか」なので、議論ではなく実測に置き換えます。

**方法**: 同一 checkout から stage2 を 2 本ビルドし (baseline = `cd9eb4f` = このブランチの merge-base、with-change = `c74bbe7`)、`handle` / `perform` / `with Async` を含む fixture 全件を両方でコンパイルして、wasm のバイト列と診断を比較しました。

| | |
|---|---:|
| 対象 fixture (`grep -l 'handle\|perform\|with Async' fixtures/*.vibe`) | 280 |
| wasm を生成し、**バイト単位で一致** | 188 |
| 両者が**同一の診断**で拒否 (`resume` 値参照の suspend class、`__DATA__` 尾を持つもの 等) | 92 |
| **差分** | **0** |

個別 probe も同じ答えでした:

- `Array::concat` を needing function の本体 + handle 本体で使う probe → 同一 (`12ac5aff…`, 4231 bytes)
- `Array::concat` を suspend-class handle 本体の `for x in Array::concat(a, b)` で使う probe → 同一 (`48e5e4a0…`, 4271 bytes)
- `HostStream::next` / `HostStream::close` — 私が組んだ Async entry の probe はホスト provider の配線が足りず両者とも拒否されましたが、**診断が byte-identical** でした。この 2 つが inert な理由は元から list 内に書かれています (`inline_direct_perform.vibe`, #1955: lowering 前に `__hs_next_option` / `__hs_next` へ retarget される)

つまり `Array::concat` は issue #2373 が言うとおり既に registry first-order fallback (#2109) 経由で到達済み、`HostStream::*` は effect pass が走る前に書き換えられている。**3 名前とも、edp が受理するようになっても観測できる差はありません。**

## 自己レビュー

6 観点 (widening / 初期化順序と memo / contract 面 / テストの厳密さ / 記述の真偽 / 他のコピーの有無) で独立にレビューし、上がった 20 件を 1 件あたり 3 つの独立した反証エージェント (correctness / already-handled / repro) にかけました。**多数決を生き残った指摘は 0 件**です。棄却理由として有用だったもの:

- `docs/effect-evidence-passing.md` に残る `edp_pure_builtin_names` の言及を書き換えなかったのは正しい — 同ファイル冒頭の **#1461 の綴り注記**が「作業ログは当時の綴りのまま残す」と明示しており、`check_doc_path_citations.sh` が archive を除外するのと同じ理由。実際このファイルには既に存在しない識別子が 22 個あり、この PR だけに全件 sweep を課すのは main が満たしていない基準でした
- gc backend の Bool 判定リストが 3 箇所にコピーされている件は**同型ではない** — gc 側は名前リストの前に checker 由来の typed channel (`gc_bool_render_offsets`) を見るので、名前の取りこぼしは "wrong" ではなく "missed classification" になり、`bool_render_channel_test.vibe` が両側を pin している

副産物としてこの PR とは無関係な観察が 1 つ: `lib/@vibe/compiler/checker_purity.vibe` (`analyze_purity` / `builtin_purity`) は production consumer が 0 で、`#634` を指すコメントも実際の pure-test cache 判定 (`file_tests_cacheable`) とは別経路です。dead code の可能性があり、必要なら別 issue に切ります。

CI は 29 checks すべて green (`ci-required` 含む)。

---

PR #2635, opened 2026-09-11
codegen: a module's oracle context is its real dependency closure (#2388)

mizchi:
The item [#2622](https://github.com/mizchi/vibe-lang/pull/2622) left first on the #2388 line (parent epic #2386), and the measurement it produces.

## The gap

The module-mode oracles handed each module the interface of **every other file** in the closure. That is wider than any real per-module compile — a dependency could read declarations from the entry that imports it — so a whole-program pass leaning on an unrelated module's declarations stayed green. Codex raised it twice on #2622 and I could not express the restriction soundly there: merge order puts a materialized contract ahead of the implementation its own derives read, and `collect_source_groups_fs` returns one group for the whole closure.

## The edge set already existed

`module_plan_data_fs` returns each module's resolved **direct** dependency paths — the same sequence the check lane projects environments onto through `project_exact_dependency_envs`, and what `ModuleJob.dep_envs` carries in production. This exports it, maps it onto the merge's file ids, closes it transitively, and gives each module exactly that context. A file with no row sees **nothing** rather than everything: a missing edge must narrow the context, never widen it.

## The mapping is checked, not trusted

`prelude_oracle_merge` now returns the per-file **full paths**, because the merge's own `files` array holds basenames — and this closure has many `index.vibe`. Matching on basenames left every dependency row empty, so every module saw nothing, and the run reported:

```
keyed missing=5 extra=0 copies=313 content=33 …
  first_missing=let:MutMap::equals__N6_String__N3_Int
  let:MutSet::equals
    whole<… MutMap::equals(a.inner, b.inner) …>
    split<… (a.inner == b.inner) …>
```

Confident, plausible, and entirely an artifact. The report now carries `graph plan_unmatched= files_unmatched=`, the two-file test pins it at `0 0`, and reverting to the basename match turns that assertion red (`2 2`) — so this reading can never again pass for a finding.

## Measured

Compiler closure (`codegen_lexer_test.vibe`, 189 files), each module seeing only what it imports:

```
FULL stmts=4764 split=4836 modules=191
  keyed missing=0 extra=0 copies=321 content=1 renames=0 dup_keys=254
  graph plan_unmatched=0 files_unmatched=0
```

Against the previous context (every other file) the only movement is `content=0 → 1`. **That one difference is real**: `PerceusAction::equals` calls `PerceusActionKind::equals(a.kind, b.kind)` whole-program and compares `a.kind == b.kind` per module. `struct PerceusAction` is declared in `perceus/index.vpkg` — materialized into a generated types module — while `enum PerceusActionKind` lives in `perceus.vibe`, and a contract does not import the implementation it fronts, so the module that owns the struct cannot see that the field type has a structural comparator.

Harmless for this type: `PerceusActionKind` is a nullary enum, so both forms compare tags. Not harmless as a shape: a payload-carrying field type in the same position would get reference equality, which is the P0 class this epic exists for. **That is the per-module lane's one open item on this corpus**, and it is now located rather than suspected — either the derive defers to the link, where both declarations are visible, or the contract carries enough about the field type for the owning module to emit the call.

## Validation

- `prelude_module_oracle_test.vibe` (3 tests, now including the graph assertion), `desugar_trait_dicts_module_test.vibe` (10), `fixtures/hoisted_lambda_owner_names_test.vibe` (2, linear and gc lanes) pass.
- Red verified for the new assertion by restoring the basename match.
- stage0 → stage1 → stage2 selfhost build clean; `scripts/vibe_fmt.sh --check` clean on every changed file; `scripts/compiler_gate.sh` running locally as I open this.

https://claude.ai/code/session_01NAeZzkckdAwtQiubBFpdmM

---
_Generated by [Claude Code](https://claude.ai/code/session_01NAeZzkckdAwtQiubBFpdmM)_

### DIFF
--- a/lib/@vibe/compiler/codegen/wasi/index.vpkg
+++ b/lib/@vibe/compiler/codegen/wasi/index.vpkg
@@ -63,8 +63,8 @@ fn effect_lowering_prelude(stmts: Array[Stmt], entry_name: String, linked_import
 fn late_dce_prune_module(mod_bytes: Bytes) -> Bytes
 fn late_dce_prune_module_for(mod_bytes: Bytes, entry_name: String, exports_stripped: Bool) -> Bytes
 fn late_dce_prune_refusal() -> Int
-fn prelude_pass_module_oracle_step(k: Int, a: Array[Stmt], c: Array[Stmt], c_file_id: Array[Int], file_count: Int, entry_name: String, checker_eq_offsets: Array[Int], checker_eq_type_keys: Array[String], render: (Stmt) -> String) -> String with Exception
-fn prelude_module_full_oracle_line(a: Array[Stmt], c: Array[Stmt], c_file_id: Array[Int], file_count: Int, entry_name: String, checker_eq_offsets: Array[Int], checker_eq_type_keys: Array[String], render: (Stmt) -> String) -> String with Exception
+fn prelude_pass_module_oracle_step(k: Int, a: Array[Stmt], c: Array[Stmt], c_file_id: Array[Int], file_count: Int, dep_ids: Array[Array[Int]], entry_name: String, checker_eq_offsets: Array[Int], checker_eq_type_keys: Array[String], render: (Stmt) -> String) -> String with Exception
+fn prelude_module_full_oracle_line(a: Array[Stmt], c: Array[Stmt], c_file_id: Array[Int], file_count: Int, dep_ids: Array[Array[Int]], entry_name: String, checker_eq_offsets: Array[Int], checker_eq_type_keys: Array[String], render: (Stmt) -> String) -> String with Exception
 fn compile_wasi_module_replayed_impl(stmts: Array[Stmt], entry_name: String) -> Bytes with Exception
 fn compile_wasi_module_sliced_impl(stmts: Array[Stmt], entry_name: String) -> Bytes with Exception
 fn compile_wasi_module_rc_impl(stmts_in: Array[Stmt], entry_name: String) -> Bytes with Exception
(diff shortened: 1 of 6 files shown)


mizchi:
Closing this: **main already has it, and main's version is better.** I built it against `dbde5bf` without checking that the line had moved; `f866bbe` carries the same slice from another session.

What landed there that this does not have:

- **One hop, not the transitive closure.** `b8c65f0` identifies exactly the bug in what I wrote: closing transitively crosses PRIVATE imports — when A imports B and B imports C privately, the BFS hands A the declarations of C, context no real compile of A has. That is the failure mode that makes an oracle's green meaningless, and this PR would have reintroduced it.
- **Per-file header scans** (`load_or_parse_module_header_fs` per flat source) rather than a walk of `module_plan_data_fs`'s plan, which also handles a source that appears in two groups: every flat index for a path gets the edge, not the first.
- **A throw, not a counter,** when the closure's indexing and the merge's file count disagree. My `graph plan_unmatched= files_unmatched=` reports it; main refuses to answer at all, which is the stronger form of the same idea.

The one thing worth carrying forward from this branch is not code but a measurement, and it is already on #2388: matching the plan's paths against the merge's `files` array silently produced empty rows, because that array holds **basenames**, and the run then reported a confident `missing=5 content=33` with a plausible first example (`MutSet::equals` comparing `a.inner == b.inner`). Main's `throw` on a count mismatch would not catch that particular shape — the counts agree, the contents do not — so if `oracle_import_closure` ever resolves paths against a basename list, the same false finding is available. It resolves against `flat` paths today, so nothing to do; noting it because I spent an hour on it.

No changes ported, branch reset to main.

---

Issue #2510, opened 2026-09-04
Per-module prelude: scope the desugar passes and the may_return / borrow-param analyses to the changed module, carrying their whole-program tables as interface data


mizchi:
Part of #2494, the other half of #2388. With the persisted body cache on, a warm compile of the compiler closure still spends 1.3 to 1.5 s in `effect_lowering_prelude` (`desugar_trait_dicts_with_typed_eq` 0.74, `evidence_dict_pass` 0.37), 0.6 s re-lexing and re-parsing every file, and 0.6 s in `compute_may_return_view_fns` / `compute_borrow_param_user_fns`, all over the whole merged program, and holds the whole merged AST (warm heap 650 to 980 MB).

What has to become per-module:

- the parse: persist the binary AST per file (`ast_binary.vibe`), keyed by source fingerprint, and merge from artifacts instead of text;
- the prelude's per-function analyses (`may_return` views, borrow-param sets, evidence dictionaries), keyed per function next to the bodies in the same artifact;
- the desugar passes, run only on changed modules. Their whole-program reads (`afe_field_types_cell`, `throw_kind_cell`, `optional_param_tbl_set`, the trait namespace tables) become interface data carried in the module artifact, which is the same move that lets `infer_` be retired (#2493's desugar issue).

Done when: a leaf-module edit re-runs the prelude for that module only, the warm compile's whole-program phases are link-time only, and the mid-size memory KPI shows the live set bounded by the edited module plus interfaces.

mizchi:
## Criterion 5 measured (it is not met), and criterion 4 is blocked on a bootstrap bump

PR #2660. Two of the five "done when" items had never been checked; both are now, and together they re-order what is left.

### Criterion 5: the per-module prelude costs a flat ~117 MB and recovers none of it

The wall-time figure next to this (+5.3%) bounds the split's cost in time. Nothing measured allocation, which is what this criterion is actually about. `scripts/prelude_split_memory.sh` now does, on the compiler's own closure (190 files, 191 modules / 4728 statements — the split shape is asserted per run, so no row can be vacuous):

| lane | cold | after a one-module edit | unchanged |
|---|---:|---:|---:|
| whole-program | 821,534,940 | 765,264,132 | 510,478,340 |
| per-module (split) | 938,747,492 | 882,476,684 | 627,625,364 |
| **split − whole** | **+117,212,552** | **+117,212,552** | **+117,147,024** |

[...]

PR #2668, opened 2026-09-11
The per-file binary AST cache (#2510 criterion 4), and the 22× `vibe symbols` fix the measurement found (#2670)

mizchi:
Follow-up to #2665 (merged at `556075b`). Same branch, restarted from `main`. Four commits: the cache, the measurement that says what to do next, the one actionable finding that measurement turned up, and the measurement that sharpens what #2669 is choosing between.

The codec moves into `lib/@vibe/compiler/cache/` — which is what it is, a serialization of one module's parse, keyed and stored like the module-header text beside it — and gains the policy and loader wiring #2510's fourth criterion asks for. It is behind `VIBE_EXPERIMENTAL_AST_CACHE` and **off by default**.

The honest headline is in the title, and it is the reason to read this PR: **with the cache on, a warm compile of the compiler's own closure allocates the same bytes as with it off.** The second commit establishes why, and it changes what criterion 5 should be. The third and fourth are what the same measuring method found when pointed elsewhere — one a 22× fix, one a result that reframes the next project.

## What is proven

Output equivalence — the property a cache must have before anything else matters. `codegen_lexer_test.vibe` through the FS lane, one process per run, a fresh `VIBE_BUILD_CACHE_DIR` per lane:

| run | bytes | sha256 |
|---|---:|---|
| off | 4,034,228 | `c5e5134c9a097852` |
| on, cold | 4,034,228 | `c5e5134c9a097852` |
| on, warm | 4,034,228 | `c5e5134c9a097852` |

The cold run really did store: **191 `selfhost_ast_v1` entries**, matching the 191 modules #2660 measured in the same closure. So the format, the key, the envelope and the fail-open decode all work.

## What is NOT proven

Heap, same protocol:

| | cold | warm |
|---|---:|---:|
| off | 872,271,128 | 571,165,952 |
| cache on | 895,355,616 | 571,165,952 |
| Δ | **+23 MB** (encode + store) | **0** |

Byte-identical warm. Not "within noise" — the same number. Allocation is deterministic on this lane, so **the cache code did not execute on the warm path at all**.

## Why — measured, second commit

A warm compile profiled with `node --cpu-prof`, through a stage2 built with `VIBE_WASM_NAMES=1`. Without that flag every frame is `wasm-function[N]`; the first attempt produced exactly that and said nothing.

| area | share of 3.66s |
|---|---:|
| compiler/codegen | 28.7% |
| runtime helpers (`__rt_*`) | 25.6% |
| compiler/normalize | 11.5% |
| host JS / other | 8.5% |
| **parser** | **6.9%** |
| core | 6.7% |
| compiler/core | 5.7% |
| compiler/perceus | 5.2% |
| **compiler/loader** | **0.1%** |

**The back end dominates**: codegen + normalize + perceus is 45.4%. The parser is 6.9% — the *ceiling* on what a perfect AST cache can remove from a warm build. And the loader, where such a cache must live because it is the only lane with an `Fs` row, is **0.1%**.

That last number is the direct explanation for the byte-identical warm heap: the code that would consult the cache barely runs. The cache is not misconfigured; it is attached to a part of the build a warm run has almost stopped performing.

Structurally: the merge's parse is `parse_program_with_path`, which is pure (`with Exception`, no `Fs`) and cannot consult any cache; the header pass, which does have `Fs`, returns from the module-header *text* cache without parsing on a warm run. A prefetch was added to that warm branch to hand the merge a stored AST through the shared memo (#2386's sharing point) — the heap did not move by a byte either.

*(CPU share, not allocation; the KPI is allocation and the two are correlated rather than identical. N=1 — enough for the structural read, since a 0.1% area is not the 30% that matters under different sampling, and not enough to rank the 5–7% rows against each other.)*

## What this means for criterion 5

**It is not reachable by caching parses.** What a warm build repeats is the back end running over the whole program, and the constraint on fixing *that* is already in `docs/incremental-build.md` from #2647: a cached function body contains call immediates that depend on **global function index assignment**, so a per-module body cache cannot replay a body into a build where indices moved. Filed as #2669 — and the fourth commit measures how far they actually move.

## Third commit: `vibe symbols lib`, 42.04s → 1.91s (#2670)

The same profiling method, pointed at the lane the repo's own tooling guidance tells an LLM to use for code navigation. `vibe symbols lib` looked parse-bound — 88% of the profile sat under a parser entry point. It was not:

| `vibe symbols lib` | before | after |
|---|---:|---:|
| wall | 42.04s | **1.91s** |
| `offset_to_line_col` | 41,283 ms (83.7%) | 7.3 ms (0.35%) |
| `line_start_offsets` | — | 68.7 ms (3.3%) |
| lex + parse, actual | 3.3% | 3.3% |

`offset_to_line_col` counts newlines from offset 0 on every call, and `symbol_spans_with_docs` calls it once per declaration — O(declarations × file length) per file, worst on the compiler's own large files. `line_start_offsets` builds the index once per file; `offset_to_line_col_indexed` binary-searches it.

**The 26,832 output rows are byte-identical** (`cmp -s`). That equality is the whole correctness condition: this changes how a line number is computed, not which one.

`offset_to_line_col` stays — it is right for a one-shot lookup (a single diagnostic), and several callers are exactly that. What was wrong was calling it in a loop, so the loop caller gets an indexed twin rather than the simple one being removed.

`line_index_test.vibe` asserts agreement at **every** offset of a source including one past the end, not on picked shapes: empty input, no-newline input, leading/trailing/consecutive newlines (what an index built by "push i+1 when byte i is 0x0A" is likeliest to get wrong), multi-byte UTF-8, and a whole realistic source. The past-the-end case checks the clamp lands on the **last** line rather than line 1, so it cannot pass vacuously. Newlines are found by byte (ADR-0108), which is what makes a byte scan and a character scan agree — 0x0A cannot occur inside a UTF-8 sequence.

After the fix the profile is bound by `readFileSync` 9.0%, `lex_one_token` 6.2%, `keyword_lookup` 5.2% — the shape a symbol sweep should have.

Two other loop callers exist (`grep.vibe` per hit, `deprecated_scan.vibe` per finding) with the same shape. Not touched: they were not measured, and #2670 records them as a lead rather than a finding.

## Fourth commit: how far function indices actually move (#2669)

`docs/incremental-build.md` has said since #2647 that a cached body "cannot be replayed into a build where indices moved". How far they move was never measured. `scripts/wasm_index_stability.mjs` compares two builds made with `VIBE_WASM_NAMES=1`, matching functions **by name** — by index would assume the stability being measured.

| | body-only edit | one added function |
|---|---:|---:|
| index kept | 4286 (100%) | 216 (5.04%) |
| distinct shift deltas | — | **1** (`+1`) |
| body byte-identical | 4285 (99.98%) | 1297 (30.26%) |
| body differs, SAME length | 0 | 2988 (69.72%) |
| relocation sites explained | — | 31256, all `+1` |
| relocation sites unexplained | — | 0 |

**A body-only edit moves nothing** — the constraint does not apply to the common edit at all. **Adding a function shifts every later index by exactly one**, a single delta, so the differing bodies differ at identical length and an in-place patch is arithmetically possible.

**But 25 of the 31256 sites are `i64.const`, not `call`**: a first-class function value is emitted as `i64.const (idx*4+2)`. It moves like a call immediate and looks like any other constant, so a relocation pass keyed on the `call` opcode would leave those wrong — silently. That is a real argument between #2669's options, and it is not in the wall-clock numbers. Full writeup in [the issue comment](https://github.com/mizchi/vibe-lang/issues/2669#issuecomment-5639146798); the architecture is still the maintainer's call.

`unexplained` prints beside `explained` because "0 unexplained" means nothing alone — and it did not hold on the first run (the `i64.const` values were 7 unexplained bodies until the decoder was added). Red-tested by flipping a body byte the classifier cannot attribute, mutation confirmed to have landed first: `unexplained=1`, naming function and offset.

## Why land the cache anyway

The parts that are finished are the parts that are hard to get right:

- **the format** — 54 tests, including `ast_binary_parse_equivalence_test.vibe`, which holds the codec against the **parser** on real compiler sources rather than against itself;
- **the fail-open rule** — a torn entry, a trailing-byte entry and an undecodable one are all misses, never a wrong AST. The trailing-byte case matters specifically: `ast_binary_read_module` stops when the statement array ends, so without that check a corrupt tail decodes "successfully" into a *prefix* of the real module;
- **the key** — inherits the codegen fingerprint, the resolution seed and the ambient `#cfg` set from `persistent_cache_version_tag`. The cfg set matters because the parser keeps or drops `#cfg`-gated statements by it, so the same text under another set is a different AST.

### Correcting one claim

An earlier version of this description said "zero default cost". The perf report shows that is **wrong for binary size**: selfcompile code is **2.84 MiB, +1.15%**, because the codec now compiles into the compiler whether or not the flag is set. Runtime is unaffected (memory −0.01%, `expr_eval` and `fib` ±0, no drift ≥±2% in the other 71 series). So: no runtime cost by default, ~33 KiB of compiler binary regardless — worth knowing when weighing whether to keep it while it is inert.

## Notes on shape

Three files rather than one, deliberately: `ast_binary.vibe` is the FORMAT, `ast_cache.vibe` the POLICY, `persistent_cache` the storage (whose "VART1" envelope and atomic tmp+rename publish this gets for free, #1326).

`ast_cache_slot_fs` returns a 0-or-1 array rather than the `Option` the cache offers, because its caller sits beside the parser chain and #708's RC-lane `unreachable` in `tk_name` reproduces when that chain is compiled next to an `Option` match — the same reason the cold lane next door dispatches on a plain `mut` Bool.

`compiler_sources_manifest.tsv` gains rows for both new files; the manifest gate caught their absence with an actionable message rather than silently dropping them (#740).

## Verification

54 codec tests, `line_index_test.vibe` (3/3), `located_parse_memo_test.vibe`, `prelude_split_fs_lane_test.vibe`, `check_ast_binary_tags.sh`, `check_builtin_shadowing.sh` (1019 files) + its self-test (10/10), `check_cli_line_termination.sh`, `check_vibe_fmt.sh` (1150 files, 0 violations), `check_doc_path_citations.sh`, `check_doc_commands.sh`, `check_gate_self_tests.sh`, `check_gate_portability.sh`, and the equivalence + heap + profile + index-stability protocols above. CI was green on `ed470bb` (29 success, 2 skipped, 0 failures) before the fourth commit.

https://claude.ai/code/session_01CCUq2wJKeDzguGzmx78HkZ


mizchi:
## The two ⚠️ drift rows are the corpus, not the parser — reproduced exactly

The perf report flags two `bytes_per_op` rows, and `bytes_per_op` is a **deterministic** series in that report (not advisory), so ±2% is meant to mean something:

> B/op `parser_bench.vibe::parse_lexer_vibe` **+11.31%** ⚠️ · B/op `parser_bench.vibe::parse_only_lexer_vibe` **+2.65%** ⚠️

Both benches take **`lib/@vibe/parser/lexer.vibe` as their input corpus** (`probe_parse_lexer_vibe` = `parse_program_count(lex(Fs::read_file("lib/@vibe/parser/lexer.vibe")))`). That is the file #2670 edited: 49,808 → 51,996 bytes, 66 → 68 statements. So the bench reads a bigger file — the parser is being handed more work, not doing the same work worse.

Measured rather than argued. One compiler (`bef330a` stage2), one runner, `--iters 100`, the branch tree against a worktree at `origin/main` — every tracked series in the file:

| bench | main | branch | Δ |
|---|---:|---:|---:|
| `parse_lexer_vibe` | 737,232 | 820,648 | **+11.31%** |
| `parse_only_lexer_vibe` | 434,736 | 446,240 | **+2.65%** |
| `parse_checker_vibe` | 7,673,080 | 7,673,080 | ±0 |
| `parse_parser_vibe` | 912,072 | 912,072 | ±0 |
| `parse_only_checker_vibe` | 4,164,784 | 4,164,784 | ±0 |
| `parse_only_parser_vibe` | 504,600 | 504,600 | ±0 |
| `parse_deep_binop_chain` | 132,116 | 132,116 | ±0 |
| `parse_stmt_let_if` | 59,136 | 59,136 | ±0 |
| `parse_type_fn_app` | 37,376 | 37,376 | ±0 |
| `parse_expr_infix_chain` | 60,672 | 60,672 | ±0 |
| `parse_expr_call_postfix` | 100,096 | 100,096 | ±0 |
| `parse_expr_match` | 80,640 | 80,640 | ±0 |
| `parse_wide_match` | 162,476 | 162,476 | ±0 |

**Both flagged figures reproduce to the reported digit**, and the other **eleven series are byte-identical**. `parse_checker_vibe` and `parse_parser_vibe` run the *same lexer and parser code* over files this PR does not touch, and they do not move at all — which is what rules out a parser regression. Only the two series whose corpus is the edited file moved.

The split between the two is where the added lines land: `parse_only_*` reuses memoized tokens, so it sees only parsing (+11,504 B, ~the 2 added statements), while `parse_lexer_vibe` re-reads and re-lexes every op, so it also pays token allocation for 2,188 more source bytes (+83,416 B total). The added code is denser in tokens per byte than `lexer.vibe`'s average, which is comment- and string-table-heavy — hence 11.31% against 2.65%.

### Two wrong turns on the way, worth recording

- My first isolation ran the two corpus versions through one compiler but put `Fs::read_file` **outside** the measured window. That answered +1.04%, and I nearly reported it as "mostly unexplained". The bench measures read+lex+parse together; excluding the read moved the answer by ten points.
- I then reached for an explanation ("at iters=1 B/op is dominated by startup") from a README table that turned out to be per-iteration **wall time**, not B/op — the tracked benches run at `--iters 100`. Discarded; the comparison above replaces it.

The instrument checks out against the harness independently: my probe's parse-only figure for the branch corpus was 446,308 against the bench's 446,240 — 0.015% apart.

### Nothing changed in response

This is a corpus artifact of benching real source files, not a regression, so there is nothing to fix in the diff and I have not widened the PR for it. It is worth knowing that these two series will move on *any* PR that edits `lib/@vibe/parser/lexer.vibe` — `parser_hotspot_probe.vibe` already carries a comment about the converse hazard (#1104, when those files became shims and the benches silently became micro-measurements). A one-line note in `bench/perf/README.md` naming this direction too would save the next reader the half hour it cost me; say the word and I will open it separately rather than attach it here.

CI is green on `4249e77` — 29 success, 2 skipped, 0 failures, plus `cli-install`.

---

Issue #2683, opened 2026-09-11
RC: a MapBuilder is never reclaimed (144 B per MapBuilder::new, more per set and freeze)


mizchi:
## Repro

```vibe
enum Tree { Leaf(Int); Node(Tree, Tree) }

let main = () -> Int {
  let mut total = 0
  let mut i = 0
  while i < 20000 {
    let mb = MapBuilder::new()
    MapBuilder::set(mb, "k", Leaf(i))
    let m = MapBuilder::freeze(mb)
    total = total + (if Map::has_key(m, "k") { 1 } else { 0 })
    i = i + 1
  }
  total
}
```

Measured with `scripts/measure_heap.mjs` on the RC lane (a stage2 built from 56cdc5b plus the #2671 branch, 2026-09-11), per iteration at N=20000:

| per-iteration shape | heap_used | per iteration |
|---|---:|---:|
| `let mb = MapBuilder::new()` (then `has_key`) | 2,881,436 B | 144 B |
| `new` + `set(mb, "k", i)` | 2,881,436 B | 144 B |
| `new` + `set(mb, "k", Leaf(i))` | 3,361,412 B | 168 B |
| `new` + `set(mb, "k", Leaf(i))` + `freeze` + `Map::has_key` | 3,841,412 B | 192 B |

The builder block and its storage are never released once the binding goes out of scope, and everything stored into it goes with it (the `Leaf` here, 24 B). The frozen `Map` costs another 24 B on top, so the Map is not reclaimed either on this spelling. The same loop with an `Array` builder (`let out: Array[Int] = []` + push, `build_arr` in `fixtures/rc_reclaim_leak_test.vibe`) stays flat, so this is specific to `MapBuilder` / `Map`, presumably the drop class of the builder block (the walking drop does not know its layout, or the builder is allocated outside the RC heap).

Not a use-after-free (nothing is dropped twice); a leak on every builder-shaped loop, which the compiler's own passes use (`MapBuilder` behind `Map` literals and json).

Found while adding the #2671 shapes to `fixtures/rc_reclaim_leak_test.vibe`: `MapBuilder::get`'s hit path also leaked the `None` it staged before its loop (16 B per hit, fixed on the #2671 branch), which is what led to measuring the builder itself.

Triage: `bug` / P1 (unbounded leak; no wrong answer) / not a blocker.

mizchi:
[...]

Issue #2682, opened 2026-09-11
RC: an owned call result passed straight into a borrowed argument position is never released


mizchi:
## Repro

```vibe
let build_arr = (n: Int) -> Array[Int] {
  let out: Array[Int] = []
  let mut j = 0
  while j < n {
    Array::push(out, j)
    j = j + 1
  }
  out
}

let main = () -> Int {
  let mut total = 0
  let mut i = 0
  while i < 20000 {
    total = total + Array::length(build_arr(3))
    i = i + 1
  }
  total
}
```

`Array::length` borrows its argument, so nothing owns the array `build_arr` returns: it is handed over and forgotten. Measured with `scripts/measure_heap.mjs` on the RC lane (a stage2 built from 56cdc5b plus the #2671 branch, 2026-09-11):

| shape | heap_used at N=20000 |
|---|---:|
| `let built = build_arr(3)` then `Array::length(built)` | 1,436 B (reclaimed) |
| `Array::length(build_arr(3))` | 1,681,436 B (84 B per call: the shell and its buffer) |
| `Array::length(Array::map(ta, incr_tree))` | 3,761,356 B (188 B per call: the result array and its retained elements) |

The same temporary in an OWNING position (a user function's parameter, a constructor field, a `let`) is released; only the borrowed-arg0 builtins (`md_is_borrow_arg0_call`: `Array::length` / `Array::get` / `Map::get` / ...) leave the temporary unowned. A leak, never a use-after-free: the value is dropped zero times, not twice.

The fix is on the codegen side of the borrow classification: when the argument of a borrow-arg0 builtin is itself an owning call (a user function or an owning builtin), park the result in a local, run the builtin, then `rc_drop` the local; or have the planner treat such an argument as a consume with a temporary drop.

Found while adding the #2671 shapes to `fixtures/rc_reclaim_leak_test.vibe`; the fixture binds every HOF result to a `let` so this gap does not mask its guard.

Triage: `bug` / P1 (unbounded leak on a common spelling; no wrong answer) / not a blocker.

mizchi:
Fixed in [#2690](https://github.com/mizchi/vibe-lang/pull/2690) (commit 7af52bd, extended by 7e043cb).

**Cause.** `Array::length(build_arr(3))` hands an owned temporary to a position that takes no ownership, and no binding holds it, so nobody dropped it: 84 B per call for the three-element array (the map result and its elements in the `Array::length(Array::map(ta, f))` shape).

[...]

Issue #2681, opened 2026-09-11
RC: an unannotated lambda parameter consumed three times is released early (drop of freed value)


mizchi:
## Repro

```vibe
enum Ty { TInt; TFn(Array[Ty], Ty) }
let consume_arr = (xs) -> { let holder = [xs]; Array::length(holder) }
let three = (tys) -> consume_arr(tys) + consume_arr(tys) + consume_arr(tys)
let main = () -> Int { let tys = [TInt, TFn([TInt], TInt), TInt]; three(tys) }
```

| lane | result |
|---|---|
| bump (`VIBE_RC=0`) | 3 |
| RC (`VIBE_RC=1`) | 3, by luck: the freed block is not reused in time |
| `VIBE_RC=shadow` | `drop of freed value at site 10; freed at site 8` |

Measured 2026-09-11 with a stage2 built from 56cdc5b plus the #2671 branch, and identically with the `entry-allows-2026-09-11_0aa2956` generation, so it predates that branch. A larger form (`Array::filter(tys, ..)` three times in one `let f = (tys) -> {..}`) traps `unreachable` on the plain RC lane.

## What decides it

The parameter is a heap value consumed at three owning positions. Each of these spellings is accounted correctly and runs clean under the shadow lane:

- two consumes of the unannotated parameter (`consume_arr(tys) + consume_arr(tys)`);
- three consumes of an ANNOTATED lambda parameter (`let three = (tys: Array[Ty]) -> Int { .. }`);
- three consumes of a `fn` parameter (`fn three(tys: Array[Ty]) -> Int`);
- three consumes of a local `let arr = [..]` binding.

Only the unannotated lambda parameter `(tys) -> ..` consumed three or more times is short one reference, whether the consumes are three `let`s or one expression. So the budget the planner gives an inferred-heap lambda parameter caps below its consume count, while the annotated and `fn` paths (and locals) budget from the real count.

Found while writing shape 10 of `fixtures/rc_shadow_regression_test.vibe` for #2671; that shape annotates its parameter to stay out of this bug's way and points here.

Triage: `bug` / P0 (the RC lane can answer from a freed block without any diagnostic; the shadow lane is the only thing that sees it) / not a blocker.

mizchi:
Fixed in [#2690](https://github.com/mizchi/vibe-lang/pull/2690) (commits d7f19f9, 8c34878, caf9df9, 2053dde).

[...]

PR #2690, opened 2026-09-11
RC lane: heap parameters of top-level lambdas (#2681), owned temporaries in borrowed positions (#2682), MapBuilder / Map reclamation (#2683), and in-place Map reuse (ADR-0092 for maps)

mizchi:
Three RC-lane accounting fixes, one commit each, the follow-ups the gate's fixtures and the unit battery forced, and — on top of the map accounting — the map counterpart of ADR-0092 reuse.

## #2681 — an unannotated parameter of a top-level lambda was planned as a scalar

Only the local `let` lambdas went through the planner's parameter heap inference, so `let three = (tys) -> consume(tys) + consume(tys) + consume(tys)` planned `tys` with no dups and no drops while every call consumed it: released three times for one reference (shadow: drop of freed value). Two changes in `perceus.vibe`: the call-site inference now runs for top-level lambdas too, and a parameter the body consumes at an owning position is marked heap whatever the calls pass — the case a scalar classification cannot afford to miss, and a scalar handed the heap treatment costs only guarded no-ops. `lambda_param_consume_rc_test` (seven shapes, bump / RC / RC-shadow agreement) is red on the committed planner.

**Follow-up 1 — cost.** The first cut seeded every top-level binding into the environment `analyze_calls` threads. That environment is an immutable `Map`, `Map::set` copies it, and the walk sets it at every `let` and every lambda parameter — so each walk copied the module's bindings again and the bump-built compiler ran out of memory compiling the split CLI (compiler gate 3cli/3: `memory access out of bounds` in `__rt_map_build_index` under `analyze_calls`). The top-level bindings are a side lookup now (`HeapInferCtx.top_env`, built once through a `MapBuilder`) that `arg_class` consults when the local environment has no entry, and a lambda's calls are collected only from the statements whose free-variable list mentions it. Measured on `lib/@vibe/cli/main.vibex` in RC output mode: the 045db6c compiler 21.6 s, the first cut a trap after 27 s, the fix 17.6 s.

**Follow-up 2 — a silent-wrong answer.** The gate's shadow fixture answered 5,025,372,489 instead of 25,377,489 on the RC lane once top-level lambdas were classified: its `lookup(names, effs, name)` returned `Some` for a name it did not hold. The planner spelled a heap-classified parameter's annotation as the empty tuple (`TyTuple([])`, a shape no source produces — `()` parses to `TyUnit`), and `desugar_trait_dicts`, which runs after `elaborate_heap_params`, seeded it as a composite eq shape: `Array::get(names, i) == name` became `eq_for_typed(TyTuple([]))`, trivially true. `linked_compile` read the same annotation as a tuple of arity 0. The marker is the nominal `__rc_heap` now: a name no type table knows is what an unannotated parameter already is to those readers, `type_expr_is_heap` recognises it, `seed_var_types` skips it by name, and `annotate_params_from_sig` lets a trait's declared type overwrite it (the unit battery's `fixtures/trait_method_generic_test.vibe` had lost its method-level `[X: Show]` witness to the marker). The `lookup` shape joins `lambda_param_consume_rc_test`; the shadow fixture is back to 25,377,489 on both the RC and the shadow lane.

## #2682 — an owned temporary handed to a borrowed position was never released

`Array::length(build_arr(3))` leaked 84 B per call. The resolved-call path now parks such a temporary (a call whose callee cannot return a view, or an array / tuple / record / map literal) in a local and releases it after the call; a callee that may return a view of its argument (`Array::get`, a user fn in the borrow_ret / view_ret sets) keeps the old behaviour, since its result aliases the temporary. `fixtures/rc_reclaim_leak_test.vibe` gains both shapes (5,441,372 B before the fix); `borrowed_temp_release_rc_test` pins that the release never frees a value still in use.

## #2683 — Map and MapBuilder are RC blocks

A `MapBuilder`, its storage, the `Map` it freezes and every `Map` that `Map::set` / `Map::delete` / a literal produce were raw bump allocations in both lanes — even values, invisible to the RC machinery — so a builder leaked 144 B per `new`, every value stored in it with it, and the frozen map on top. On the RC lane they are rc blocks now:

- a `Map` is class 6 (the `[count][index][entries]` layout rc_drop's arm already walked), allocated through `cc_map_alloc_rc` by `Map::set`, `Map::delete`, `MapBuilder::freeze` and the `EMap` literal (`Map::new()` and `Map::from_pairs` parse to it), and handed out tagged; every reader already untagged with `& -2`. The side index `__rt_map_build_index` builds from 8 entries is a class-0 leaf allocated through rc_alloc and freed by the class-6 arm;
- a `MapBuilder` is a class-9 handle over a class-0 storage leaf, and `rc_drop` gained the arm: drop the stored keys and values, release the storage, free the handle. `MapBuilder::set` releases the storage it grows out of and the value it replaces; `freeze` reads the handle's rc word — uniquely held, the entries move (count zeroed, handle and storage freed); shared, every copied key and value is retained;
- a key or value copied from one map into another is retained in both (`Map::set`, `Map::delete`, `Map::keys`, `Map::values`), a replaced key the caller retained for storage is released, and a literal's value that is a borrowed view (a projection, a borrow-returning call, a loop-borrowed name) is retained as the array / tuple / record literals do. A map's index slot is zeroed at allocation, since a block rc_alloc recycles is not.

**Follow-ups the reclaim fixture forced** (it kept 64 B per iteration against its bound): `Map::values`, `Map::size` and the `Map::has` alias were missing from `md_is_borrow_arg0_call`, so the planner handed them an owned reference no lowering dropped — invisible while maps were bump blocks, 48 B per map once they were rc blocks; and the inline lowerings that borrow their map (`Map::keys`, `Map::values`, `Map::size`, `Map::has_key`, `MapBuilder::set`, `MapBuilder::has_key`) now release an owned temporary handed there (`Map::set(Map::new(), k, v)`, 16 B per call), the #2682 rule applied by the arms themselves; `Map::get` / `MapBuilder::get` return a view and keep the leak on the safe side.

Pinned by `map_builder_rc_test` (nine shapes through bump / RC / RC-shadow) and the `MapBuilder` / `Map` shape of `fixtures/rc_reclaim_leak_test.vibe`: red on the 045db6c compiler with heap_used 36,321,540 B. The fixture's heap bound moves from 2,000 to 4,000 B — its steady state is a constant (3,204 B at N = 20000 and at N = 40000 before the reuse shapes, 3,568 B with them), while a leak scales with N. The bump lane keeps its raw allocations; the gc lane has its own Map lowering and is untouched.

## Map reuse — a uniquely held Map updates in place (ADR-0092 for maps)

`Map::set` and `Map::delete` **consume** their map now (they left `md_is_borrow_arg0_call`). On the RC lane the lowering reads the source's rc word: exactly one reference (`1 | 6 << 24`; a saturated block fails the test) and the entry is replaced, appended or removed in place and the same block is the result; anything else takes the copy path, after which the consumed reference is released. In place: the capacity comes off the block's size word, a full block moves its entries into one of twice the capacity, the side index is inserted into (`__rt_map_index_insert`, now on `CompileCtx`) while it stays at or under the half load `__rt_map_cap_for` sizes for and is released and rebuilt otherwise, and a delete shifts the tail down and rebuilds the index or drops it under 8 entries. The bump lane keeps the copy.

The planner already keeps a source that is read after — or through a view around — a consuming use shared (`pctx_apply_borrow_retention`; measured on four array shapes before the change), so `let v = Map::get(m, k); let m2 = Map::set(m, k, x)` copies and `v` survives. What it did not do was let an **accumulator** reach the unique path: `m = Map::set(m, k, v)` was retained whenever `m` had any other use, the callee copied, and the assignment lowering — which drops nothing for a self-referential right-hand side — left the old map behind: every intermediate value of a `let mut` accumulator leaked, any heap type, since #699 (864 B per iteration for nine sets on a map, 420 B for four `Array::concat`). `x = f(.., x, ..)` with `x` at exactly one owning position and nowhere else in the right-hand side now **moves** that occurrence (`pe_self_reassign_moves` / `pe_use`, the binding's remaining count topped back up by one so its scope end still releases what it holds last), the assignment lowering keeps its loop-carried dup off the same occurrence (`md_self_reassign_move_arg`), and the move is withheld when another binding still views the value (`view_src`).

Measured with viberun (RC lane, 20,000 iterations), before → after:

| shape | fuel | heap peak |
|---|---:|---:|
| a 21-step unique chain of `Map::set` through a consuming helper | 2,022,100,188 → 499,280,141 (−75%) | 9,504 → 2,480 B |
| nine `m = Map::set(m, ..)` on a `let mut` inside the loop | 234,520,377 → 149,420,287 (−36%) | 17,280,760 → 840 B |
| two `m = Map::set(m, ..)` on a `let mut` inside the loop | 19,420,266 → 20,980,273 (+8%) | 960,496 → 544 B |
| four `xs = Array::concat(xs, [..])` on a `let mut` | 68,880,301 → 62,480,229 (−9%) | 8,400,532 → 700 B |

Pinned by `map_reuse_rc_test` (thirteen shapes through bump / RC / RC-shadow: unique and shared sources, growth past the index threshold, a replace through the index, deletes back under it, a view read before and one held across a consuming set, a source consumed twice, accumulators inside and outside loops, reads between reassignments, the array accumulator) and by the accumulator shapes of `fixtures/rc_reclaim_leak_test.vibe` (`2i + 23` per iteration, total 7,801,510,000): the compiler before the move keeps 28,723,336 B on it against the 4,000-byte bound.

## Cost on the compiler's own RC-lane self-compile

Seed entry compiled by an RC-built stage2 of the PR base and of the tree before the reuse (viberun fuel / allocation, three runs each, identical to the digit):

| | base (bca14ae) | before reuse (2218ca5) | delta |
|---|---:|---:|---:|
| fuel | 1,177,029,188,611 | 1,194,445,790,779 | +1.48% |
| bytes allocated | 1,835,120,428 | 1,839,410,132 | +0.23% |
| heap peak | 1,835,229,036 | 1,839,518,828 | +0.23% |
| RC stage2 size | 4,889,148 | 4,895,557 | +0.13% |

The fuel is maps being accounted at all; the peak does not fall because the compiler's own maps live to the end of the compile. The same measurement with the reuse follows in a comment when its runs finish.

## Validation

Compiler gate green at 2053dde; stage2 == stage3 fixpoint at every step; full unit battery 1193/1193 at 2053dde; book doctest 130 blocks, 0 fail; the RC e2e files through bump / RC / RC-shadow (45 tests with the reuse shapes). The gate, battery and doctest at the reuse commits are running and will be reported in a comment.

https://claude.ai/code/session_01ASfkaC6HgSCSydTZcKSF2M


mizchi:
Validation of the map reuse commits (7f39a7d, 8dd24e4, 94b7dfe):

- compiler gate (`scripts/compiler_gate.sh`) green from a fresh worktree at 7f39a7d (18 min) and at 8dd24e4 with the extended reclaim fixture (14 min): reclamation guard 3,568 B at N = 20000, shadow guard 25,377,489, 0 failures;
- stage2 == stage3 fixpoint at 7f39a7d;
- full unit battery on that stage2: 1194/1194 files;
- book doctest on it: 130 blocks, 0 fail;
- the RC e2e files through bump / RC / RC-shadow: 45 tests.

RC-lane self-compile KPI (seed entry compiled by an RC-built stage2; viberun fuel / allocation; two runs each, identical to the digit):

| | base (bca14ae) | before reuse (3c7ba5b) | with reuse (7f39a7d) |
|---|---:|---:|---:|
| fuel | 1,177,029,188,611 | 1,194,445,790,779 (+1.48%) | 1,194,859,964,204 (+1.51%) |
| bytes allocated | 1,835,120,428 | 1,839,410,132 (+0.23%) | 1,841,373,196 (+0.34%) |
| heap peak | 1,835,229,036 | 1,839,518,828 (+0.23%) | 1,841,481,940 (+0.34%) |
| RC stage2 size | 4,889,148 | 4,895,557 (+0.13%) | 4,906,173 (+0.35%) |

The reuse is neutral on the compiler's own self-compile (+0.03% fuel over the tree before it) because its map traffic is not the shape the reuse serves: of 117 `Map::set` sites, 17 are self-reassignments (the planner's `env = Map::set(env, n, ..)` threading), 7 are `let`-bound, and the rest pass `Map::set(scope, name, id)` straight into a recursive call while the caller keeps `scope` — shared by construction, so they copy. The shapes it does serve (a unique chain −75% fuel, the `let mut` accumulator −36% fuel and its 17 MB leak gone, the array accumulator −9% and 8 MB gone) are in the PR body and `docs/perceus-reuse.md`.

---

Issue #2759, opened 2026-09-13
Adding one host import means editing six lists; the four the gc backend owns are the ones nothing checks


mizchi:
## Corrected 2026-09-13 — the premise I filed this on was wrong

I filed this saying *"only the contract JSON is machine-checked"* and that the cheapest fix was *"extending `scripts/check_host_runtime_contract.py` to assert BOTH runners register exactly the contract's set — that alone would have caught this PR's viberun miss."*

It already asserts that. Measured against a worktree at `63017d1` — PR #2756's **first** commit, before either Codex finding was fixed:

```
$ python3 scripts/check_host_runtime_contract.py
host-runtime-contract: emitter signature drift: missing=[] extra=['fs_remove_file'] changed=[]
exit 1
```

and then, applying the only fix for that message (classify the import in `portableCore` + `importTypes`) and leaving the Rust runner untouched:

```
host-runtime-contract: viberun lacks portable imports: ['fs_remove_file']
exit 1
```

So the existing checker catches **both** of Codex's P1 findings, in sequence, from the first commit. The tool was sufficient. I did not run it.

## What is actually unchecked

`check_host_runtime_contract.py` reads exactly four paths:

| list | read by the checker? |
|---|---|
| `lib/@vibe/compiler/codegen/wasi/linked_compile.vibe` — the **linear** emitter | yes (`EMITTER`) |
| `docs/wasm/host-runtime-contract.json` | yes (`MANIFEST`) |
| `runtime/viberun/src/main.rs` | yes (`RUST`, via `portableCore <= rust`) |
| `scripts/wasm_vibe_host_runner.js` | yes (`NODE`, via `portableCore <= node`) |
| `codegen/gc/backend_body.vibe` — `use_host` | **no** |
| … `host_defs` | **no** |
| … `host_imports` | **no** |
| … `hbo` (the base offset every generated body index is relative to) | **no** |

The four with no coverage are exactly the four the gc backend's own comment warns about: *"FOUR places encode this list … moving one without the others is silently wrong."* `hbo` is the sharpest — it is an offset, so getting it wrong does not produce a missing-import error, it shifts every call index in every generated body.

## So the fix is two things, neither of which is the one I first proposed

[...]

PR #2768, opened 2026-09-13
#2765 follow-up: check host_defs arity/return against its import's ABI signature

mizchi:
Follow-up to #2765. Codex posted a third P2 on `175d99c` at 14:55:52; the PR merged at 14:58:29, three minutes later, so the finding never reached it. It is real — reproduced below — and it lands as its own commit here.

## The finding

`175d99c` started keeping the *import's* ABI type id but still discarded `host_defs`'s `params` and `ret`. Reproduced against the merged checker:

```
("Fs::exists", 1, 5, 1) -> ("Fs::exists", 0, 5, 1)   arity 1 -> 0   PASSES
("Fs::exists", 1, 5, 1) -> ("Fs::exists", 1, 5, 0)   ret 1 -> 0     PASSES
```

while the paired import stays `("fs_exists", 3)` = `(i64) -> i64`. Those two fields populate `fn_param_counts` and `fn_returns_list` independently of the import's type, so either can drift and the generated call is emitted against the wrong shape.

## I argued against this check one round earlier, and was wrong

Replying on the type-id thread I wrote that comparing against `host_defs`'s arity would be *"a re-derivation of what the manifest already states"*. That conflated two different things: the **import's type id**, which the manifest does state and which #2765 now checks against it, and **`host_defs`'s own arity fields**, which the manifest does not state and which nothing derived. They are separate data and can disagree. Worth recording because the reasoning was confident and wrong, not careless.

## The fix adds no table

`coreTypeSignatures` is already in the manifest:

```json
{ "1": "(i64) -> ()", "3": "(i64) -> i64", "4": "(i64, i64) -> i64", "5": "() -> i64", "6": "(i64, i64) -> ()" }
```

`core_signature_shape` parses those to `(params, ret)` rather than transcribing them, and each `host_defs` entry is checked against its paired import's signature. Measured before asserting: all 22 already agree.

## The sweep repeated its lesson a third time

The new arity check **subsumed** the import-type-id case. No two `coreTypeSignatures` entries share a `(params, ret)` shape, so any backend-side type-id mutation trips arity too, and `test_import_abi_type_change_fails` stopped isolating the assertion it was named for — caught by the same disable-each-assertion sweep that caught the `hbo` case in #2759 and the count/membership cases in `175d99c`.

The two assertions guard different **pairings**: type-id is the backend against the manifest's `importTypes`; arity is the backend's two lists against `coreTypeSignatures`. So that test now mutates the **manifest** side — the backend keeps type 3 and its matching shape, and only the contract disagrees. That is the only mutation which isolates it.

Final sweep, nine assertions, each disabled in turn:

| assertion disabled | test that failed |
|---|---|
| `use_host`/`host_defs` drift | `test_use_host_losing_a_builtin_fails` |
| `host_defs` index order | `test_host_defs_index_out_of_order_fails` |
| `host_imports` count | `test_host_imports_gaining_an_entry_fails` |
| `host_imports` positional name | `test_same_abi_type_import_reorder_fails` |
| `hbo == len(host_defs)` | `test_hbo_desynced_from_host_defs_fails` |
| `header == hbo + 1` | `test_import_vector_header_off_by_one_fails` |
| `host_imports` in manifest | `test_import_absent_from_the_manifest_fails_even_when_self_consistent` |
| import ABI type id | `test_import_abi_type_disagreeing_with_the_manifest_fails` |
| `host_defs` arity/ret vs signature | `test_host_def_arity_…` + `test_host_def_return_…` |

No assertion is uncovered. (The last covers two fields, hence two tests.)

Green: `bash scripts/compiler_gate.sh` → `[compiler-gate] ok`, with `host-runtime ABI contract ok` inside it. All five meta-gates ok. `pkf run check-host-runtime-contract` ok (58 static imports; 48 portable; 22 gc host imports) + 23 tests.

https://claude.ai/code/session_01CCUq2wJKeDzguGzmx78HkZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01CCUq2wJKeDzguGzmx78HkZ)_


mizchi:
Added `e940057`: a self-audit of this checker, not a review finding.

Three review rounds each found a real gap in `validate_gc_lists`, so rather than wait for a fourth I went looking myself. Two candidates, one real.

**Real.** The ABI type check read `declared is not None and declared != type_id`, so an import the manifest's `importTypes` did not carry was silently unchecked. Band membership does not imply an `importTypes` row — they are separate keys — so this is reachable for any gc import the linear emitter does not also emit. Now fails closed.

**Not real, deliberately not asserted.** A duplicate name inside `use_host`. That list is a boolean OR chain, so `A || A` is `A` and a duplicate changes nothing. Asserting on it would be noise dressed as rigour, and I would rather say why I left it than have it look overlooked.

The interesting part is what happened next. Fixing the first initially added a separate `declared is None` branch, and the sweep then reported **two** assertions as uncovered — both reports correct:

- **The `None` branch cannot be isolated by any mutation.** The comparison after it already fails closed on `None`, so disabling the branch changes nothing. It read like a second check and was only a second *message*. Collapsed to one assertion whose message names which case fired. An assertion no mutation can distinguish is a branch, not a guarantee.
- **Band membership became uncoverable** once the type check failed closed — the three-way-rename mutation was caught by the missing `importTypes` row instead. Its test now also *gives* the invented name a matching `importTypes` row, leaving the backend fully self-consistent and typed, so band membership is the only thing left that can see it.

Final sweep, nine assertions, each disabled in turn:

| assertion disabled | test(s) that failed |
|---|---|
| `use_host`/`host_defs` drift | `test_use_host_losing_a_builtin_fails` |
| `host_defs` index order | `test_host_defs_index_out_of_order_fails` |
| `host_imports` count | `test_host_imports_gaining_an_entry_fails` |
| `host_imports` positional name | `test_same_abi_type_import_reorder_fails` |
| `hbo == len(host_defs)` | `test_hbo_desynced_from_host_defs_fails` |
| `header == hbo + 1` | `test_import_vector_header_off_by_one_fails` |
| `host_imports` in manifest bands | `test_import_absent_from_the_manifest_bands_fails` |
| import ABI type vs `importTypes` | `test_import_abi_type_disagreeing…` + `test_import_missing_from_importTypes_fails_closed` |
| `host_defs` arity/ret vs signature | `test_host_def_arity…` + `test_host_def_return…` |

None uncovered, and none is a branch no mutation can distinguish.

`bash scripts/compiler_gate.sh` → `[compiler-gate] ok` on this exact tree, with `host-runtime ABI contract ok` inside it. All five meta-gates ok. 24 tests.

---

Issue #2792, opened 2026-09-14
RC lane: `Array::set: index N out of bounds for length N` compiling an owning wrapper assignment followed by another `let`


mizchi:
## Repro

```vibe
fn f() -> Int {
  let mut x = [1, 2]
  x = {
    let ys = [[3, 4]]
    Array::get(ys, 0)
  }
  let r = Array::length(x)
  r
}

export fn main() -> Int {
  f()
}
```

```
VIBE_RC=1 ... --invoke cli_main <compiler> m5.vibe m5.wasm main
Array::set: index 12 out of bounds for length 12
```

The compiler aborts. No diagnostic, no source position — the message is an
internal `Array::set` bounds failure, so it says nothing about what to edit.

## Lane split

| lane | result |
|---|---|
| `VIBE_RC=1` (production default) | **abort**, `Array::set: index 12 out of bounds for length 12` |
| `VIBE_RC=0` (bump) | compiles, answers `2` (correct) |

## Not a regression

Reproduces on the **committed seed** (`bootstrap/seed/compiler.wasm`,
sha256 `00f5022935c7ada9d1abc713be5c4f34358dad2b6316e94c53f1f280579acb1b`)
and on a stage2 built from current `main`. It predates #2760 / #2783 / #2790;
those touched the neighbouring retain decision, not this.

## Trigger

Two ingredients, both required — drop either and it compiles:

1. an owning mutable slot assigned a **wrapper** whose binder **owns** its value
   and whose tail escapes an element of it (`let ys = [..]; Array::get(ys, 0)`), and
2. **another `let` afterwards in the enclosing scope**.

Measured on the seed:

| variant | result |
|---|---|
| the repro above | abort |
| same, with `let r =` removed (`Array::length(x)` as the tail) | compiles, `200` |
| same, but the binder is a borrow (`let ys = Array::get(xss, 0)`) | compiles |
| same, but no trailing statement after the assignment | compiles |

The enum/`Array` element type is irrelevant — the repro above uses no
user types at all.

## Where to look

The index is one past the end, and it grows with the number of locals the
function needs, which points at the scope-end **promotion** in the `ELet`
lowering (`lib/@vibe/compiler/codegen/expr/compile_expr_tail.vibe`, the
`ctx.rc_drop_func_idx >= 0 && scope_tail_proj_root(body) == name` arm of
`compile_expr_tail`). That arm is the one that takes an extra local:

```
emit_local_set(buf, nl2)
[...]

Issue #2760, opened 2026-09-13
RC: assigning a wrapper whose result leaf borrows into an owning mutable slot keeps the slot owning, and its drop frees the container's element


mizchi:
Found in review on #2750 and verified there to be **pre-existing on `main`**, so it is filed rather than folded into that PR. Same family as #2733's body — a wrapper hiding a borrow from a test that reads the value's top node — on the **assignment** side rather than the binding side.

## The hole

The assignment lowering in `codegen/expr/compile_expr_tail.vibe` decides whether a replacement value borrows by looking at its top node: a top-level `EDot`, or a direct `ECall` through `call_returns_borrow`. A wrapper — a block, i.e. an `ELet` / `ESeq` chain — is neither, so no retain is emitted for it, while the slot stays in `heap_binding_names`. Its next assignment or its scope-end drop then releases an element the container still owns.

## Repro

```vibe
enum Node { Leaf(Int); Pair(Node, Node) }

fn total(n: Node) -> Int {
  match n {
    Pair(a, b) => total(a) + total(b),
    Leaf(v) => v
  }
}

fn once(xs: Array[Node]) -> Int {
  let mut x = Pair(Leaf(1), Leaf(2))
  x = {
    let k = 0
    Array::get(xs, k)
  }
  let r = total(x)
  let filler = Pair(Leaf(7), Leaf(8))
  r * 100 + total(Array::get(xs, 0)) * 10 + total(filler) * 0
}

fn main() -> Int {
  let xs = [Pair(Leaf(3), Leaf(4))]
  once(xs)
}
```

| lane | result |
|---|---:|
| `VIBE_RC=0` (bump) | `770` |
| default / `VIBE_RC=1` | **`850`** |

`770` is correct: `total(x)` is 7 and the later `total(Array::get(xs, 0))` is 7, so `7 * 100 + 7 * 10`. The RC answer reads the second one as **15** — the filler, handed back out of the block that `x`'s drop released. No trap; RC is the production default, so this is the answer a plain `vibe run` gives.

Measured identically on `main`, on `401a0a1`, and on `a3eb2d9` (#2750's head), which is why it is not that PR's.

## Fix shape

The retain condition needs the value's RESULT SPINE, the same way #2750 gave it to the `let mut` initializer (`pc_count`'s ELetMut `borrowed_call` and the matching retain in `compile_expr_tail`, both reading a straight-line spine leaf).

Two cautions from #2750's rounds, both paid for there:

[...]

PR #2794, opened 2026-09-14
Fix the use-after-free #2790 put on main: the retain suppression must establish the retain, not the name

### LINKED ISSUE #2793: RC lane use-after-free: the #2790 retain suppression fires on a spine binder that never acquired an owned reference
**Fixed on `main`** by #2794 (`8e82e90`, merged as `702c134`).

---

**On `main` now** (`553244d`, merged as #2790). Silently wrong, no trap.

## Repro

```vibe
enum Node {
  Leaf(Int);
  Pair(Node, Node)
}

fn total(n: Node) -> Int {
  match n {
    Pair(a, b) => total(a) + total(b),
    Leaf(v) => v
  }
}

fn assign_borrow_bound_binder(xss: Array[Array[Node]]) -> Int {
  let mut x = Pair(Leaf(1), Leaf(2))
  x = {
    let ys = Array::get(xss, 0)
    Array::get(ys, 0)
  }
  let r = total(x)
  let filler = Pair(Leaf(7), Leaf(8))
  r * 100 + total(Array::get(Array::get(xss, 0), 0)) * 10 + total(filler) * 0
}

export fn main() -> Int {
  assign_borrow_bound_binder([[Pair(Leaf(3), Leaf(4))]])
}
```

| lane | answer |
|---|---:|
| bump (`VIBE_RC=0`) | 770 |
| RC (production default) | **850** |

`850` is the second read of the element answering with the filler, `15`, instead
of `7` — the owning slot's drop released a reference `xss` still owns, and the
next allocation reused the block. The same signature #2760 was filed for.

## Cause

`spine_leaf_unowned` in `lib/@vibe/compiler/codegen/expr/compile_expr_tail.vibe`
declined the assignment retain when `scope_tail_proj_root(value)` named a binder
the spine itself bound, on the reasoning that the inner `ELet` lowering had
already promoted the escaping element. That reasoning holds only when the binder
**owns**: `classify_let_value_heap` leaves a borrow-returning call at
`rc_kind == 0`, so `let ys = Array::get(xss, 0)` gets no scope-end drop and the
promotion guarded by it never runs. There was no earlier retain to duplicate,
and declining left the slot holding a reference nobody took.

Reported by Codex review on #2790 (P1 there), after that PR had merged.

## Attribution

| compiler | RC lane | bump lane |
|---|---:|---:|
| committed seed (pre-#2760) | 850 | 770 |
| `main` before the fix | 850 | 770 |
| `main` with #2794 | **770** | 770 |

The seed row is the point: the shape was already broken before #2760, #2760's
spine walk **fixed** it, and #2790's suppression partly undid that fix for every
shape whose borrow root is a spine binder. Not a new class — a regression of
#2760's own repair, trading a bounded leak for a wrong answer.

## Fix as landed

`spine_binder_promoted` mirrors the promotion's own guard instead of
approximating it:

```vibe
fn spine_binder_promoted(root: String, binders: Array[String], ctx: CompileCtx) -> Bool {
  array_contains_str(binders, root) && array_contains_str(ctx.heap_binding_names, root) && array_contains_str(ctx.rc_drop_names, root)
}
```

`ctx.heap_binding_names` is the record `classify_let_value_heap` writes on
exactly its `rc_kind = 2` paths, and both callers of `spine_leaf_unowned` run
after the value has been compiled. Reading those flat whole-function lists by
name is exact rather than approximate because normalize alpha-renames every
shadowing binder (#715).

## Both directions have a lane now

They are opposite errors and one fixture cannot hold both — which is why #2790's
gates were green while shipping this:

| direction | failure | lane |
|---|---|---|
| suppress too widely | frees a live reference — wrong answer | `assign_borrow_bound_binder` in `fixtures/rc_wrapper_assign_borrow_test.vibe` |
| suppress too narrowly | leaks 80 B per assignment | gate row 40e2 |

## Verification on the merged state

On a stage0→stage1→stage2→stage3 generation built from the fix:

| check | result |
|---|---|
| the P0 shape, both lanes | 770 / 770 |
| #2760's own repro, both lanes | 770 / 770 |
| gate row 40e2 leak bound | `heap_used=164 B` at N=20000 |
| `scripts/compiler_gate.sh` | **`[compiler-gate] ok`**, 0 failures |
| self-compile fixpoint | stage3 sha256 == stage2 |
| CI on #2794 | all 38 checks green |
| Codex review on `8e82e90` | no findings |

Red → Green both recorded: the fixture fails `actual: 850 / expected: 770` on a
stage2 built from the unfixed tree and passes 7/7 on the fixed one.

## Related

- #2760 — the retain this suppression sits on top of.
- #2792 — a pre-existing RC-lane compiler abort found while measuring this,
  still open; it is why the owning-binder counterpart is bounded by heap rather
  than asserted by value.

mizchi:
Fixes #2793. Found by Codex review on #2790 — **after that PR had already merged**, so the use-after-free is live on `main` (`553244d`) right now.

## What is broken on main

#2790 declined the assignment retain whenever the borrow's root was a name the **spine itself bound**, on the reasoning that the inner `ELet` lowering had already promoted the escaping element. That reasoning holds only when the binder **owns**. `classify_let_value_heap` leaves a borrow-returning call at `rc_kind == 0`, so

```vibe
x = {
  let ys = Array::get(xss, 0)
  Array::get(ys, 0)
}
```

binds `ys` in the spine exactly as the owning shape does and promotes nothing. There was no earlier retain to duplicate, and declining left the owning slot holding a reference nobody took — its drop then freed an element `xss` still owns.

Measured, one fixture function, RC lane against bump:

| compiler | RC lane (production default) | bump lane |
|---|---:|---:|
| committed seed (pre-#2760) | 850 | 770 |
| **`main` today** | **850** | 770 |
| this branch | 770 | 770 |

`850` is the second read of the element answering with the filler, `15`, instead of `7`. No trap, no diagnostic — the lanes just disagree.

`compile_expr_tail.vibe` is byte-identical between `db85678` and `main`, so the `850` measured on a stage2 built from `db85678` is main's answer for this path.

## This is a regression of #2760's own repair, not a new class

The seed row above is the point. The shape was already broken before #2760; #2760's spine walk **fixed** it; #2790's suppression partly undid that fix for every shape whose borrow root is a spine binder. It traded a bounded leak for a wrong answer, which is the worse trade by this project's priority order.

## The fix

`spine_binder_promoted` mirrors the promotion's **own** guard instead of approximating it:

```vibe
fn spine_binder_promoted(root: String, binders: Array[String], ctx: CompileCtx) -> Bool {
  array_contains_str(binders, root) && array_contains_str(ctx.heap_binding_names, root) && array_contains_str(ctx.rc_drop_names, root)
}
```

- `heap_binding_names` is the record `classify_let_value_heap` writes on exactly its `rc_kind = 2` paths, and both callers of `spine_leaf_unowned` run **after** the value has been compiled, so the record is already there to read. The same `heap_binding_names && rc_drop_names` pair is how this file already decides `is_heap_mut`.
- Reading those flat whole-function lists by name is exact rather than approximate because normalize alpha-renames every shadowing binder (#715): within one function a name is one binding.
- `reuse_fusion` is the promotion's remaining guard and is deliberately untested — it requires the binder's body to be an `EMatch`, and `scope_tail_proj_root` answers `""` for one, so a spine whose root is this binder never took that branch.

The four cases compose as they should: an empty root retains (unchanged); a non-binder root retains (#2760's own `xs` parameter repro); an owning binder declines (the leak this clause exists for); a borrow binder retains (the bug).

## Both directions need their own lane

They are opposite errors, which is why #2790's gates were green while shipping this — the leak side had a bound and the value side had nothing:

| direction | failure | pinned by |
|---|---|---|
| suppress too widely | frees a live reference — a wrong answer | `assign_borrow_bound_binder` in `fixtures/rc_wrapper_assign_borrow_test.vibe` |
| suppress too narrowly | leaks 80 B per assignment | gate row **40e2**, `fixtures/rc_wrapper_assign_double_retain_test.vibe` |

**Red first:** the new case fails `actual: 850 / expected: 770` on a stage2 built from the unfixed tree.

## Known gap, deliberately not touched

The sibling `EDot` arm declines on the same name-only test. It is a **gap, not a regression** — `main` emits nothing at all for a wrapped projection, so declining there keeps main's answer. Closing it needs the ownership test plus its own measurement of the drop ordering that arm's comment appeals to, so it is left for #2760's follow-up rather than widened into this PR.

## Also filed while measuring this

**#2792** — on the RC lane the compiler aborts with `Array::set: index N out of bounds for length N` when an owning wrapper assignment is followed by another `let` in the same scope. Reproduces on the committed seed, so it predates all of this. It is why the owning-binder counterpart is bounded by heap rather than asserted by value.

## Validation

A stage2 is building from this tree and the local gate results will be posted here when it lands. What is established so far:

| check | result |
|---|---|
| the P0 shape on a stage2 built from the unfixed tree | 850 (RC) vs 770 (bump) — reproduces |
| the same shape on the committed seed | 850 — predates #2760 |
| Red test on the new fixture case | fails `actual: 850 / expected: 770` |
| `scripts/vibe_fmt.sh --check` on both changed files | ok |
| `pkf run pre-commit` (review-derived lint gates) | passed |
| compiler sources compile (stage1 built from an equivalent tree) | ok |

Do not merge before the gate results are posted — this PR exists because the previous one merged early.

---


mizchi:
## Validation — the RC-sensitive checks, on a stage2 built from this tree

A fresh stage0→stage1→stage2→stage3 generation from this branch (`8e82e90`, base `553244d`).

| check | result |
|---|---|
| the P0 shape, RC lane | **770** (was 850) |
| the P0 shape, bump lane | 770 — lanes agree |
| #2760's own repro, both lanes | 770 / 770 — not regressed |
| gate row 40e2 heap bound (leak direction) | **`heap_used=164 B`** at N=20000, `result=0` |
| `fixtures/rc_wrapper_assign_borrow_test.vibe` | **7/7** |
| self-compile fixpoint | `stage3` sha256 **==** `stage2` |

### Red → Green, both recorded

The same fixture, same command, two compilers:

```
# unfixed tree (= main's codegen for compile_expr_tail.vibe)
failing test: 2760 a wrapper binder that is itself a borrow retains nothing to reuse
  actual:   850
  expected: 770
[vibe-test] 0 passed, 1 failed (1 files, 7 tests)

# this tree
ok   fixtures/rc_wrapper_assign_borrow_test.vibe
[vibe-test] 1 passed, 0 failed (1 files, 7 tests)
```

### The leak bound is the one that could have gone the other way

`spine_binder_promoted` adds two conjuncts (`heap_binding_names`,
`rc_drop_names`) to a test that previously only asked for the name, so it can
only **narrow** the suppression — and narrowing it is exactly what re-opens the
1.6 MB double-retain leak that gate row 40e2 exists to catch. It measures
**164 B**, unchanged from before this branch, so the owning-binder shape still
takes the suppression it needs. That is the check I would have expected to
break, and it did not.

The full `scripts/compiler_gate.sh` and the unit suite are running against this
same stage2; I'll post those here when they land.

---

mizchi:
## Full gate, on the same stage2 — after the merge

```
[compiler-gate] gate self-tests ok (#2248)
[compiler-gate] ok
```

`scripts/compiler_gate.sh` with `VIBE_STAGE2_WASM` pointing at the stage2 built
from this branch: **exit 0, 0 failures**, including the
`checked-module-parity` sweep (632 modules), `rc-entry-parity`, the
`selfbuild seed→stage1→stage2→stage3` row and the gate self-tests.

Combined with what was posted before the merge, the complete record:

| check | result |
|---|---|
| the P0 shape, both lanes | 770 / 770 |
| #2760's own repro, both lanes | 770 / 770 |
| gate row 40e2 leak bound | `heap_used=164 B` at N=20000 |
| `fixtures/rc_wrapper_assign_borrow_test.vibe` | 7/7 |
| `scripts/compiler_gate.sh` | `[compiler-gate] ok`, 0 failures |
| self-compile fixpoint | stage3 sha256 == stage2 |
| CI | all 38 checks green |
| Codex review on `8e82e90` | no findings |

Confirmed on `main` at `702c134`: `spine_binder_promoted` is present and the
call site reads `!spine_binder_promoted(tail_root, binders, ctx)`.

#2793 is closed. #2792 stays open — the pre-existing RC-lane compiler abort
(`Array::set: index N out of bounds for length N`) found while measuring this.

---

mizchi:
Unit suite, closing the record:

```
[unit-test-runner] using prebuilt stage2: _build/_p1gen_fixed/stage2.wasm
ok:   fixtures/rc_wrapper_assign_borrow_test.vibe
ok:   fixtures/rc_wrapper_assign_double_retain_test.vibe
[unit-test-runner] 1269/1269 active unit-test files passed
[unit-test-runner] ok
```

**1269/1269**, 0 failures, on the same stage2 the gate ran against — both lanes
of this contract green: the value fixture (suppress-too-widely → wrong answer)
and the heap-bounded one (suppress-too-narrowly → leak).

That completes the validation table for this PR. Nothing outstanding on it.

---

Issue #2592, opened 2026-09-09
Wiring enforcement does not cover the 27 `*_gate.sh` scripts


mizchi:
Follow-up to #2580 / #2591. `check_gate_wiring.sh` enforces that every gate is reachable from a workflow, but its corpus is `scripts/check_*.sh` + `scripts/lint_*.sh` — which is what #2580 defined. Raised by Codex on #2591; scoped out there deliberately rather than widened mid-change, and the gate now states the boundary in its own header.

## What is not covered

27 scripts, measured on main:

```
compiler_gate.sh   minify_gate.sh
test_async_component_gate.sh              test_async_sleep_component_gate.sh
test_async_string_lift_probe_gate.sh      test_concurrent_awaits_component_gate.sh
test_future_value_component_gate.sh       test_host_future_value_component_gate.sh
test_host_stream_value_probe_gate.sh      test_hostfuture_source_component_gate.sh
test_http_body_read_probe_gate.sh         test_http_body_stream_probe_gate.sh
test_interleaved_tasks_probe_gate.sh      test_named_hostfutures_component_gate.sh
test_named_hoststreams_component_gate.sh  test_parallel_warm_pool_gate.sh
test_serve_async_lift_gate.sh             test_serve_body_stream_gate.sh
test_spawned_future_component_gate.sh     test_stream_value_component_gate.sh
test_wasi_cli_stdin_p3_probe_gate.sh      test_wasi_cli_stdin_provider_component_gate.sh
test_wasi_cli_stdin_provider_guest_component_gate.sh
test_wasi_cli_stdin_provider_source_component_gate.sh
test_wasi_http_p3_full_gate.sh            test_wasi_p3_guarantee_gate.sh
test_wit_async_import_component_gate.sh
```

An unwired one there is invisible in exactly the way #2580 exists to fix, and it does not even need an allowlist row to stay that way.

## Why it is worth doing, with evidence

The `check_*` / `lint_*` half found a **fourth** dark gate the moment it was enforced correctly: `check_book_console.sh` ran nowhere (only its self-test was wired), and it was **red** — the launcher had grown an `at off=62` line the book chapter did not carry. That is the same shape as `lint_tracked_experiment_names.sh`, `check_task_inputs.sh` and `check_portable_boundary.sh` before it. There is no reason to expect the component-gate family to be cleaner, and the `test_*_component_gate.sh` scripts are the ones exercising WASI p3 / async surfaces where a silent gap is expensive.

[...]

PR #2847, opened 2026-09-15
Keep the withheld-capability name on viberun's bench path (#2844 follow-up)

mizchi:
Follow-up to #2844, which merged before its last two review items landed.

## 1. The bench path loses the refusal (Codex P2, now a defect in `main`)

#2844 taught `run()` to print which capability was withheld. viberun renders **bench** errors separately, and that path drops it: `bail!("bench ...: {e}")` formats Display — only wasmtime's outer `error while executing at wasm backtrace` — and then builds a *new* error whose chain no longer carries the cause at all. So `--bench` exited with generic trap text and no capability name, which is a diagnostic that does not name the edit.

One helper now answers "does this chain carry a refusal", used by `run()` and by both bench sites. Measured with no debug variables set:

```
viberun: bench `...::read_one`: trap during warmup: error while executing at wasm backtrace:
    0:    0x491 - <unknown>!__bench_read_one (vibe capability withheld: fs_read_file)
```

The gate gains a bench case. **The red case is worth recording because my first attempt at it passed**: viberun has two bench error sites, per-block and whole-module, and I mutated the one a `bench` block never reaches. Both are fixed; the case pins the per-block one, and the script says so rather than implying both are covered.

## 2. The input list — and what measuring it actually showed

`ensure_viberun.sh`'s `src_hash` hashes every file under `runtime/viberun/src` plus `Cargo.toml` and `Cargo.lock`; the task named only `src/main.rs`, a strict subset. Now mirrored, so the two cannot disagree about which binary a cached verdict stands for.

But **the replay that finding describes does not currently happen**, and the task now says so. Two consecutive runs with *nothing changed*:

```
run A (populate)        -> pkf: 2 task(s) · 2 ran · 5m43s
run B (nothing changed) -> pkf: 2 task(s) · 2 ran · 5m57s
```

Never cached — `selfhostGeneration` is `cache = false`, and a dependent of an always-running task runs too. So the list describes what *would* decide the answer rather than closing a live hole. Spelled out anyway: the day that dep goes away, an input list that was a subset of `src_hash` becomes a wrong answer with nothing to say so.

## Verification

`host_capability_withhold_test.sh` green under `env -u RUST_BACKTRACE -u VIBE_RUNNER_BACKTRACE` (both runners, plus the new bench case and its red case) · `check-gate-portability` ok · `check-task-inputs` ok (29 compiler-touching tasks) · `pkf format --check Taskfile.pkl` rc=0. No compiler sources touched.

https://claude.ai/code/session_012pM2xx7LYgDYJDKVqD75PY


mizchi:
## `wasi-p3-gate (wasmtime v47.0.2)` is red here, and it is not this PR's

```
named hoststreams component gate FAILED:
  _build/bench/selfhost_named_hoststreams_component/stream_for_proj.vibe
  did not compile: unknown name: HostStream::equals
```

This PR's diff is `Taskfile.pkl`, `runtime/viberun/src/main.rs` and one shell script. Nothing in it touches `HostStream`, the component gates, or name resolution.

**Red on the base branch too**, identically — same job, same message, three consecutive merges:

| main | run | `wasi-p3-gate` |
|---|---|---|
| `0cb7e97` (#2842) | 34978044979 | ✅ |
| `8c966c3a` (#2843) | 34981318089 | ✅ |
| `1bb59a4` (#2846) | 34983481297 | ❌ `unknown name: HostStream::equals` |
| `642069e` (#2838) | 34983989075 | ❌ |
| `d92802a` (#2844) | 34984374389 | ❌ |

**It is not a regression — it is a pre-existing failure that just became visible.** `1bb59a4` is #2846 (#2592, "require a workflow path for `*_gate.sh` scripts"), which wired twelve previously-unwired async gates into this CI entry point, `test_named_hoststreams_component_gate.sh` among them. Its own diff says why: *"Each had a pkf task and no workflow path, so an unwired one was invisible."* That gate has presumably been failing for as long as `HostStream::equals` has been missing; nothing ran it.

So #2592 did exactly what it set out to do, and the first thing it found is a real one.

**No fix exists to port.** The failure is in the gate's own fixture against a missing name, which is a compiler/stdlib question well outside a Taskfile-and-diagnostic PR, and it blocks every PR against `main` equally. Filed separately rather than widened into this one. No re-run spent: the base branch reproduces it deterministically, which is stronger evidence than a re-run would be.

---

Issue #2848, opened 2026-09-15
A field read on a type FORMAL receiver resolves by spelling against an unrelated struct


mizchi:
Found while fixing #2820. It is #2807's defect one rung over, and it is the rung #2807 left tolerated on purpose — so this is not a missed entry in a list, it is the case that needs its own answer.

## Measured

On a stage2 built from `8c966c3`, and identically on the #2820 branch (that change does not touch this path):

```vibe
struct Crate {
  payload: Int
}

fn read[T](m: T) -> Int {
  m.payload
}

fn main() -> Int {
  read(7)
}
```

Compiles clean — `vibe check` is empty, exit 0 — and **answers `0`**. That is slot 0 of a tagged integer, read through `Crate`'s field offset. There is no `Crate` anywhere near the call; the only thing connecting them is that `Crate` is the struct in the program that happens to declare a field spelled `payload`.

This is the exact shape #2807 measured for an enum receiver (`e.payload` answering `7` through `Box`'s offset), including the part that makes it P0 rather than P1: the program produces a number, and nothing says the number is meaningless.

The second failure mode is the #2820 one. With **no** struct declaring the name:

```vibe
fn read[T](m: T) -> Int {
  m.v
}

fn main() -> Int {
  read(3)
}
```

is clean in `vibe check` and dies in the build with an unlocated `unknown struct field: v`. So the formal receiver has both halves: silently wrong when some struct happens to declare the name, check-clean/build-fails otherwise.

## Why it is not a one-line addition

The arm that tolerates it (`lib/@vibe/compiler/checker/checker.vibe`, the `EDot` field-type match) tolerates **three different things at once**, and only one of them is this bug:

- an unresolved type formal (`CtNamed("T", [])`) — always wrong today, since there is no bound that can promise `T` has a field, and no monomorphization pass to specialize the body per instantiation;
- an unresolved **cross-module** name on the incremental per-file lane, which resolves later;
- a kinded / applied formal shape.

Refusing the arm wholesale false-positives on the second. Telling the first from the second is the work, and it is the same shape of problem as #2737 / #2745: what a generic body is allowed to do with a value of its own formal, when the formal's bound promises nothing.

Two directions, neither obviously right:

1. **Refuse a field read whose receiver type is a formal in scope** — the checker would have to know, at this arm, that the head is a binder's formal rather than an unresolved import. `dtd_scope_formals` carries that fact in the desugar pass; the checker does not have an equivalent here.
[...]

Issue #2807, opened 2026-09-15
Field access on an enum value resolves against an unrelated struct's field and answers silently


mizchi:
## What

`e.f` where `e` is an **enum** value and `f` is a field of some **unrelated struct** in the program is accepted by both `vibe check` and the build, and answers by reading the enum's representation at that struct's field offset. No diagnostic, a plausible-looking number.

Measured 2026-09-15 against a stage2 built from main at `f8d0b24` (and reproduced identically on a later branch build, so it is not branch-local).

```vibe
struct Box {
  payload: Int
}

enum E {
  A;
  B(Int)
}

export fn main() -> Int {
  let e = B(7)
  e.payload
}
```

| lane | result |
|---|---|
| `VIBE_CHECK_ONLY=1` | clean |
| `VIBE_FS_COMPILE=1` build | compiles |
| running it | **`7`** |

`7` is the enum payload read through `Box::payload`'s offset. It looks right, which is the worst property it could have.

## It is the offset, not the value

Give the struct a second field and read that one instead:

```vibe
struct Box {
  first: Int;
  second: Int
}

enum E {
  A;
  B(Int)
}

export fn main() -> Int {
  let e = B(7)
  e.second
}
```

→ compiles, answers **`0`**.

And the field's declared TYPE is taken on faith, which is the sharpest form — a `String` field over an `Int` payload:

```vibe
struct Named {
  label: String
}

enum E {
  A;
  B(Int)
}

export fn main() -> Int {
  let e = B(7)
  String::length(e.label)
}
```

→ compiles, answers **`14`**. The tagged integer is read as a `(ptr|len)` string and a length is reported for memory that is not a string.

## The guard exists but only fires when no struct claims the name

With no struct in the program declaring the name, the build DOES refuse:

```vibe
enum E {
  A;
  B(Int)
}

export fn main() -> Int {
  let e = B(7)
  e.no_such_field
}
```

| lane | result |
|---|---|
| `VIBE_CHECK_ONLY=1` | **clean** |
| `VIBE_FS_COMPILE=1` build | refused: `unknown struct field: no_such_field` |

So there are two separate defects here, and the second is the reason the first is reachable:

1. **The field name is resolved without checking that the receiver is a struct at all.** A field belonging to any struct in the program is accepted on an enum receiver, and codegen emits the load. This is the silently-wrong half.
[...]

Issue #2475, opened 2026-09-03
Untyped-empty array pushes under a source-owned `struct Int` cannot tell a literal from the struct


mizchi:
**Body restated 2026-09-10.** The symptom and the fix direction are unchanged; what is new is that the fix direction has now been attempted five times and each attempt failed differently, which changes what the first step has to be. The comments below hold the measurements.

## Symptom

A program may declare a struct whose spelling is a scalar's (`struct Int { value: String }`; #2456 round 17 made a source-owned spelling take its declared comparator). The #2157 push scan names a pushed Int literal `Int` — the builtin — but the spelling is the struct's, and a struct literal push spells the same `Int`. The shape channel cannot tell the two apart.

```vibe
struct Int {
  value: String
} derive (Eq)

test {
  let left = []
  Array::push(left, 1)
  let right = []
  Array::push(right, 2)
  inspect(left == right, content = "false")
}
```

#2471 (Codex round 4) made the array rung — and the #825 static-shape rung and the bare `Option` / `Result` rung's row use — fail closed when a shape mentions a scalar spelling the source owns (`ty_mentions_owned_scalar`). #2475 moved those refusals from a runtime trap to a compile error naming the edit. Today all five `structural_eq_owned_scalar_*_refused.vibe` fixtures refuse on both lanes, and the same programs with the ownership removed answer correctly (`structural_eq_owned_scalar_control_test.vibe`, #2627).

## Fix direction

The spelling cannot carry the distinction; the shape channel needs a representation that does — a builtin-scalar marker the scan and the literal inference write for a literal (unforgeable, like `?EqUnknown` after #2471), with `eq_for_typed` giving it the builtin arms.

## What five attempts established

`!` is a sound marker character (no source type name can contain one) and `type_key` is length-prefixed, so `!Int` and `Int` key distinctly. The problem is not the marker's design; it is that **this channel has more consumers of a leaf spelling than producers of one, and an unaudited consumer does not fall back to the old behaviour — it does something new.**

| attempt | what was marked | result |
|---|---|---|
| 1 | `dtd_scalar_shape`, `self_describing_fallback`, `infer_static_eq_type` | FS lane: `literal_push` compiled and answered **`true`** — silently wrong |
| 2 | `dtd_scalar_shape` only | no fixture moved |
| 3 | as 1, measured on the flat lane | three fixtures moved from a clean refusal to **`field access .value on a non-struct value`** |
[...]

Issue #2263, opened 2026-08-24
Decide Map's key story for 0.1.0: the builtin Map is String-keyed while its type syntax reads Map[K, V]


mizchi:
Freeze decision, raised while pinning the 0.1.0 stdlib scope: stable-surface §3 freezes `Map` (`get` / `has_key` / `keys` / `values` / `set` / `size`), and §2.1 lists the type as `Map[K, V]` — but the builtin only works String-keyed. The syntax promises a generality the library does not deliver, and freezing that mismatch bakes it in (owner: this is explicitly not the intuition we want to ship).

## What exists today (measured 2026-08-24)

- Builtin `Map` ops in the registry are `Map[String, V]`.
- `@vibe/core` already carries the generic-key bridge: `trait Hash { hash_key(Self) -> String }` with impls for Int / Float / Double / Bool / Char / String, and the `get_by[K: Hash] / set_by / has_by / delete_by / get_or_by / get_or_else_by` family that encodes the key to `String` and uses the plain map underneath. Note `hash_key` is a **key encoding**, not a hash — for the six primitive impls it is injective (`Int::to_string`, String identity), so no collisions.
- `MutMap[K, V]` / `MutSet[T]` take explicit `(hash_fn, eq_fn)` dictionaries; `Set[T: Hash]` uses the same Hash bridge.

## Options

**A. Generalize the builtin surface over `[K: Hash]` before the tag.** `Map::get[K: Hash, V](m: Map[K, V], k: K)` lowers through `hash_key` to the existing String-keyed storage — the runtime representation does not change, only the checker-facing signatures and the lowering. `Map[Int, V]` then works out of the box, which is the intuitive surface. Two things to settle if chosen:
  - the injectivity contract: a user `impl Hash` whose `hash_key` collides silently **merges keys** — P0-shaped. Either rename the trait for this role (`MapKey` / `ToKey`, stating injectivity) or keep `Hash` and document it as an encoding, and consider a debug-lane collision check.
  - `keys` must decode or keep the original keys (probably store originals alongside, or return `Array[K]` only when K's decode exists — simplest is to store `(K, V)` pairs and keep the string only as the index).

[...]

PR #2849, opened 2026-09-15
Refuse field access on a builtin container receiver (#2820)

### LINKED ISSUE #2820: `vibe check` accepts field access on an Array / Map / Bytes / String receiver that `vibe build` rejects
Found while measuring #2475 (PR #2819), where it very nearly caused a wrong decision — see the bottom.

## The gap

`.field` on a receiver whose type is a builtin container is **clean in `vibe check` and rejected by `vibe build`**. #2807 added the non-struct guard for enum, tuple and `Option` receivers; `CtArray`, `Map`, `Bytes` and `String` are not in that allow-list.

Measured on `main` at `d321d4d`, four one-file programs, each the whole file:

```vibe
fn read(m: Array[Int]) -> Int {
  m.v
}

fn main() -> Int {
  read([11, 22, 33])
}
```

| receiver | `vibe check` | `vibe build` |
|---|---|---|
| `Array[Int]` | **clean, exit 0** | `unknown struct field: v` |
| `StringMap[Int]` | **clean, exit 0** | `unknown struct field: v` |
| `Bytes` | **clean, exit 0** | `unknown struct field: v` |
| `String` | **clean, exit 0** | `unknown struct field: v` |
| `Option[Int]` | ``field access `.v` on a non-struct value`` | (same) |

The `Option` row is the control: that one is caught, and its message is the right shape. Note there is no struct in these programs at all, so this is not a name-collision case — there is simply nothing for `.v` to mean.

## Why it matters

Two separate problems, both named by AGENTS.md:

1. **`vibe check` cannot answer "does this file compile"**, which is the property that section states as the goal for the CLI query surface ("答えられないなら、それは診断の穴であって呼び出し側の作法の問題ではない"). Empty output has to mean clean.

2. **The build's message does not lead with the edit and has no location.** `unknown struct field: v` asserts the receiver *is* a struct whose field list lacks `v`; the truth is that the receiver is an `Array` and field access does not apply to it. Compare the `Option` wording, which says what is wrong.

## Suggested fix

Extend the `not_struct` allow-list in the `EDot` field-type match in `lib/@vibe/compiler/checker/checker.vibe` — the same place #2807 added `CtEnum` and tightened the `CtTuple` arm — to cover `CtArray`, the `Map` shape, `CtBytes` and `CtString`, reusing the `Option` arm's message. Doing it there gives the diagnostic a location for free, and makes the build-lane message unreachable rather than merely nicer.

## How this was found

#2475 refuses a declared type that takes a builtin's spelling. Deciding whether `struct Array[T]` collides needed a probe, and the obvious one was:

```vibe
struct Array[T] {
  pad: Int;
  v: T
}

fn read(m: Array[Int]) -> Int {
  m.v
}
```

That is accepted and answers `22` — the *second* field, which is normally how you tell a genuinely resolved struct from a first-field offset landed on by luck. It proves nothing, because of this bug: the identical program with no struct named `Array` anywhere is also accepted. The question had to be re-asked through `match` on an `enum` spelling, which does check its scrutinee, and the answer came out the other way.

So the cost of this gap is not only the missing diagnostic: it makes `Array` and `Map` receivers unusable as evidence about anything else.

## Current implementation constraint

The EDot exclusion for String/Bytes/Array is deliberate in the source, but its method-call rationale is stale for builtin methods (see the measurement in the existing comment). Fix value-position field access while preserving legitimate method resolution and unknown/type-variable receivers; do not widen a non-struct guard by spelling alone.

mizchi:
Closes #2820.

`.v` on an `Array` / `Map` / `Bytes` / `String` receiver was **clean in `vibe check`** — empty output, exit 0 — and rejected by the build with an unlocated `unknown struct field: v`. Empty output has to mean clean, or `vibe check` cannot answer "does this file compile"; and the build's message asserts the receiver *is* a struct whose field list lacks `v`, when the truth is that the receiver has no fields at all.

## The stale rationale is worse than stale

#2807 added the non-struct guard for enum, tuple and `Option` receivers. These four were excluded with a reason: "`.m` on them parses as a method call elsewhere, not a value-position EDot". That situation **cannot occur**. A dot-call is `ECall(EDot(recv, m), args)` and is typed by the callee-is-EDot arm of the ECall case (`checker.vibe:8396`), which checks the **receiver alone** and never recurses into the whole EDot node — so a call never reaches the arm to be refused. #2807's tuple arm already says this; the builtin exclusion was written as if it did not.

Measured on a stage2 built from `8c966c3`, one file, with `fn Array::my_len(xs: Array[Int]) -> Int` declared:

| written | before | after |
|---|---|---|
| `xs.my_len()` | **103** | **103** |
| `s.shout()` (`String` twin) | **204** | **204** |
| `xs.my_len` — no parens | clean in check, codegen died with `unknown struct field: my_len` | ``call `Array::my_len(...)` if you meant a method; …`` @ 7:14-20 |

A method value is not a supported form on any receiver — the struct twin `let f = p.my_x` was already refused here, located, as `unknown field 'my_x' on struct Pt`. Flagging the builtin receiver makes the two agree rather than taking a working program with it, and the third row is a second instance of this same bug, closed by the same arm.

## The four rows

| receiver | check, before | check, after |
|---|---|---|
| `Array[Int]` | clean, exit 0 | ``call `Array::v(...)` if you meant a method; field access `.v` needs a struct, and Array is a builtin type`` @ 9:5-6 |
| `StringMap[Int]` | clean, exit 0 | same, naming `Map` @ 14:5-6 |
| `Bytes` | clean, exit 0 | same, naming `Bytes` @ 3:5-6 |
| `String` | clean, exit 0 | same, naming `String` @ 3:5-6 |
| `Option[Int]` (control) | `field access `.v` on a non-struct value` | **unchanged** |

The build lane now reports that same located message, so the message #2820 objected to is unreachable for these receivers rather than merely nicer.

`Option` keeps the general wording on purpose: a field read on an `Option` has a second plausible edit (match it), so leading with "call `Option::v(...)`" would be wrong. The builtin message is reserved for receivers with no fields under any reading, and `err_type_option_field_access` pins the split.

## The `Map` half is a different arm

`StringMap[V]` has no type of its own — `resolve_type_expr_core_shadowed` rewrites it to `CtNamed("Map", [CtString, V])` when no source declaration claims the spelling — so a map receiver arrives with no typedef and never reached the non-struct arm. Adding `CtArray`/`CtBytes`/`CtString` to the allow-list does nothing for it.

Its test is the compiler's **own type-constructor registry at the arity the name was applied with**, not the spelling #2820 asks this not to widen by: `compiler_owned_type_arity(name)` owns the name **and** `find_typedef` does not. Both keys are needed — `Map` is not in #2475's refused set (that covers `Array`/`Option` at arity 1 and `resolve_builtin` at arity 0), so a user's `struct Map[K, V]` is legal, takes the `TDStruct` arm, and still answers **9** in one file and across a module boundary alike.

Arity ≥ 1 keeps a shadowing type **formal** out. Measured on both compilers: `fn read[Map](m: Map)` unapplied arrives as `CtNamed("Map", [])` and compiles clean; `fn read[Map](m: Map[String, Int])` is already rejected upstream by the kinded-formal check; `fn read[Array](m: Array[Int]) { m.v }` resolves through `CtApplied` and stays clean.

The `--single-file` lane, which resolves no imports, does not reach this arm either: with no typedef visible, `Map[Int, String]` is read as the builtin map with a non-String key and #2263's annotation diagnostic fires first — measured identical before and after.

**The accepted cost, named rather than closed by guessing:** the four arity-0 builtins (`StringBuilder`, `ByteStream`, `HostStream`, `PromptText`) stay tolerated, because at arity 0 the builtin and an unapplied formal are the same `CtNamed(name, [])` and refusing one refuses the other.

## Found on the way — filed, not fixed here

#2848 (P0): a field read on a type **formal** receiver resolves by spelling against any struct declaring the name. `struct Crate { payload: Int }` + `fn read[T](m: T) { m.payload }` + `read(7)` compiles clean and answers **0** — slot 0 of a tagged integer read through `Crate`'s offset. It is #2807's shape one rung over, and not fixed here because the same arm tolerates an unresolved cross-module name on the incremental lane. `builtin_receiver_bare_formal_ok` pins the current tolerance and says so, so fixing #2848 will have to change that fixture — the intended signal.

## Fixtures

Four refusals, the `Option` message-split control, and four controls that would catch a refusal grown too wide: the two UFCS answers (**103** / **204**), a bare formal plus a formal named `Map` (**10**), and a declared `struct Map[K, V]` whose field still resolves (**9**). All four controls were measured green on the **baseline** stage2 before the change, so they are evidence rather than decoration.

## What was run

- **stage1 → stage2 succeeded**, so the compiler's own sources contain no field access on a builtin container receiver.
- `check_typecheck_fixtures.sh`: **347 rows ok**, both lanes.
- `check_lambda_bound_refusal.sh`: 71 fixtures ok.
- Cross-module controls: declared `struct Map` **9**, `MutMap` from `@vibe/core` **7**.
- `lint_architecture_debt.sh`: the ARCH010 ratchet for `checker.vibe` tightened 3 → 2, since the change removes one `None => CtUnknown`. Only that line is touched — `--update` also duplicates a hand-annotated `cache.vibe` entry and re-sorts the block, which is unrelated churn.

The unit suite and the stage3 fixpoint build were still running when this was opened; both are reported in a follow-up comment.

---


mizchi:
## `wasi-p3-gate` is red, and it is not this PR's

`wasi-p3-gate (wasmtime v47.0.2)` fails with:

```
named hoststreams component gate FAILED: .../stream_for_proj.vibe did not compile: unknown name: HostStream::equals
```

**It is red on the base branch too.** On main's tip `d92802ab`, `wasi-p3-gate` is the *sole* failing job, with the byte-identical message. And the compile failure reproduces locally on a stage2 built from **`8c966c3`** — this PR's base — so it predates the diff:

| compiler | result |
|---|---|
| stage2 from `8c966c3` (this PR's base) | `unknown name: HostStream::equals` |
| stage2 from `8c966c3` + this PR | `unknown name: HostStream::equals` |

What changed is wiring, not the compiler: `fd75349` (#2846, #2592) added `bash "$SCRIPT_DIR/test_named_hoststreams_component_gate.sh"` to `scripts/test_wasi_p3_guarantee_gate.sh`. That gate existed but nothing in CI ran it, so #2846 did exactly what it was written to do and the first run that could see this bug reported it. Main has been red on this job since.

**No fix exists yet to port**, so there is nothing to bring into this PR. I filed **#2851** (P1, blocker) with the diagnosis: `eq_name_is_invisible_nominal` excludes scalars, type params and formal marks but not *compiler-owned builtin* names, so `==` involving a `HostStream` emits `HostStream::equals` as a reference the link is meant to resolve against the declaring module — and `HostStream` is a builtin that no module declares. The issue also names the prior question that has to be answered before patching the predicate: the probe writes no `==` at all, so the `for b in h.s` desugar is generating one, and fixing the predicate blind risks trading an unresolvable name for a silently wrong loop.

This PR does not touch `==` lowering, so fixing it here would widen it. I did not spend the one re-run: the failure is a deterministic compile error that I reproduced on the base compiler and that main's own CI shows red, which is strictly stronger evidence than a re-run would give.

## Everything else is green, including the fixpoint

`ci-required` **success**, and `wasi-p3-gate` is the only non-success job in the whole run (31 jobs):

| check | result |
|---|---|
| `ci-required` | ✅ |
| `compiler-build` (seed→stage3, **stage2==stage3 fixpoint**) | ✅ |
| `unit-tests` shards 0–7/8 | ✅ all eight |
| `compiler-gate` (early, mid, late, preflight, selftests, docs, examples, playground, stage2 oracles, contracts core, contracts compile-only) | ✅ |
| `gc-gate` (heap, validate) | ✅ |
| `smoke` (install, library, lsp, debug, profile, js) | ✅ |
| `vibe-fmt-check`, `structural-lint`, `review-regressions`, `perf-metrics` | ✅ |
| `wasi-p3-gate (wasmtime v47.0.2)` | ❌ — #2851, red on main |

That settles the two items the PR body left open. The **stage3 fixpoint** is verified by `compiler-build` rather than a local run, which is the better check anyway — it builds from a clean checkout.

Locally, alongside it: the full unit suite on the candidate stage2 reported **1292/1292 active unit-test files passed, 0 failures**.

Codex reviewed `2397e67` and posted no findings.

---

PR #2852, opened 2026-09-15
test: ratchet documented-declaration counts (#2822)

### LINKED ISSUE #2822: Ratchet the count of documented declarations: merging two `///` blocks silently transfers one doc and deletes another
## The failure mode

Inserting a function between an existing `///` doc block and the declaration it documents merges the two blocks. Consecutive `///` lines are one block attached to the following declaration, so:

- the **new** function inherits the **old** function's documentation, and
- the old function loses its documentation entirely.

Nothing catches this. It compiles, formats clean, passes every gate, and the only visible symptom is that `vibe doc-at` and `vibe symbols` now describe the wrong function — which is exactly the surface AGENTS.md points readers at for code navigation.

It happened **twice in one session** while landing #2475 (PR #2819, caught both times by review, not by a gate):

1. inserting a helper between an enum helper's `///` block and its `fn` line;
2. appending the #2475 block directly after the #1464 duplicate-constructor block, so `builtin_spelling_decl_error` carried #1464's text and `duplicate_ctor_errors` carried none.

## The detector that does NOT work

The tempting one is lexical: *"a `///` block whose non-first line opens like a new block (`#NNNN:`)"*. It is a **proxy**, and measurement kills it. Over `lib/`, excluding generated bundles:

```
doc blocks scanned: 4527
candidate hits:      110
```

Spot-checking, all of them are ordinary multi-issue doc blocks that begin a paragraph that way — `lib/@vibe/parser/lexer.vibe:448`, `lib/@vibe/concurrent/concurrent.vibe:574`, `lib/@vibe/builtin/int.vibe:91`, `lib/@vibe/cli/entry.vibe:76`, and so on. A gate at a 2.4% hit rate on legitimate code gets allowlisted into uselessness — the shape AGENTS.md describes under #2248 ("代理が必要になる構造を無くす").

## The detector that does work

Count **documented declarations**, not blocks. Merging N blocks into one always loses exactly N−1 docs, and the declaration total does not move. `vibe symbols` already answers this directly: a documented declaration gets a fifth `DOC` field, an undocumented one has four.

Measured on `lib/@vibe/compiler/checker/checker_stmt.vibe` across the broken and fixed states:

| state | declarations | documented |
|---|---:|---:|
| merged (broken) | 303 | 155 |
| separated (fixed) | 303 | **156** |

That is exact, not a heuristic, and it needs no per-file expectations — a per-file monotonic floor is enough, since a deliberate doc removal is rare and can lower the floor in the same commit that does it.

## Shape of the fix

- `vibe symbols <dir>` sweeps a whole tree in one process (#2381 — the whole tree is ~43 s, against ~17 minutes one process per file), so a repo-wide sweep is affordable in a gate.
- Store the per-file floor next to the other ratchets and fail when a file's documented count drops below it.
- **The gate owes a `scripts/<name>_test.sh` that mutates real input and asserts the gate FAILS** (#2248). The natural mutation is the exact defect: delete the blank line between two doc blocks in a fixture and assert the count drops by one. The self-test must also confirm the mutation landed, since an edit that matches nothing passes while proving nothing.
- It must not read its own corpus, and it must not assume `rg` or GNU-only `sed -i` (#2252, `check_gate_portability.sh`).

## Why P1 and not P2

It is not silently wrong at runtime, so it is not P0. But it silently corrupts the answer `vibe doc-at` / `vibe symbols` give, and those are the surfaces AGENTS.md designates as the LLM-facing IDE query layer — a wrong answer there is paid on every future navigation, which is the cost that section exists to prevent.

mizchi:
Closes #2822

#2820 already landed in #2849. This PR is only the documented-declaration floor.

`vibe symbols` DOC fields, per file, against `scripts/documented_decls_floor.txt`. Merging two `///` blocks onto one declaration drops the count and fails. Self-test mutates a fixture, asserts the mutation landed, and asserts FAIL.

Work directories are per-process (`mktemp`) so the gate and its self-test can run as pkfire siblings.

Wired next to builtin-shadowing in compiler-contracts CI.


mizchi:
Codex P1s:

- Concurrent `_build/_documented_decls`: fixed in 7aaafb47c (per-process `mktemp`).
- Diagnostic leading with the edit: not taking this. #2820 asked to reuse the Option wording (`field access \`.v\` on a non-struct value`) so check and build agree. Changing the message is a follow-up, not this bug.

This PR is blocked on main's wasi-p3-gate (`HostStream::equals`). That is https://github.com/mizchi/vibe-lang/pull/2854.

PR #2854, opened 2026-09-15
fix: skip derived equals for HostStream struct fields

mizchi:
Main CI is red on `wasi-p3-gate` after #2846 actually ran named-hoststreams:

```
named hoststreams component gate FAILED: stream_for_proj.vibe did not compile: unknown name: HostStream::equals
```

`desugar_derives` emits `T::equals` for every struct. A field of `HostStream` is not content-comparable (the unsafe-types walk already knows that), but generation ignored the walk and `eq_for_typed` treated HostStream as an invisible nominal.

Skip equals generation for unsafe structs/enums. Pinned by `hoststream_struct_eq_test.vibe` and `ok_hoststream_struct_field`.

This unblocks every PR, including #2852.


mizchi:
First commit skipped `T::equals` for unsafe structs. That unblocked `HostStream` fields, then CI showed `TaskCell::equals` still being called (late gate, unit shards).

Follow-up: compiler-owned opaques are not invisible nominals, so `eq_for_typed` uses raw `==` for `HostStream`. `TaskCell::equals` is still generated.

The next named-hoststreams probe (`HostStream::next` Option protocol) cannot compile through the Async handle (ADR-0076, even on seed). Rewritten to `host_stream_next` so the gate still checks two live streams / EOS / close.

Issue #2851, opened 2026-09-15
A struct with a builtin-nominal field cannot be declared: the comparator generator ignores the exclusion the `==` dispatch honours


mizchi:
`wasi-p3-gate (wasmtime v47.0.2)` is failing on **main** and therefore on every open PR. On main's tip `d92802ab` it is the sole failing job:

```
named hoststreams component gate FAILED: .../stream_for_proj.vibe did not compile: unknown name: HostStream::equals
```

## What actually breaks

Not streams, not `for`, not `==`. **A struct declaration with a builtin-nominal field, on its own.** Measured on a stage2 built from `8c966c3`, each the whole file:

```vibe
struct Holder { s: HostStream }

let run: () -> Int with Async = () -> {
  0
}
```

→ `unknown name: HostStream::equals`

The struct is never instantiated, never compared, and the program contains no `==`. Varying only the field type:

| declared field | result |
|---|---|
| `HostStream` | `unknown name: HostStream::equals` |
| `ByteStream` | `unknown name: ByteStream::equals` |
| `StringBuilder` | `unknown name: StringBuilder::equals` |
| `Future[Int]` | `unknown name: Future::equals__N3_Int` |
| `Array[Int]` (control) | compiles |

The gate's `for b in h.s` probe is incidental — it happens to be the only place in the tree that declares such a struct. `build_host_stream_for` emits `EBinOp("<=", EInt(0), ...)` on an `Int`; it generates no `==` at all, and `for b in s` over a bare `HostStream` passes in the same gate run.

## Root cause: two passes disagree

`lib/@vibe/compiler/normalize/desugar_trait_dict.vibe:17659`:

```vibe
// Generate `T::equals` for every struct (not just derive(Eq)) so the
// `==` dispatch below always resolves; structural equality is the
// sensible default for any struct, ...
if !dtd_source_op_defined(stmts, qualified_name(name, "equals")) {
  dtd_fresh_reset_for_decl()
  Array::push(gen, gen_eq_fn(name, fields, known_types))
}
```

[...]

Issue #2856, opened 2026-09-15
`HostStream::next` is invisible to the `Async` handle pass; the bare `host_stream_next` spelling compiles


mizchi:
The second bug the newly-wired `wasi-p3-gate` was hiding. #2851 is the first; with that fixed the gate reaches `stream_protocol.vibe` and fails there instead.

**Pre-existing, not introduced by #2851's fix.** Everything below is measured on a stage2 built from `8c966c3` — the last commit whose CI was green, and which predates both #2846's gate wiring and the #2851 fix.

## Minimal repro

Two spellings of the same operation, same program shape:

```vibe
// COMPILES
let run: () -> Int with Async = () -> {
  let left = host_stream_named("left")
  let b = host_stream_next(left)
  if b >= 0 { b } else { 0 }
}
```

```vibe
// FAILS
let run: () -> Int with Async = () -> {
  let left = host_stream_named("left")
  match HostStream::next(left) {
    Some(byte) => byte,
    None => 0
  }
}
```

```
handle of effect 'Async' cannot be compiled here: every `perform Async` this
handle covers has to be visible to it, and one of them is not. No call in the
handled body itself is opaque to this pass, so look one level down -- in the
functions the body calls that carry `with Async`, a `perform Async` reached
through a parameter or through a binding that is not a top-level `fn` is
invisible here. Give that parameter or binding an effect row (`with Async`), or
move the `handle` into the function that performs. (ADR-0076 evidence-passing
migration.)
```

The qualified `HostStream::next` is the only difference. Narrowed by elimination — none of these is the trigger:

- **not the helper function.** Inlining the `match` (the second sample above, no `fn` at all) fails identically.
- **not a `perform` nested in a call argument.** Hoisting it to its own `let` binding first fails identically.
- **not the `Option` return.** `host_stream_next` returns a raw `Int` and compiles, but that is the same call the `for` desugar emits, which is why the loop probes pass.

## Why the gate did not catch it before

`build_host_stream_for` lowers `for b in s` to the **bare** `host_stream_next`, so every `for`-based probe in `test_named_hoststreams_component_gate.sh` takes the compiling spelling. `stream_protocol.vibe` (#1955, the nominal-`Option` protocol probe) is the only one that calls `HostStream::next` directly, and it sits after the `stream_for_proj` step that #2851 makes fail — so it was never reached.

[...]

PR #2857, opened 2026-09-15
Restore the unmintable-comparator trap that two merges cancelled out (#2851)

mizchi:
Follow-up to #2851. This is about a state neither fix produced on its own.

## The two fixes cancel each other

Two fixes for #2851 landed within the hour:

- **`d3e32af`** (#2854) narrowed `eq_name_is_invisible_nominal` to answer `false` for every `compiler_owned_type_arity` entry.
- **#2855** added `eq_name_is_unmintable_builtin` — true *only* for such a name — and guarded both comparator-generating arms on `eq_name_is_invisible_nominal(n) && eq_name_is_unmintable_builtin(n)`.

Those conditions are mutually exclusive. On main the conjunction is `false` for every name, so **both `eq_trap()` arms are unreachable** and the unmintable leaf falls through to the raw `==` below — which compares an aggregate by reference.

## What that answers

One file, `struct Holder { s: StringBuilder }`, compared two ways:

| probe | `8c966c3`, before either fix | #2855's compiler | main `12cd9f0` |
|---|---|---|---|
| two builders, same content | ICE: `StringBuilder::equals` unresolved | trap | **`0`** |
| the same builder, twice | ICE | trap | **`1`** |

The answer depends on object identity, at a comparison the program performs, in a program that compiled clean. AGENTS.md ranks that above a crash — P0 (silently wrong) over P1 (it crashes) — so main is a severity regression against `8c966c3` even though it makes more programs run.

The declarations both fixes exist to keep compiling are unaffected: `struct Holder { s: HostStream }` and `struct Holder { s: Future[Int] }` answer 42 on all three compilers, and `d3e32af`'s two tests stay green.

## A prediction that was wrong, and why it matters

I expected a larger regression: that the six registry names `lib/` also declares — `Attempt`, `MutMap`, `TaskGroup`, `TaskHandle`, `Sender`, `Receiver` — would lose content comparison, since the narrowing excludes them too. Measured, they do not. A visible declaration takes an earlier arm, so the invisible-nominal path never sees them; `struct W { a: Attempt[Int, String] }` holding two equal `Granted(7)` values answers `true` on all three compilers.

That is why this PR is about the leaf and not about those six. The scope came from the measurement, not from the diff.

## The fix

`eq_name_is_invisible_nominal` goes back to its pre-`d3e32af` body, and the conjunction both arms test becomes one exported name, `eq_name_needs_unmintable_trap`. So:

- the unmintable leaf traps again — fail-closed, and the treatment AGENTS.md already records for a site that fires while *generating* a comparator rather than at a comparison the program performs;
- "can anything mint `<n>::equals`?" is asked in one place, by the predicate that asks it precisely: the registry MINUS the names `lib/` declares, with `dtd_eq_type_owners` on top;
- the two halves cannot drift apart again without both call sites changing.

## Why nothing caught the collision

Every test covering those arms asked only whether the program **compiles**, and falling through to raw `==` compiles fine. That is the #2248 shape exactly: a check that trusts a proxy (does it build?) instead of the property (what does it answer?). Two independently correct fixes then composed into a third behaviour with nothing red.

`lib/@vibe/compiler/tests/registry_name_struct_eq_test.vibe` asserts the conjunction itself, plus the content answer and the compile-side cases.

**Red test, not yet run.** The mutation is exact — re-apply `d3e32af`'s narrowing on top of this branch and rebuild; the first `inspect` must go from `true` to `false`. It is building; I will post the result. Until then the test is unproven, and #2248 says an unproven gate means nothing.

## If reference identity is the wanted semantics

Then the fix is the opposite deletion — drop `eq_name_is_unmintable_builtin` and its two arms, and record the decision in the ADR, since `==` answering by identity for an opaque handle is a language decision rather than an implementation detail. I have implemented the reading that matches the written policy, but it is your call.

What is not tenable is keeping both: main currently ships 25 lines of comment describing behaviour the code no longer produces, and a test suite that cannot tell the two apart.

## Verification

**In flight at the time of this push, reported in a comment when they finish.** I opened #2855 before its verification finished and CI caught a regression minutes later; the lesson I took was not to delay the PR but not to claim a verification I had not run. So, explicitly:

| check | status |
|---|---|
| `vibe fmt --check` on all three changed files | **pass** |
| `stage0 → stage1` | **pass** |
| `stage1 → stage2` (the self-build: stage1 carries the change and compiles the compiler with it) | running |
| behaviour probes — the two `StringBuilder` rows must go back to trapping, `HostStream` / `Future[Int]` declarations must still answer 42, `Attempt` / `Array` / `Option` controls must stay `true` | queued |
| `registry_name_struct_eq_test`, `hoststream_struct_eq_test` (`d3e32af`'s), `desugar_trait_dicts_module_test` (the synthesis invariant) | queued |
| `check_typecheck_fixtures.sh`, including `ok_hoststream_struct_field` | queued |
| per-module split lane (`selfhost_build_metrics.mjs … prelude`) | queued |
| full unit suite | queued |
| the red build | queued |

The unit suite matters more than usual here. #2855's CI ran on the merge ref, which already carried `d3e32af` — so **those arms were dead in every CI run that has ever covered them.** This branch is the first time they are live under a full suite.

---


mizchi:
## Verification complete, including the red test the PR body listed as unproven

Merged before these finished, so here they are for the record. All on a stage2 built from this branch (`2bd36b6`), passed explicitly to every oracle.

| check | result |
|---|---|
| `stage0 → stage1 → stage2` | **pass** — the compiler compiles itself with the arms live |
| **full unit suite** | **1298/1298 passed**; 1298/1298 compiled with **0** left to the one-shot fallback |
| `check_typecheck_fixtures.sh` | **348 rows ok**, both lanes, 19 known debt |
| per-module split lane (`selfhost_build_metrics.mjs … prelude`) | **clean**, 29.3 s — the lane that reproduced CI's ICE for the designs #2855 discarded |
| `check_documented_decls.sh` + self-test | 1140 files, 0 above floor; 4/4 |
| Codex | **0 findings** |
| perf | no drift ≥ ±2%; output checks 10/10 |

Behaviour, one file each:

| probe | `8c966c3` | main `12cd9f0` | this branch |
|---|---|---|---|
| `Holder{StringBuilder}` == two equal builders | ICE | `0` | **trap** |
| `Holder{StringBuilder}` == the same builder | ICE | `1` | **trap** |
| `struct Holder { s: HostStream }` | fails | 42 | **42** |
| `struct Holder { s: Future[Int] }` | fails | 42 | **42** |
| `struct W { a: Attempt[Int, String] }`, equal `Granted(7)` | `1` | `1` | **1** |
| `Array[Int]` / `Option[Int]` controls | `1` | `1` | **1** |
| `stream_for_proj.vibe` (the gate probe #2851 failed on) | fails | compiles | **compiles** |

`registry_name_struct_eq_test` 5/5, `hoststream_struct_eq_test` (`d3e32af`'s) 2/2, `desugar_trait_dicts_module_test` 20/20.

## The red test: confirmed, on the second attempt

Re-applying `d3e32af`'s narrowing on top of this branch and rebuilding, the new test fails on precisely the assertion that matters:

```
FAIL lib/@vibe/compiler/tests/registry_name_struct_eq_test.vibe
       failing test: the unmintable-trap condition holds for names nothing can declare
         actual:   false
         expected: true
```

and the mutated compiler reproduces main `12cd9f0` row for row — `q1 -> 0`, `q2 -> 1`, everything else unchanged. So the diagnosis is closed: that one narrowing line is the entire cause, and the test catches it.

**The first attempt at this red test passed, and that result was worthless.** I ran the test in the normal checkout against the *mutated compiler binary*. But `eq_name_needs_unmintable_trap` is compiled from `lib/` sources, and those were already reverted to green — the red compiler was only compiling the test, not supplying the predicate under test. The mutation has to be in the **sources the test's imports resolve from**, so the run has to happen inside the mutated tree (I used a `git worktree`, which also keeps the main checkout clean).

That is the same shape as the bug this PR fixes, one level up: a check that looked like it was testing the property while actually testing something adjacent, and whose green told me nothing. Worth stating plainly, because "I ran the red test and it passed" would have been the wrong conclusion, and nothing in the output said so.

## Still open

- **#2856** — unchanged by this. The gate covers it by using the bare spelling (`d3e32af`), so no CI job is blocked, but `HostStream::next` still cannot compile through the `Async` handle.
- **The design question in the PR body stands.** This restores fail-closed because that is what the written policy says. If reference identity is the intended `==` for an opaque handle, the fix is the opposite deletion plus an ADR row; the merge does not settle that, it just stops main from carrying both mechanisms at once.
- The follow-up #2855 named: preflighting a composite `==` at the ladder so an unsafe leaf refuses at build time with a message instead of trapping at runtime.

---

Issue #2828, opened 2026-09-15
0.1.0: settle and implement frozen capability grants across build, instantiate and host providers


mizchi:
Replaces #2332's stale compile-time `NotGranted` plan. #2825 records the owner's newer instantiate-time decision; PR #2811 carries the ADR-0088 amendment and is still open at this audit. Review and land that decision before implementing against it. Do not describe it as current main behavior.

This is a pre-release contract task because later changes affect program behavior, host imports, binding locks and cache identities. It is synchronous capability work; async providers remain 0.2.0.

## Acceptance criteria

- [ ] Pin deterministic precedence for L1 allow/deny inputs, L2 persisted `BindingLock.optional_resolution`, and L3 preflight. Missing required authority fails before `main` with the edit that supplies it.
- [ ] Implement the recorded strict rule for unresolved authority and document migration from today's non-interactive fallback; an explicit parent handler remains a distinct case.
- [ ] Freeze the grant table before execution. Compile capability-independent bodies with both optional branches where required by the accepted ADR; evaluate arguments only in the granted branch.
- [ ] Specify the host-set grant constant and non-granted import stubs BEFORE changing lowering. Denied imports must not prevent instantiation, and stubs must not grant ambient authority.
- [ ] Represent granted-success and granted-failure (`Errored`) as the same contract on node and viberun, linear and GC, or refuse an unsupported lane explicitly.
- [ ] Pin `.wit`/raw host import ownership, public direct-builtin/provider semantics (#1962), and the synchronous host contract (#1346).
- [ ] Exercise denied, granted, failed, missing-required and explicit-handler cases end to end, plus cache reuse across different grant sets without stale authority.

Do not remove general constant folding just because capability selection moves: compiler feature/lane specialization (#2497) is a different use of `fold_const_bool_params`.

mizchi:
## The opening premise is stale: the decision is made and step 1 has landed

The first paragraph reads:

> #2825 records the owner's newer instantiate-time decision; **PR #2811 carries the ADR-0088 amendment and is still open at this audit. Review and land that decision before implementing against it.**

Checked today, none of that is still true:

- **PR #2811 is merged.** (Its title is "Incremental build: enforce the invalidation contract…", so it is also not the PR that carries the amendment — the reference looks like a mis-citation on top of being stale.)
[...]

Issue #2825, opened 2026-09-15
Incremental build: where the line stands, measured, and the one decision that unblocks it


mizchi:
Index issue for the incremental-build line. Supersedes #1959, #2818 and #2824, which are closed into this one.

Everything below is measured, on a stage2 built from the checkout that measured it, with isolated caches. Work is implemented on **PR #2811, still open as of 2026-09-15**; these are branch measurements until that PR merges.

## 1. Where the time actually goes

A warm compile of the compiler's own CLI closure, by **inclusive** time (self time cannot answer — leaf utilities are ~30 % spread across every phase):

| region | inclusive | share |
|---|---:|---:|
| post-merge linked compile | 6260 ms | 78 % |
| ├ effect-lowering prelude | 2563 ms | 32 % |
| ├ trait-dict desugar | 688 ms | 9 % |
| ├ body codegen | 626 ms | 8 % |
| ├ borrow + perceus | ~945 ms | 12 % |
| └ const-fold, DCE, index assignment, assembly | ~1180 ms | 15 % |
| merge | 832 ms | 10 % |
| source load + parse | ~970 ms | 12 % |
| **checker** | **0 ms** | **0 %** |

**The checker is already free warm.** Every remaining second is in the back end.

## 2. So the checked-module artifact cannot pay for itself

#1959's fifth criterion — "default-on promotion requires oracle parity and measured wall/heap benefit" — was asked four ways. Parity holds every time. The benefit never appears:

| variant | artifact | cli warm wall | cli warm heap |
|---|---:|---:|---:|
| `off` (control) | 9 MB | 9.4 s | 1957 MB |
| production `on` | 179 MB | 17.66 s (+87 %) | 3142 MB (+61 %) |
| drop `parsed_stmts` | 138 MB | 16.78 s (+77 %) | 3116 MB (+59 %) |
| … and the unread `CheckedProgram` fields | 95 MB | 13.54 s (+41 %) | 2731 MB (+40 %) |
| … and the verbatim input identity | 24 MB | 11.42 s (+22 %) | 2563 MB (+31 %) |

By bytes: 23 % parsed AST, **24 % fields the reuse path never reads back**, **40 % a verbatim copy of its own input identity** (this module's whole source *plus every direct dependency's full TypeEnv text* — superlinear in the dependency graph), 13 % what is read.

[...]

Issue #2669, opened 2026-09-11
A warm build re-runs the back end over the whole program: cached bodies carry global function indices (#2510 criterion 5)


mizchi:
Split out of #2510. Its first four criteria are done; the fifth — *the live set bounded by the edited module plus interfaces* — is blocked on something none of them describe, and #2668's measurements now say what.

## What a warm build actually spends itself on

Warm compile of the compiler's own closure (`codegen_lexer_test.vibe`, FS lane, cold run first into the same `VIBE_BUILD_CACHE_DIR`), profiled with `node --cpu-prof` through a stage2 built with `VIBE_WASM_NAMES=1`:

| area | share of 3.66s |
|---|---:|
| compiler/codegen | 28.7% |
| runtime helpers (`__rt_*`) | 25.6% |
| compiler/normalize | 11.5% |
| host JS / other | 8.5% |
| parser | 6.9% |
| core | 6.7% |
| compiler/core | 5.7% |
| compiler/perceus | 5.2% |
| compiler/loader | 0.1% |

codegen + normalize + perceus is **45.4%**. The front end is small and the loader is nearly absent.

Two things follow, both already demonstrated rather than argued:

- **Caching parses cannot reach the KPI.** #2510's fourth criterion assumed the front end. Built and measured in #2668: with the per-file AST cache on, a warm compile's heap is byte-identical to one with it off (571,165,952 both). Not "within noise" — the same number, on a lane whose allocation is deterministic. The parser's 6.9% is the ceiling, and the loader's 0.1% is why even that is not collected.
- **A one-module edit is nearly a cold build.** Measured in #2660: 765,264,132 against a cold 821,534,940, i.e. **93%**. The per-module prelude (#2575 item 2) does not change it — it costs a flat ~117 MB and recovers none of it, in every temperature.

## The blocker

Recorded in `docs/incremental-build.md` since #2647, and it is the thing to solve:

> a cached function body contains call immediates, and those depend on **global function index assignment**. A per-module body cache therefore cannot store a body and replay it into a build where indices moved.

So the back end re-runs for the whole program whenever anything moves, which on this codebase is every edit.

## What would close it

Not a shape I want to pick unilaterally — recording the options as I understand them:

1. **Index-independent bodies.** Emit call immediates against a per-module symbol table and resolve them at the link. Cached bodies then survive any index assignment. Cost: a relocation representation, and the link gains a patch pass.
[...]

Issue #2507, opened 2026-09-04
Four build units along the existing seams: frontend, checker, lowering + linear codegen, link


mizchi:
Part of #2494. The compiler is one closure of 314 files / 232 k lines; one compile is 8.7 s and holds 1.35 GB because everything is live at once. The seams for a split already exist: `.vpkg` contracts (ADR-0070), the binary AST (`lib/@vibe/compiler/ast_binary.vibe`, `docs/ast_binary_abi.md`), the checked-body transport (#1958, promoted per #2505), the persisted `CodegenBodyCache` with edit-stable index spaces (#2388, #2394, #2400), and the `vibec-hosted` world (`docs/vibec-component.md`).

Units:

1. **frontend**: lexer, parser, loader, contract. In: source files. Out: binary AST + module headers per file.
2. **checker**: in: headers + AST for one module and its dependencies' interfaces. Out: a checked module artifact.
3. **lowering + linear codegen**: in: one checked module + dependency interfaces. Out: relocatable bodies (the `CodegenBodyCache` entry shape).
4. **link**: monoify instantiation set, structural-eq comparators, capability const-fold + DCE, index assignment, assembly. The whole-program step, over cached bodies.

Tooling (fmt, grep, editor queries, lsp, gc backend, component/serve) are units a compile-only host never loads (#2499 is the prerequisite).

Each unit is its own build unit with its own tests, and every boundary is a serialization; the per-module design has to persist exactly these artifacts anyway, so the cost is paid once. The measured reason this matters more than caching: an RC-lane compiler uses *more* heap than bump on every input (full closure 1,526 MB vs 1,354 MB), so the live set can only be bounded by making the unit smaller.

**Memory accounting (Codex review on #2491, three rounds):**

- Separate linear memories do not make a process peak equal to the largest unit: every live instance's reserved pages plus the serialization buffers count together.
- Dropping the JavaScript reference to a `WebAssembly.Instance` is not reclamation (no disposal operation; pages stay until V8 collects them), and a request boundary in a reused isolate does not force collection either.
- Resetting an arena pointer inside a finished component that owns its own memory shrinks nothing and hands nothing to the next component.

[...]

PR #2866, opened 2026-09-16
Design: residual `vibe:host` WIT, self-describing build artifact, lazy CLI dispatch, and the component build convention (#1346)

mizchi:
Two design documents for #1346 (under #2828), plus ADR-0112 / ADR-0113 (proposed) and index/pointer rows.

## 1. `docs/internal/design/host-contract-artifact-lazy-cli.md` (ADR-0112)

#1346 (the `.wit` / raw host contract), #2499 (per-verb CLI artifacts) and `component-lazy-dispatch.md`'s "capability comes first" blocker are stuck on one missing object: a machine-readable statement of what an artifact requires from its host, written once by the emitter where it decides the import section, and read three ways.

- **WIT** — a `vibe:host` catalog, one interface per registry provider label and one function per raw field, *generated* from the builtin registry and `docs/generated/host-runtime-contract.json` (plus gated `resources` and `field_labels` tables). A raw `vibe.*` import is the vibe-raw lowering of the same function; the gate checks the mapping so the two lowerings cannot drift. Each entry world inlines only the used functions, never a catalog interface whole. Failure stays a trap until #2828's `Errored(E)` ABI exists.
- **Build artifact** — no new container: the core module plus custom sections. `vibe.entry` names the kind, `core-export` and `component-export`; `vibe.capabilities` carries `used` rows and (core-module only) optional rows. Required = import section, sidecar = projection. The per-verb intermediate is the capability-independent body cache (#2825), not the checked-module artifact.
- **§2.5 / §2.6** — a component boundary is an executable boundary, never a library boundary: WIT has no type parameters, function types or effect rows and two instances have two memories, so erased generics, closures, witness dictionaries and evidence frames cannot cross and no metadata recovers them. A verb component is a whole-program link; the component is the unit for authority, isolation and loading, not for optimization; per-verb linking still pays the per-link phases #2825 measured until #2507 / #2669 land.
- **Lazy CLI** — `vibe-commands-v2` entries are `<label>:<interface>.<function>`, derived at build and verified against the artifact on load, so authority is settled and the host composed before the artifact is read; `stdout.write-stream` as an import gives streaming output; `--trust-precompiled` is scoped to the toolchain directory.
- **Gate** — three sides (generator diff, both-runner semantic fixtures pinning values, manifest-row agreement), each with its red test. The `Fs::remove` parity item closed with #2823.

## 2. `docs/internal/design/component-build-convention.md` (ADR-0113)

The build-side half: **the export surface is the contract, WIT is derived, vibe-to-vibe is transparent.**

- `vibe build --component` derives the `.wit` sidecar, a `vibe.contract` section (the vibe-facing contract as a `kind = component` `.vpkg`), the `vibe.entry` kind and the import list. A package builds from its `index.vpkg` surface; a loose `.vibe` stays path-only under `vibe:app`. An unannotated or unprojectable export is refused naming the edit, never skipped. A hand-written `.wit` is a bug.
- The kind is read from the surface: exactly `vibe_command` → `command`, exactly `handler` → `handler`, else `service` with the facade interface named after the package's last segment under `scope:pkg@x.y.z`. `command` / `handler` refuse rows that would change their fixed ABI.
- Admitted boundary types are a bijection with WIT: today's mapping plus `export struct` ↔ `record`, `export enum` ↔ `enum` / `variant`; type parameters, function types, opaque, private and recursive boundary types are refused.
- `Exception[E]` on an export projects to `result<T, E>`, with `handle { Ok(body) } with Exception[E] { Throw(e) => Err(e) }` generated in the lift and the re-raise in the consumer's shim (revises #1324's trap stance for components only). One exception kind per export. User effects cross as tail-resumptive imports only.
- A vibe consumer writes the ordinary `import @scope/pkg { f }`: the store holds a derived `index.vpkg` extracted from `vibe.contract`, the loader switches linkage on `kind = component`, calls lower to canonical imports through generated shims, a `service`'s host requirement is the union over its exports, and a component dependency makes the consumer a component. Foreign WIT enters through the inverse projection at `vibe add`. Three round-trip laws form the gate.

## Decisions (owner, 2026-09-17)

ADR-0113's four open items are settled and recorded where each applies: `Exception[E]` → `result<T, E>` adopted; a component dependency makes the consumer a component, as a refusal on a core `vibe build`; a boundary `struct` / `enum` must be `export`ed; the facade interface is named after the package's last segment. Nothing remains open in that document. ADR-0112 still leaves the manifest column's existence to the owner.

## Review

The owner's round on 85581fc and the Codex rounds are addressed in c4e0043 and ea6bc7d; the per-thread answers are collected in [this comment](https://github.com/mizchi/vibe-lang/pull/2866#issuecomment-5711020088) because an unsubmitted draft review on this PR blocks inline replies from the same account.

## Checks

All 31 checks green on the current head. `check_doc_classification.sh`, `check_doc_path_citations.sh`, `check_doc_commands.sh` pass locally; pre-commit review lint passed.

https://claude.ai/code/session_01RN3TSMAuiEoFyLQ1SXRge7


docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,559 @@
...
+  resource response {
+    status: func() -> s64;
+    header: func(name: string) -> string;
+    body: func() -> string;
+  }
+  request: func(method: string, url: string, headers: string, body: string) -> response;
+}
+```
+
+`stdin`, `stdout`, `stderr`, `process`, `sh`, `tcp`, `profiler` and `clock`
```
mizchi: **[bug]** Section 1.1 says leftover WIT interfaces are `process`, `sh`, `tcp`, and `clock` (`sleep`), and that those "are the labels the registry already carries" plus `wit_kebab_ident`. They are not. `lib/@vibe/compiler/core/builtin_registry.vibe` tags `vibe_tcp_*` as `Some("Socket")`, both `sh` / `sh_capture*` and `process_exit` as `Some("Process")`, and `sleep` as `Some("Async")`. `standard_effect_policy.vibe` lists host providers as `Fs`, `Env`, `Stdin`, `Stdout`, `Stderr`, `Console`, `Process`, `Profiler`, `Http`, `Socket` — there is no `Tcp`, `Sh`, or `Clock`, `Async` is a runtime-managed effect rather than a host provider, and `Console` is a provider this list omits. A generator that followed this paragraph would emit `vibe:host/tcp` and `vibe:host/clock` while grants, `vibe.capabilities` rows, and ADR-0088 `allows` still speak `Socket` / `Process` / `Async`. The kebab of those labels is `socket` / `process` / `async`, not the names written here.

**Suggestion:** Derive interface names only by kebab of the registry's `Some("…")` label (one interface per distinct provider). Put `tcp_*` under `socket`, `sh*` and `process-exit` together under `process`, drop `clock`, and either keep `sleep` out of this residual (it is `Async`, which this document defers to #2832) or name that exception explicitly. Include `console`. Pin the emitted interface set against `standard_host_provider_resource_defaults` in the generator-diff gate.

docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,559 @@
...
+  text. A raw signature that does not match its WIT function's mapping fails
+  the gate, so the two lowerings cannot drift.
+- A component host and a core host implement one semantics. `viberun`'s
+  `--commands` face today serves `read-dir` as sorted names joined by `\n`
+  (`register_vfs_imports`), while the core lane's `fs_read_dir` returns the
+  same content as a packed string; under this rule both are the canonical and
+  the raw lowering of `fs.read-dir: func(path: string) -> list<string>`, and
+  the join is an artifact of the packed lane, not part of the contract.
+
+**Handles are WIT resources.** `http_request` returns an opaque `i64` today and
```
mizchi: **[bug]** "Handles are WIT resources" is decided, with `http_request` → `resource response`, implicit drop, and the same pattern for `sh_capture` / `tcp_connect`. Neither generator input encodes that grouping. `docs/generated/host-runtime-contract.json` has only field names, bands, and core type indices (`http_request` and `http_close` are sibling `i64` functions). The registry types those returns as `CtInt`, not as a handle of a named resource. `wit_type_text` maps `Int` → `s64`. Implemented from those sources, the WIT would be `request: func(...) -> s64` plus a separate `close`, which is not the example in §1.1. Resource methods also break the "one function, two lowerings" rule the gate is told to enforce: WIT `response.status: func() -> s64` has no handle argument, while raw `http_response_status` takes an `i64`; WIT drop has no function, while raw `http_close` is a real import. A fourth-side check that every raw field equals "the mapping of its WIT function" has no WIT function for the destructor and the wrong arity for the methods. That is a second hand-maintained table unless the inputs grow a resource grouping, which this document forbids.

**Suggestion:** Add resource grouping to the generator input (which ops construct / method / drop which resource) and make that table the owned source, gated like the rest of the JSON. Then define the fourth-side mapping as: constructor ↔ WIT constructor returning the resource, methods ↔ WIT methods with `self` dropped from the raw handle argument, destructor ↔ implicit drop (raw field exists, no WIT function). Until that encoding exists, do not claim the residual WIT is generated from the registry and JSON alone, and do not show `resource response` as the emitted shape.

docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,559 @@
...
+  import vibe:host/env;
+  import vibe:host/stdout;
+  export run: func(args: string) -> string;
+}
+```
+
+With the interface granularity being the label and the authority granularity
+being the operation, the world imports whole interfaces and the artifact's
+`vibe.capabilities` section (§2) says which operations are required and which
+are optional. A host may implement an interface partially; the preflight in
```
mizchi: **[bug]** The per-entry world in §1.3 imports whole package interfaces (`import vibe:host/fs`) and then says "a host may implement an interface partially; the preflight in §2.4 is what decides whether that is enough, operation by operation." That is true of a core `vibe.*` import section. It is false of a component world. Wasmtime satisfies an imported WIT interface in full; a missing function fails instantiate. So a `check` world that `import vibe:host/fs` requires write/remove/chdir even when `collect_used_builtin_names` only kept `fs_read_file`. That undoes the three-round compiler-host audit (six reachable `fs` ops, not the full surface), expands `vibec-hosted.wit`'s four root-level reads into the whole `fs` interface, and contradicts §3.1 step 2 ("the linker registers exactly the `vibe:host/*` functions the row names"). Open question 1 records the alternative and then takes whole interfaces anyway; the rest of the document already uses that choice as if partial WIT implementation existed.

**Suggestion:** Keep `vibe:host/fs` as the catalog of every Fs operation, but generate each entry world as an operation-shaped import of only the used functions (the compiler-host / `vibec-hosted` / current vfs wrap shape: inline interface or `import vibe:host/fs.{read-file, exists, …}`). Let `vibe.capabilities` optional rows stay a core-module fact. Then §3.1's "wrap is the world" and the generator claim that the world cannot list an op the module does not import are the same statement. If whole-interface imports stay, delete the partial-implementation sentence and accept that every fs user receives the full fs host.

docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,559 @@
...
+
+The issue asks for initialization, invocation lifetime and re-entry constraints
+independent of Wasmtime flags. Today they are conventions per entry shape,
+known to the launcher and to nobody else. The section names the shape so a
+host applies the right protocol without being told:
+
+| `kind` | `export` | protocol |
+|---|---|---|
+| `wasi-command` | `_start` | instantiate, write grant globals, call once, discard the instance. A second `_start` on the same instance is out of contract (`__heap_ptr` and every `let mut` at module scope are live state). |
+| `command` | `run` | one instance per invocation; `args` is argv joined by NUL, the result is `vibe-command-result-v1`-framed. `viberun` already does exactly this (`invoke_command` builds a fresh `Store` and instance per call). |
```
mizchi: **[bug]** `vibe.entry` is specified as a custom section on the core module (§2.1), and the `command` row sets `export=run`. That name is the *component* export `commands.rs` looks up (`COMMAND_EXPORT = "run"` after `comp_emit_component_wasm_command(core, "vibe_command", "run")`). The core export a command module actually has is `vibe_command` (`command_component_entry()` in `cli_support.vibe`, asserted in `dispatch_test.vibe`). A core-module host that reads `vibe.entry` and calls `run` will not find it; a component host that looks for a core custom section named `vibe.entry` on the `.component.wasm` / `.cwasm` will not find a core export list either unless the wrap copies and rewrites the section. The world example in §1.3 (`export run`) and the "entry points `vibe.entry` names" sentence in §2.5 repeat the same layer mix. `_start` is a core name; `run` is a lifted name; they are not interchangeable on one section.

**Suggestion:** Put both names in `vibe.entry` (for example `kind=command`, `core-export=vibe_command`, `component-export=run`), or emit the section per artifact kind so a core module names `vibe_command` and the wrap rewrites it to `run` on the component. Have the wrap copy the section into the component blob if component hosts are expected to read it. Gate both spellings (`command_component_entry()` and `COMMAND_EXPORT`) so they cannot drift again.

docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,559 @@
...
+| `handler` | `handler` | the 4-string `vibe serve` contract, or its `stream<u8>` form; one instance may serve many requests, and a handler that keeps state across them is the author's choice, not the host's. |
+| `library` | none | no entry; the module is an input to composition (`vibec`'s `compile` face, a bench or test harness). Re-entry is per export and is the export's own contract. |
+
+Initialization is one order for every kind: **link → write grant globals →
```
mizchi: **[bug]** Initialization is "link → write grant globals → call the export" for every `vibe.entry` kind, including `command`. `__vibe_granted$<label>` is a core-module exported mutable i64 (`capability-host-contract.md` §3). It is not a WIT export, and the `vibe:host` worlds in this document do not export it. `invoke_command` instantiates a component and calls `run`; nested core globals are not `Instance::get_global` on that component. A `--commands` host that follows this protocol cannot write the grant. When #2825 step 2 starts emitting optional rows, every `perform?` on a verb component would read the default `0` and take `NotGranted` even when the launcher intended to grant — silently wrong, not a trap. The older capability contract at least hedged this as "component-model global, or a host-set config value"; this document dropped the hedge and stated one protocol.

**Suggestion:** For core hosts, keep the exported global. For component hosts, either lift each grant as a WIT-exported `static mut` (or a config argument set before `run`) and name that in the world, or say that the `--commands` lane has no optional capabilities until that lift exists and therefore does not write globals. Do not tell a component host to `get_global("__vibe_granted$…")` on the component instance.

docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,559 @@
...
+vibe-commands-v2
+check	commands/check.cwasm	fs.read-file,fs.exists,fs.stat-token,fs.read-dir,fs.write-bytes,env.get,stdout.write-stream
+fmt	commands/fmt.cwasm	fs.read-file,fs.write-file,stdout.write-stream
+symbols	commands/symbols.cwasm	fs.read-file,fs.read-dir,stdout.write-stream
+```
+
+The column is a comma-separated list of `<interface>.<operation>` names, an
+operation suffixed `?` when the artifact declares it optional. It is written by
+`vibe build --component` from the import section and the `vibe.capabilities`
+section — never by hand — and it is a **cache of the artifact, verified on
```
mizchi: **[bug]** A `vibe-commands-v2` row is "a cache of the artifact, verified on load: when the artifact is opened, its own sections are compared to the row, and a disagreement is refused." Production rows under `$VIBE_HOME/toolchains/<name>/` are `.cwasm` from `Engine::precompile_component` (`commands.rs` `precompile_component` / `Component::deserialize_file`), loaded only with `--trust-precompiled`. That AOT image is wasmtime's compiled blob, not the original component binary: `vibe.entry`, `vibe.capabilities`, and the nested core import section are not there to compare. On the path the launcher actually uses, the column cannot be checked against the artifact, so it becomes the second truth this paragraph says it cannot be. `Component` type info after deserialize can still name WIT imports; core custom sections cannot.

**Suggestion:** Verify a `.cwasm` row against the component type's import list (available after deserialize), not against core custom sections, which means the world must be operation-shaped (issue 3). Alternatively keep the `.component.wasm` next to the AOT image and verify that, treating `.cwasm` as a cache of the component the same way the column is a cache of the requirement. State explicitly that `--trust-precompiled` rows do not re-read `vibe.capabilities` / `vibe.entry` from the AOT blob.

docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,698 @@
...
+  (emitter names, runner names, type indices) gains a fourth side: the WIT
+  text. A raw signature that does not match its WIT function's mapping fails
+  the gate, so the two lowerings cannot drift. Resource methods use the
+  mapping in the grouping table below, not a 1:1 field-to-function equality.
+- A component host and a core host implement one semantics. `viberun`'s
+  `--commands` face today serves `read-dir` as sorted names joined by `\n`
+  (`register_vfs_imports`), while the core lane's `fs_read_dir` returns the
+  same content as a packed string; under this rule both are the canonical and
+  the raw lowering of `fs.read-dir: func(path: string) -> list<string>`, and
+  the join is an artifact of the packed lane, not part of the contract.
```
mizchi: **[bug]** §1.2 decides that `fs.read-dir: func(path: string) -> list<string>` is the WIT definition and that the raw packed string is the same operation, with the `\n` join "an artifact of the packed lane, not part of the contract". That is not a `wit_type_text` lowering, and the generator inputs cannot emit `list<string>` at all. `vibe_fs_read_dir_raw` and `vibe_sh_lines_raw` are both `CtFn(..., CtString, ...)` in `builtin_registry.vibe`; `host-runtime-contract.json` types `fs_read_dir` and `sh_lines` as core type `3` (`(i64) -> i64`, the packed-string shape of `fs_read_file`); `wit_type_text` therefore yields `string`, not `list<string>`. The `Array[String]` surfaces (`Fs::readdir`, `sh_lines`) live in checker special-cases and `compile_call` (`packed_lines_to_array_expr` / `String::split` on `\n`), which are not generator inputs. Today's component face is also a joined string, not a list: `register_vfs_imports` returns `names.join("\n")` as `String`, and `vibec-hosted.wit` is `read-dir: func(path: string) -> string`. Implemented as written, a name containing a newline is one list element on the canonical lane and two strings after the raw split — a new silent disagreement, not a lossless pair of lowerings. The fourth-side gate cannot pass both `list<string>` and type index `3`.

**Suggestion:** Keep the residual type `string` on both lanes (matching registry, JSON, and the current vfs/hosted WIT), and treat newline-join + split as the language-surface encoding of `Array[String]`, not as the host contract. If the WIT definition really must be `list<string>`, give the raw lane a lossless list encoding (not `\n` join), put that encoding in the mapping table, and add it to the generator inputs so the fourth-side check can reject a type-`3` field. Either choice also has to cover `sh_lines`.

docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,698 @@
...
+| `vibe.capabilities` | `version=1`, then one row per **optional** operation: `optional\t<label>\t<module>\t<field>\t<grant global>` | designed in [capability-host-contract.md](capability-host-contract.md); emitter pending on the `perform?` lowering (#2825 step 2) |
+| `vibe.entry` | `version=1\nkind=<kind>\ncore-export=<name-or-empty>\ncomponent-export=<name-or-empty>\n` | **new, this document** |
+| `name` | function names | exists |
+
+Two rules keep this from becoming a second import section:
+
+- **Required is the import section.** A `vibe.*` import with no
+  `vibe.capabilities` row is required. The section never lists required
+  operations; a host that needs the label for a field reads the mapping from
+  the toolchain's contract manifest, and a host that lacks the mapping can
```
mizchi: **[bug]** §2.2 / §2.4 treat the import section as required operations and say a host that needs "the label" for a field reads it from the toolchain contract manifest, then refuses naming the label and the grant that supplies it. `docs/generated/host-runtime-contract.json` has field names, bands, and type indices — no provider labels. The registry labels are not 1:1 with fields either: `Stdout::write_stream` (`Some("Stdout")`) and `Console::write_stream` (`Some("Console")`) both lower to `vibe.stdout_write_stream`. The same collapse holds for `Stdin`/`Console` read, `Stdout`/`Console` write_char, `Stderr`/`Console` write_err_*, and `println`/`print`. A core host looking at `stdout_write_stream` cannot tell whether to require `Stdout` or `Console`. The catalog rule would also emit `stdout.write-stream` and `console.write-stream` as two WIT functions for one raw field, which breaks the "one function, two lowerings" bijection. A v2 column that names `console.write-stream` would then disagree with a core import section that only has `stdout_write_stream`.

**Suggestion:** Make the field→label map an explicit, gated generator input, and decide the many-to-one cases: either alias the legacy `Stdout`/`Stderr`/`Stdin` operations onto the `console` WIT functions (one raw field, one grant), or record a set of labels per field and require the union. Until that table exists, do not claim a core host can name "the" grant from the import section plus today's JSON.

docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,698 @@
...
+name; `run` is a lifted name. The wrap copies `vibe.entry` into the component
+blob and fills `component-export`. The gate pins both spellings
+(`command_component_entry()` and `COMMAND_EXPORT`).
+
+| `kind` | `core-export` | `component-export` | protocol |
+|---|---|---|---|
+| `wasi-command` | `_start` | (none) | instantiate, apply grants, call `_start` once, discard the instance. A second `_start` on the same instance is out of contract (`__heap_ptr` and every `let mut` at module scope are live state). |
+| `command` | `vibe_command` | `run` | one instance per invocation; `args` is argv joined by NUL, the result is `vibe-command-result-v1`-framed. `viberun` already does exactly this (`invoke_command` builds a fresh `Store` and instance per call). |
+| `handler` | `handler` | `handler` | the 4-string `vibe serve` contract, or its `stream<u8>` form; one instance may serve many requests, and a handler that keeps state across them is the author's choice, not the host's. |
+| `library` | (none) | (none) | no entry; the module is an input to composition (`vibec`'s `compile` face, a bench or test harness). Re-entry is per export and is the export's own contract. |
```
mizchi: **[bug]** The `vibe.entry` kind table classifies `vibec`'s `compile` face as `library` with empty `core-export` and `component-export` ("no entry"). §2.5 then lists "`vibec`'s `compile`" among "the entry points `vibe.entry` names" and requires a component boundary to sit only on those named, monomorphic, first-order exports. Those two sentences cannot both be implemented. The live face is an executable export, not an unnamed library: `vibec-component.md` compiles the core with `__no_entry__` but still exports `compile_cli_request(source, request) -> String`, and the wrap lifts `compile: func(source: string, request: string) -> string`. A host that follows the kind table will not call `compile`; a host that follows §2.5 expects `vibe.entry` to name it. That undoes the section's point — initialization without a flag — for the one compile-only component the rest of the document already treats as a boundary.

**Suggestion:** Add a `compile` kind (`core-export=compile_cli_request`, `component-export=compile`, hosted `compile-file` if that wrap is in scope) and pin both spellings the same way `vibe_command`/`run` are pinned. If `library` is only for uncomposed `__no_entry__` cores, drop `compile` from the §2.5 named-entry list and say the wrap, not `vibe.entry`, publishes the compile export.

mizchi:
Review round addressed in `ea6bc7d` (inline replies were refused because a pending review exists on this PR, so the per-thread answers are collected here; the threads themselves are resolved).

**Owner review on 85581fc** — all three fixed by c4e0043: `fs.read-dir` / `sh_lines` stay `string` on both lanes; a gated `field_labels` table plus `used` rows name the grant; a `compile` kind for `vibec` (`compile_cli_request` / `compile`, hosted `compile_file_request` / `compile-file`).

**Codex on `host-contract-artifact-lazy-cli.md`**
- *Idempotent close*: the raw close field is an explicit idempotent `close: func()` method on the resource; the implicit drop is a no-op after it; the component-side shim keeps a raw-handle → resource table so a second close of a copied handle returns success without touching the canonical handle table; the close-twice fixture runs on three lanes and pins the value.
- *Grant label in v2 rows*: an entry is `<label>:<interface>.<function>` (e.g. `Console:stdout.write-stream`), taken from the `used` rows; a function reached under two labels is listed once per label. Canonicalizing `Stdout` / `Console` into one authority unit is left to ADR-0088.
- *`service` in the entry protocol*: added as a `vibe.entry` row (no entry call; consumers call the interface; the requirement is the union over exports, satisfied at instantiation); preflight step 4 is skipped for it.

**Codex on `component-build-convention.md`**
- *Package components from the `index.vpkg` facade*: a package component is `vibe build --component <dir>` with the contract's declarations as the surface, implementations resolved through the loader and conformance; a loose `.vibe` stays path-only under `vibe:app`, and sitting beside an `index.vpkg` does not change that.
- *Service-wide capabilities*: a `service`'s host requirement is the union over its exports, carried as a `requires =` contract line, and any consumer importing anything from the package takes it into its own row (refused naming the missing label otherwise).
- *Map duplicates*: the inbound shim refuses a duplicate key by name; a foreign `list<tuple<string, V>>` is `Array[(String, V)]`, never `Map`; only a vibe producer's own `vibe.contract` spelling brings a `Map` back.
- *Recursive exported types*: refused by SCC with the edit.
- *Component bytes in the identity*: the package pin (`pkg:b3:`) hashes the component bytes with the extracted `.vpkg`; the contract identity (`ct:b3:`) hashes the contract text alone.
- *Continuations across the boundary*: a user effect crosses only as a tail-resumptive import (returns exactly once); stored, late or absent `resume` is not expressible, and the contract says so.
- *ABI-changing rows on `command`*: `command` refuses `Exception[E]` and `Async` on `vibe_command` (`run` stays exactly `func(string) -> string`; the frame is the failure channel); `handler` refuses `Exception[E]` and takes `Async` only with a `HostStream` body.
- *Multiple exception labels*: exactly one exception kind per export; two or more (directly or via an effectset) are refused with the edit to discharge into one payload type.
- *`Ok` wrapping*: the generated lift is `handle { Ok(body) } with Exception[E] { Throw(e) => Err(e) }`.

---

mizchi:
Codex round on `ea6bc7d` addressed in `cb4aa09`. Both findings were confirmed against the source before deciding. (Inline replies are still refused — an unsubmitted draft review by the same account blocks them — so the per-thread answers are here and the threads are resolved.)

**Align invalid-byte handling across string lowerings** — confirmed, and worse than the finding states: the raw lane does not merely differ, it silently corrupts. `vibe_read_packed_str` ends `String::from_utf8_lossy(&buf).into_owned()` (`runtime/viberun/src/main.rs`) and the node runner's `new TextDecoder().decode(..)` is non-fatal by default (`scripts/wasm_vibe_host_runner.js`), so `Fs::write_file(path, bad)` writes U+FFFD where the program had bytes and reports success. Nothing documented it. The contract is now that a `string` carries valid UTF-8 and an invalid byte string is **refused on both lanes**, naming the operation and the first bad byte offset. Arbitrary bytes cross as `Bytes` / `list<u8>` (`fs_write_bytes` / `fs_read_bytes`), which is lossless on both lanes and is the edit the refusal names. Making the component lane lossy to match would have spread the corruption instead of removing it. This is a behaviour change to two shipped runners, so it carries a migration note rather than a flag, and the conformance fixture's red test restores one runner's lossy decode.

**Verify grant labels for precompiled command rows** — confirmed: a deserialized component type names the WIT functions but never the grant labels, so the AOT path could check the function part of a v2 row and never the label part, and flipping `Console:` to `Stdout:` would pass while changing the authorization decision. A `.cwasm` row now names and digest-pins its source `.component.wasm`; the launcher verifies that digest, reads the labels from that component's `vibe.capabilities` `used` rows, and checks the function part against the image's component type. A missing, mismatched or disagreeing source is refused, with no fallback to checking the type alone. What the pin does not buy is stated rather than implied: nothing binds the image to the component cryptographically, so an image built from different code still runs — that residual is exactly what `--trust-precompiled` already means, and the pin narrows the decision to a pinned artifact instead of an unverifiable blob.

---

mizchi:
Codex round on `cb4aa09` addressed in `b441dc1`. All three confirmed against the source first. (Inline replies are still blocked by an unsubmitted draft review from the same account; answers here, threads resolved.)

**Reconcile `Console` with the one-function-per-field catalog** — confirmed, and the registry states it itself: the comment above the `Console` block reads "`Console::write_stream` is already taken by stdout", and all six `Console::*` operations lower to fields `stdin` / `stdout` / `stderr` already own. A `console` interface would repeat `stdout.write-stream` under a second name or be empty. So a provider that owns no raw field of its own **gets no catalog interface and stays a grant label**: it appears in `field_labels`, in the `used` rows and in a manifest column entry, a policy can grant or deny it, and no world imports it. The generator-diff gate row already said "minus labels that only alias another field", so the prose was the half that was wrong.

**Store contract identity outside `generated_hash`** — confirmed: `scan_package_header` throws `generated_hash must be empty or #pkg:sha1:<40hex> or #pkg:b3:<64hex>`, so a `ct:b3:` value there would make every installed component contract fail to parse before linkage. `generated_hash` keeps its package-pin domain, now explicitly over the extracted `.vpkg` **plus** the component bytes, and the contract identity moves to a new `contract_hash = #ct:b3:<64hex>` directive. Admitting that directive in `scan_package_header` / `parse_contract_header` is named as a prerequisite of the section rather than assumed.

**Add an executable entry for component-dependent programs** — confirmed. A `.vibex`'s `main` and the harness entry `vibe test` / `vibe bench` synthesize match none of the three surface-derived kinds, because they are roots, not surfaces an author writes. They now take a **derived** `wasi-command` form (`core-export=_start`, `component-export=wasi:cli/run`), decided by the dependency and never by a flag; explicit `--component <file.vibex>` stays refused, since that asks for an export surface a `.vibex` does not have. And the honest part: that producer does not exist today, so rather than routing `run` and `test` to a lane that cannot run, a component dependency is refused on **build, run and test alike** until it lands, with one message naming the package and the missing lane. Building that producer is now the first implementation item of the convention, ahead of the store and the shims.

---

docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,838 @@
...
+```
+
+The remaining interfaces are named by kebab of the registry's `Some("…")`
+label — the rest of the ten names `standard_host_provider_resource_defaults`
+lists (`lib/@vibe/compiler/core/standard_effect_policy.vibe`): `stdin`,
+`stdout`, `stderr`, `console`, `process`, `profiler`, `socket`. There is no
+`tcp`, `sh`, or `clock`. `vibe_tcp_*` is tagged `Socket`; `sh*` and `process_exit`
+share `Process`; `sleep` is tagged `Async`, which is a runtime-managed
+effect rather than a host provider, and stays out of this residual (#2832).
+`Console` is a provider this list must include. The kebab mapping
```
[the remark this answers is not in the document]
mizchi: Fixed in b441dc1, and the registry says it in its own words: the comment above the `Console` block reads "`Console::write_stream` is already taken by stdout", and all six `Console::*` operations lower to fields `stdin` / `stdout` / `stderr` already own. A provider that owns no raw field of its own now gets **no catalog interface** and stays a grant label — it appears in `field_labels`, in the `used` rows and in a manifest column entry, a policy can grant or deny it, and no world imports it. The generator-diff gate row already pinned the interface set to the labels minus alias-only ones, so the prose was the half that disagreed.

---

docs/internal/design/component-build-convention.md
```diff
@@ -0,0 +1,475 @@
...
+
+`vibe add <spec>` of a component package materializes
+`.vibe/store/@scope/pkg/` holding the `.component.wasm` and an `index.vpkg`
+**extracted from its `vibe.contract` section**. That `.vpkg` is derived like
+every other artifact here and is never edited. Two identities, both on
+ADR-0093's line: the **package pin** (`pkg:b3:`, the `require` row in the
+consumer's `index.vpkg`) hashes the component bytes together with the
+extracted `.vpkg`, so a component with different code and the same contract
+does not satisfy the pin; the **contract identity** (`ct:b3:`, the
+`generated_hash`) hashes the contract text alone, which is what a consumer
```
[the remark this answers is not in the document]
mizchi: Fixed in b441dc1. Confirmed: `scan_package_header` throws `generated_hash must be empty or #pkg:sha1:<40hex> or #pkg:b3:<64hex>`, so a `ct:b3:` value there would make every installed component contract fail to parse before linkage. `generated_hash` keeps its package-pin domain, now explicitly over the extracted `.vpkg` **plus** the component bytes, and the contract identity moves to a new `contract_hash = #ct:b3:<64hex>` directive. Admitting that directive in `scan_package_header` / `parse_contract_header` is named as a prerequisite of the section rather than assumed.

---

docs/internal/design/component-build-convention.md
```diff
@@ -0,0 +1,475 @@
...
+copied, records and variants marshalled field by field, `Exception[E]`
+re-raised from `result` (§3.1). The consumer's own artifact therefore
+declares `import acme:greeter/greeter@0.1.0` in its world.
+
+**A component dependency makes the consumer a component** (decided by the
+owner, 2026-09-17, on the safe side: a refusal can be relaxed later, a
+second linkage in the runner cannot be taken back). A core module has
+no component imports, so a program that imports a component package cannot be
+emitted as a core `.wasm`; `vibe build` of such a program refuses with the
+edit (`--component`), and `vibe run` / `vibe test` take the component lane
```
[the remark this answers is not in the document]
mizchi: Fixed in b441dc1. A `.vibex`'s `main` and the harness entry `vibe test` / `vibe bench` synthesize match none of the three surface-derived kinds, because they are roots rather than surfaces an author writes. They now take a **derived** `wasi-command` form (`core-export=_start`, `component-export=wasi:cli/run`), decided by the dependency and never by a flag; explicit `--component <file.vibex>` stays refused, since that asks for an export surface a `.vibex` does not have.

And the part the finding is really about: that producer does not exist today, so rather than routing `run` and `test` to a lane that cannot run, a component dependency is now refused on **build, run and test alike** until it lands, with one message naming the package and the missing lane. Building that producer is the first implementation item of the convention, ahead of the store and the shims.

---

docs/internal/design/component-build-convention.md
```diff
@@ -0,0 +1,502 @@
...
+| two or more exception kinds on one export (`Exception[IoError] + Exception[ParseError]`, directly or through an effectset) | refused: one `result` has one error type, and choosing either kind would lose the other; the edit is to discharge into one payload type (an enum) in the body |
+| a bare `Exception` / legacy `Error` with no payload type | refused: annotate the payload type, because `result<T, ?>` has no `?` |
+
+**The host requirement of a `service` is the union over its exports.** A
+component is one instance with one world, and every world import is
+satisfied when the instance is created, whichever export is later called. So
+a `service` that exports a pure `add` and an `Fs`-reaching `load` has an `fs`
+import that a consumer of `add` alone must still see satisfied. The derived
+contract therefore carries a `requires = <labels>` header line, the union of
+the surface's host labels, and a consumer that imports **anything** from the
```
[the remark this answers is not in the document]
mizchi: Fixed in 4ab88be. Authority is the operation (ADR-0071 / ADR-0084) and a label denotes all of them, so `requires = Fs` would have made a consumer of a pure export grant the whole filesystem to satisfy a package whose only reach is `Fs::read_file`. The line now reads `requires = Fs::read_file, Env::get` — the union of normalized operations, exactly the row a `with` clause carries — and `allows Fs::read_file` satisfies it.

---

docs/internal/design/component-build-convention.md
```diff
@@ -0,0 +1,502 @@
...
+
+The mapping is a bijection on this subset, and §8 checks it as one.
+
+## 3. Effects at the boundary
+
+An export's row is projected label by label:
+
+| label in the row | projects to |
+|---|---|
+| a host capability (`Fs`, `Env`, `Http`, .. — the registry's provider labels) | an **inline** interface named by the label, holding only the used functions of the `vibe:host` catalog (host-contract §1.3; never `import vibe:host/<label>` whole, which would demand the full surface); derived from the emitted import section, so it can never list a function the code does not reach |
```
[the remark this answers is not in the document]
mizchi: Fixed in 4ab88be — this was my own inconsistency from the previous round, introduced by the same commit that made `Console` interface-less. The inline interface is now named by the catalog interface that **owns the raw field**, not by the label: `Console::write_stream` lowers to `stdout_write_stream`, so the world gets `stdout.write-stream` and `Console` survives only as the grant label in the `used` rows and the manifest column. The alternative the text had would have asked for a `console` interface no host registers.

---

docs/internal/design/component-build-convention.md
```diff
@@ -0,0 +1,502 @@
...
+second linkage in the runner cannot be taken back). A core module has no
+component imports, so a program that imports a component package cannot be
+emitted as a core `.wasm`.
+
+**An executable or test root takes the `wasi-command` form, derived, never
+requested.** A `.vibex`'s `main` and the harness entry `vibe test` / `vibe
+bench` synthesize match none of §1.3's three surface-derived kinds, because
+none of them is a surface the author writes: they are roots. So the
+dependency decides the form, not a flag — `kind=wasi-command`,
+`core-export=_start`, `component-export=wasi:cli/run` (host-contract §2.3).
```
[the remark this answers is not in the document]
mizchi: Fixed in 4ab88be. The `wasi-command` row in host-contract §2.3 now carries both layers: a bare core module has no component export and a host calls `_start`; a **composed** one — the derived form an executable or test root takes under §6.2 — has the wrap lift `_start` to `wasi:cli/run`, and a component host calls that, since it cannot reach the nested core `_start` the preflight step named. The row also says that producer does not exist yet, so the composed column describes what it must emit rather than what ships.

---

docs/internal/design/host-contract-artifact-lazy-cli.md
```diff
@@ -0,0 +1,849 @@
...
+| **semantic conformance** | a fixture set of small programs (one per operation family: read/write/exists/read-dir ordering, `stat-token` on a non-regular path, `env.get` on an unset name, handle close-twice, **an invalid-UTF-8 byte string through a `string` parameter**, a withheld optional operation on the **core** lane) runs on both runners with byte-identical stdout and identical exit, on the linear lane | edit one runner's `read-dir` to skip the sort, and restore one runner's `from_utf8_lossy` in place of the refusal; each must fail on that runner only |
```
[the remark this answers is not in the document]
mizchi: Fixed in 4ab88be. The gate row now runs the close-twice case on the component lane as well, and its red test drops the second-close entry lookup from the component shim's handle table — the one newly introduced thing that can trap, which a two-runner row would have passed straight over.

---

docs/internal/design/component-build-convention.md
```diff
@@ -0,0 +1,502 @@
...
+
+## 9. Stability
+
+The derived WIT is public surface. A change to it is a change to the package
+version under the SemVer table in
+[../../user/reference/stable-surface.md](../../user/reference/stable-surface.md)
+(a removed or narrowed export is Major, an added one is Minor), and the
+`generated_hash` of the derived contract pins the exact surface a consumer
+was built against. Because the contract is derived, the version bump is
+checkable: the gate can diff the previous published contract against the
```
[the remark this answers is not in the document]
mizchi: Fixed in 4ab88be. §9 now pins the surface with `contract_hash` and says why `generated_hash` cannot serve: it covers the component bytes too, so an implementation-only rebuild moves it while the WIT is identical, which would report a contract change where none occurred and reject a valid Patch. The gate diffs the published contract and refuses a Patch whose `contract_hash` moved.

---

Issue #2778, opened 2026-09-14
Specialize nested generic lambdas: thread witness dictionaries for lambda binders (route 2 of #2737)


mizchi:
Route 2 of #2737, which closed on route 1 (refuse the bound rather than dispatch through a borrowed witness, PR #2746). Route 1 stopped a silent-wrong; it did not make a nested generic lambda work, and nothing tracks the part that would.

Filing it separately because it is no longer only #2737's remaining half: **it is the ending for #2745 and #2773 as well**, and those have now produced a measurement worth quoting.

## What is missing

A lambda binder gets no witness dictionary of its own. `build_gens` / `thread_dict_params` transform **top-level** generics; `rewrite_expr`'s `EFn` arm rebuilds a nested lambda with its bounds untouched and never extends `dict_binds`. So `let inner = [T: Eq](a: T, b: T) -> Bool { T::equals(a, b) }` declares a bound nothing honours, and since #2746 it is refused with a message naming the edit (move the lambda to a top-level declaration).

Making it work means two things:

1. `build_gens` treats an `EFn` carrying bounds as a generic, so the lambda gets dictionary parameters.
2. The call site materializes a dictionary the way `synth_dicts` already does for a top-level call.

## Why this is also #2745's ending

`"\{a}"` where `a`'s type is a formal a **nested** binder bound renders the tagged pointer, because the value is **erased** — one lowering serves `inner(42)` and `inner(Qt::{ .. })`. #2745 refuses it, and to do that `dtd_erased_interp_formal` has to answer "is this interpolated value an erased formal?" from a pass (`normalize`) that has **no types**. It reads syntax and carries an ad-hoc environment.

A nested lambda specialized per instantiation has no erased formal. The interpolation gets a concrete type, `interp_missing_show_type` (#1445) answers by name as it already does for a top-level generic, and **the whole approximation is deleted rather than extended**.

## The measurement that makes the case

PR #2761 fixed #2745 by enumeration. As of `dca9478`:

| | |
|---|---:|
| review rounds, each finding something real | **26** |
| distinct routes past the approximation | **63** |
| of those, over-refusals (a correct program rejected) | **7** |
| refusal fixtures | 70 |
| green-control rows | 96 |

It is not tapering, and the *shape* of the remainder is visible in the record:

[...]

Issue #2397, opened 2026-08-29
Library code cannot abort with a message: `assert` traps silently, `__rt_oob_abort` is out of reach


mizchi:
Found while doing #2362, which asked for a broken invariant to "say so" rather than hang. The bound landed; the saying-so did not, because library code has no way to do it.

## What is available

Measured on `8de462a`, from ordinary library code (no effect row):

| how | outcome |
| :--- | :--- |
| `assert(false)` | `RuntimeError: unreachable`, exit 1 — **no message** |
| `Array::get(xs, 99)` on a length-3 array | `Array::get: index 99 out of bounds for length 3`, then traps |
| `throw("...")` | carries the message, but puts `Exception` on the function's row and every caller's |

The middle row is what a broken invariant should look like. It is produced by `__rt_oob_abort` (#2199), a codegen-level runtime helper that assembles `<prefix><idx> out of bounds for length <len>` and traps — reachable only from builtin bodies, and only on the linear lane (`backend_body.vibe`: the gc lane "has no `__rt_oob_abort` helper").

So today the choice from a library is: terminate silently, or color the public API.

## Why it matters

CLAUDE.md ranks silent-wrong worst and puts actionability next: a diagnostic "leads with the edit that fixes it". A bare `unreachable` with no line and no name is the third state — it does not lie, but it says nothing, and it is indistinguishable from a compiler bug. That is precisely the complaint #2199 fixed for bounds checks; the same argument applies one layer up.

Concretely, `lib/@vibe/core/hashmap.vibe`'s new guard traps without being able to say *"MutMap: the table has no free slot; `size + tombs` accounting has drifted"*, which is the one sentence that would send a reader to the actual cause.

## Shape of a fix

A builtin along the lines of `abort(msg: String) -> Never` (or `panic`), with:

- an effect row of `None`, so it does not color its caller — it terminates rather than returning, so there is no authority to track
- the message on the same stream `__rt_oob_abort` uses, in the same framing
- a body on **both** lanes, or an explicit row in `scripts/builtin_parity_classification.tsv` saying why not

`assert` could keep its current shape and gain a two-argument sibling rather than changing, so nothing existing has to move.

Worth settling alongside: whether `assert`'s own silent trap should carry at least the failing expression's source text, which the compiler has at the call site.

## Not urgent

Nothing is silently wrong today — the states that reach these traps are unreachable through the public APIs. This is about what a failure says when the invariant does break, which is when a person is already confused.

[...]

Issue #2219, opened 2026-08-22
The bootstrap bump landed: delete the legacy assert-abort block recognizer; decide whether abort provenance moves out-of-band


mizchi:
Implementation status at the 2026-09-15 audit: the existing comment reports work on `claude/compassionate-feynman-b4vwmx`; it is not present on main 77000b46e. Review and land that prepared work before marking this issue complete.

**The precondition fired.** This was written as "after the next bootstrap bump"; that bump has happened, so the first half is work to do now rather than a wait. Title and body updated 2026-09-06 — see the verification comment.

Follow-up consolidating the declined halves of the #2202 trap-echo work (PR #2213, Codex rounds 3 and 5).

## Where #2213 leaves things

`vibe test`'s condensers (`vt_fail_detail` in `scripts/vibe_test.sh`, `condense_test_trap` in `runtime/vibe`) suppress the `trap: RuntimeError: unreachable` echo after an assert_eq failure when either recognizer fires:

1. **Marker** — the generated abort's closing line `assert failed: aborting` (`lower_assert_eq`, `normalize/desugar_trait_dict.vibe:10157`), adjacent to the trap. Robust to multiline rendered values.
2. **Legacy text fallback** — a complete consecutive `assert_eq failed`/`expected:`/`actual:` block adjacent to the trap. Exists **only** because the committed seed predated the marker, and `vibe test` compiles with the seed by default.

Both are in-band: a program that *deliberately* prints the exact marker (or the exact block) immediately before an unrelated trap gets that trap suppressed. This was reviewed twice and accepted twice as the cost of keeping the default lane fixed — the forgery requires printing an exact sentinel and then trapping on purpose, which is self-sabotage rather than an accident. It cannot be closed with any in-band text scheme, by construction: assert output and user output share the same stream.

## Ready now: delete the legacy fallback

**The committed seed emits the marker.** Verified 2026-09-06 against `bootstrap/seed.json`'s pinned artifact, `seed/cfg-spans-emission-2026-09-04` (`sha256=3acf3c3f…`, adopted in #2519): the string `assert failed: aborting` is present in `bootstrap/seed/compiler.wasm`, and an `assert_eq(1, 2)` failure run through `scripts/vibe_test.sh` on the seed reports the condensed form with no trap echo. Reason 2's justification — "the committed seed predates the marker" — no longer holds.

[...]

PR #2868, opened 2026-09-17
Production trap linemap, rowless abort, nested witness forwarding

mizchi:
## Summary
- Production RC compiles emit compact `vibe.linemap` so OOB traps at `Array::get`/`set`, `Bytes::get`/`set`, and `String::byte_at` still print `<op>: index <idx> out of bounds for length <len>` and also report an editable `path:line` (#2199). Stripped mapping degrades to a wasm frame. wasm-gc stays the engine-native OOB trap (gated limitation).
- `abort(msg)` is a `None`-row builtin on both lanes: print then trap, no `Stdout` (#2397). One-arg `assert` stays silent.
- Test condensers no longer recognize a consecutive `assert_eq failed` / `expected:` / `actual:` block; only `assert failed: aborting` suppresses the trap echo (#2219).
- Nested generic lambdas overlay and **forward** enclosing trait dictionaries, so `T::equals` / UFCS / `T::to_string` inside `let inner = [U](...)` use the outer witness instead of an unresolved table call (#2778 review).

## Test Plan
- [x] `tests/gates/early/run.sh` 4e: five OOB ops + stripped mapping
- [x] `lib/@vibe/compiler/tests/abort_message_test.vibe` (linear + gc compile, message + trap, silent `assert`)
- [x] `scripts/vibe_test_smoke.sh` marker-less block reports `trap:`; marker still suppresses
- [x] `lib/@vibe/compiler/tests/nested_lambda_witness_test.vibe` compile + runtime pins for outer `T` under inner `[U]` / `[U: Eq]`
- [x] `bash scripts/size_ratchet.sh <stage2.wasm>` after compact-mapping rebaseline
- [x] CI compiler-gate / early 4e on this PR's stage2


mizchi:
## Codex review: all eight findings resolved

Six threads went outdated when main's #2858 moved `lib/@vibe/cli/user_dispatch.vibe` to `lib/@vibe/compiler/user_dispatch.vibe`. Outdated is not the same as answered, so each was re-checked against the new location rather than closed on the move.

**Already addressed at the new location** (verified, not assumed):

| Finding | Where it holds now |
|---|---|
| P1 unstable-surface gate | `compile_user_artifact` is the single door (`reject_unstable_entry_surface` → compile → `reject_unstable_closure_surface`); the test lane spells both gates inline around its own compile |
| P1 GC test/bench selectors | `user_backend_kind("VIBE_TEST_BACKEND")` / `("VIBE_BENCH_BACKEND")`; the `"mvp"` literal sits inside the `backend_kind == "mvp"` guard, not a hard-code |
| P2 `VIBE_BREAK_FILE` | `runtime/vibe:445` — `envp+=("VIBE_BREAK=$brk" "VIBE_BREAK_FILE=$(basename "$src")")` |
| P2 stream-body adapter | `runtime/vibe:831` resolves `vibe_http_p3_body_stream_adapter.component.wasm` when the plan's line 6 says the handler takes a stream, with an actionable message naming the build command |

**P1 serve `__no_entry__` DCE — does not reproduce.** Measured: `vibe serve fixtures/serve_handler_smoke.vibe --no-run` produces a 2054-byte component whose exports include `handler` and whose data section carries the whole handler body (`x-token: secret`, `200`, `content-type: text/plain`, `unauthorized`). The `handler` `SLet` survives DCE under that sentinel.

**Three that were real and are now fixed:**

- **P1, single-file diagnostics.** A throw became `[]` — empty stdout in text mode, `[]` in JSON, exit 0 either way. The recoverable half already degrades inside the collector, so what escapes is the checker or a query breaking; this lane sells "empty output = clean", and spending that sentence on a check it could not run is the silent-wrong failure the design policy ranks above crashing. It is now an error diagnostic and exit 1. `vibe diagnostics` keeps the swallow deliberately: that verb is deprecated and frozen at exit 0, and unfreezing it is a separate decision.

- **P2, dependency origins.** Measured on an entry importing a dep whose failing line holds 9 bytes of Japanese:

  ```
  before  "range":{"start":{"line":2,"character":15},"end":{"line":2,"character":27}}
          "message":"binding type mismatch for a local let: expected Int, got String"
  ```

  Entry line 2 is `export fn main() -> Int {`, 25 bytes, so characters 15..27 run off its end, and nothing said the error was in another file. Now the range is synthetic (this document has no honest span for another file's error, and converting against the wrong text answers a plausible number rather than failing), the message keeps the origin path, and `data` carries `originPath` / `originLine` / `originByteColumn` — named BYTES, because converting needs that file's text. Only a caller that names its own file gets this, so the single-file lanes are untouched and the entry's own prefixed diagnostic still slices its literal.

- **P2, string diagnostic ends.** Exactly right about `String::length(s)` being the decoded length. The end is now recovered in `locate_type_error`, the one place holding both the offset and the source text, by scanning to the matching close — the lexer's own span. On `let a: Int = "\n\t\\x"`: `line 1:14-20` → `line 1:14-23`. It can never shorten a range, since escapes only make the source longer.

Each fix was red-checked: the guard mutated, the specific new assertion observed failing with the pre-fix value, and each mutation verified to land on its own (`assert` traps and masks later tests in the same file).

---

Issue #2065, opened 2026-08-17
async: let a TaskGroup task await a host future/stream (retire the Async-entry mixing guard)


mizchi:
Split out of #1342 (scope item 3 of 4), so that issue can close on what it actually delivered.

## The guard, and why it exists

An entry whose row carries `Async` gets a synthesized **tail-resumptive** boundary handler. A program that also uses **suspend-class** `Async` handling (`TaskGroup::spawn_suspend`) then mixes two conventions for one effect, and the ADR-0076 guard rejects it. That is a clear diagnostic, not a silent miscompile — the two supported shapes are:

- a row-free entry + `TaskGroup` (the library discharges `Async` itself), or
- an `Async`-row entry with no suspend-class handlers.

The guard itself is in good shape: #1342 already made it position-independent and keyed on the boundary that is actually *injected* (`fixtures/err_async_boundary_mixed_convention.vibe`, `err_async_boundary_mixed_operand.vibe`, `async_boundary_user_sleep_test.vibe` pin all three halves).

## What retiring it needs (revised 2026-08-24 — the original D2 framing was disproved by measurement)

This body originally said the forward path is ADR-0089 D2's third `park_kind`. Two rounds of measurement (recorded in the comments) revised that:

1. **The guard is a spelling artifact, not a semantic wall** — flipping it was measured working, then reverted pending the real blockers.
2. After #1727's gaps landed (#2081, 50e9fde), **the first wall is evidence-migration eligibility for `Async` closure values** (ADR-0076 追記34): a spawn-free `TaskGroup::run` hits it before suspend eligibility, before the flip, and before any scheduler work.

So the working order is the one in the 2026-08-18 comment: eligibility for `Async` closure values first, then suspend eligibility, then the guard flip, then the D2 scheduler work (third `park_kind`: take the region tag from the waitable slice and move the task onto `sp_same_region`) for actual host-waitable awaiting.

Two stale pointers to fix when this lands: the guard text (`linked_compile.vibe`) still says "need #1727 … tracked in #2065" (that half is done), and `docs/wasip3-effect-alignment.md` still records the superseded "#1727 → #2065" ordering.

## Why the sleep half is relevant

[...]

PR #2902, opened 2026-09-19
#2065 wall 1: admit a first-order row-variable callee in the evidence pass

mizchi:
## What this is

0.2.0 work chosen to not collide with the in-flight 0.1.0 release. It lands the first of the three rungs of #2065's "wall 1", by porting a rule the sibling suspend pass has shipped since #1536.

## The wall

`edp_append_effect_irrelevant_row_fns` (`lowering/effects/evidence/evidence.vibe`) bailed unconditionally when a callee's declared row mentioned a row VARIABLE:

```
} else if edp_row_labels_have_var(labels) { () }
```

So the callee never entered `pure_fns`, the migration failed to plan, and the ADR-0076 追記34 V1 guard raised its hard error — the one that says the effect "has closure values whose row carries it … but its evidence migration is not fully eligible".

## The port

The suspend pass drew the identical line (`scps_row_has_var`, `suspend.vibe:675`) and relaxed it per-callee in #1536 via `scps_callee_first_order` (`:556`). `edp_callee_first_order` is that rule on the evidence side, with the argument written there: when every declared parameter is annotated and mentions no function type, no argument can carry a row, so the row variable can only be instantiated to the EMPTY row. The call cannot perform the effect whatever the call site looks like — which is exactly the claim `pure_fns` membership makes.

`scps_ty_has_fn` / `scps_ty_list_has_fn` move to the shared `queries` package, so both row-variable rules are built on one copy rather than two that must be kept in step.

**Stricter than the suspend rule in one place, deliberately.** Suspend reads parameters only; this also refuses a declared RETURN type that mentions a function type. Both passes ask the same question of a call, but this pass's guard is specifically about row-carrying closure VALUES, and a callee that hands one back is a shape nobody has measured against the guard. Refusing it keeps the classification absent rather than wrong, and costs nothing for the callees this unblocks — they return a bare type parameter.

## What this does NOT do

**It does not unblock TaskGroup.** `TaskGroup::run[T, rg, e](body: (TaskGroup[rg, e]) -> T with e)` takes a function-typed parameter, so it is not first order. Reaching it needs the call-site argument-inertness rule (`scps_callee_fn_args_inert`, `suspend.vibe:609`) as well — and even then, the interesting TaskGroup programs hit wall 2 (suspend-class eligibility) behind that. `fixtures/err_async_rowvar_higher_order_refused.vibe` pins this boundary with `TaskGroup::run`'s shape named in its header.

It also does not touch the mixing guard, `park_kind`, or any scheduler code.

## Verification

Measured on a stage2 built from this branch with `VIBE_REGEN_MODULE_SOURCE=1` — without it the build silently reuses the old flat module source and answers about a compiler that does not contain the change.

The refusal is invisible on the `vibe_test.sh` lane: measured, it reports `ok` for `fixtures/err_async_boundary_mixed_convention.vibe`. The lane that sees it is the one `tests/gates/late/run.sh` uses, and it was proven able to fail before any measurement was trusted.

| case | before | after |
|---|---|---|
| `async_first_order_rowvar_boundary_test.vibe` (new positive) | REFUSED | **COMPILED → 42** |
| `err_async_rowvar_higher_order_refused.vibe` (new negative) | REFUSED | REFUSED |
| `err_closure_value_evidence_ineligible.vibe` | REFUSED | REFUSED (after repair) |
| `err_effect_handle_replay_removed.vibe` | REFUSED | REFUSED (after repair) |
| `err_async_boundary_mixed_convention.vibe` (control) | REFUSED | REFUSED |
| `async_sleep_boundary_test.vibe` (control) | 42 | 42 |

**Full unit corpus: 1316/1316 passed.** A differential sweep of 238 effect/async fixtures — compiled *and executed* under both compilers, answers compared — shows exactly one difference, the intended one.

New gate **128/128** (late lane, registry row added) is a PAIR on purpose, and proven able to fail on both halves with the mutations proven to have landed: feeding the positive half the higher-order fixture fails it, feeding the negative half the first-order fixture fails it. `check_gate_registry.sh`, `check_gate_portability.sh`, `check_fixture_snapshots.mjs`, `check_fixture_execution.sh` and `check_documented_decls.sh` all pass.

## Two existing pins this broke, and why they had to change

Both declared a first-order row-variable callee, so the port admitted it, the migration planned, and each program started compiling and answering correctly — leaving the fixture pinning nothing. The first was predicted by reading the code; the second the sweep found.

- `err_closure_value_evidence_ineligible.vibe` — `fn via(n: Int) -> Int with e`, began answering 8.
- `err_effect_handle_replay_removed.vibe` — `fn opaque(n: Int) -> Int with e`, began answering 3.

Both repaired by giving the callee a function-typed parameter, which is what keeps it genuinely un-seeable — the condition each fixture exists to hold, rather than an incidental property of how it was written. Red-tested: reverting to the first-order form makes the fixture compile again.

## The diagnostic now names its culprits

The old wording offered three candidate causes and for the measured shapes none is the one to act on — the decisive contributor is a declaration nothing calls:

```
effect Async has closure values whose row carries it (type-directed evidence,
ADR-0076 追記34), but its evidence migration is not fully eligible --
the pass cannot see through `via`, whose declared row has a row variable and
which takes or returns a function. Give it a concrete row, or make its
parameters and return type mention no function type.
The closure VALUES here are the Async-row function parameter of
`takes_async_fn`, which counts even if it is never called -- drop it, or drop
its row, if it is not needed.
Replay fallback is not available for closure-value programs
```

The per-callee condition is factored into one predicate (`edp_rowvar_callee_withheld`) that both the appender and the reporter read, so the message cannot start naming callees the pass did not withhold.

## Four stale pointers retired

#2065 named three; a fourth was in a fixture header. #1727 landed in #2081 and #2065's own comment retracts the ordering, but `linked_compile.vibe:3930` (a user-facing diagnostic), its doc twin at `:2048`, `wasip3-effect-alignment.md:317` and `err_async_boundary_mixed_convention.vibe` all still told the reader to wait on it.

## Measurement notes

`documented_decls_floor.txt`: `queries.vibe` 16 → 17 for the moved doc block, `evidence.vibe` 75 → 78 for this PR's three new blocks, and a new row for `evidence_test.vibe` at 2. `suspend.vibe` **keeps** its floor of 81 — it was at 82, so the move does not drop it below. (A review pass had predicted this would need lowering to 80; the ratchet run says otherwise.)

Perf: `selfcompile` +0.08% memory, +0.05% code; no drift ≥±2% in the 77 other tracked series; output checks 10/10.


mizchi:
## Verification, as promised in the description

All measured on a stage2 built from this branch with `VIBE_REGEN_MODULE_SOURCE=1` (`_build/selfhost/generations/bytes-capacity-2026-09-15_a381b0d/stage2.wasm`), through the refusal-visible lane `tests/gates/late/run.sh` uses — not `vibe_test.sh`, which reports `ok` for `fixtures/err_async_boundary_mixed_convention.vibe` and so cannot see these refusals at all.

### The port

| case | before | after |
|---|---|---|
| `async_first_order_rowvar_boundary_test.vibe` | REFUSED | **COMPILED → 42** |
| `err_async_rowvar_higher_order_refused.vibe` | REFUSED | REFUSED |
| `err_closure_value_evidence_ineligible.vibe` | REFUSED | REFUSED (after repair) |
| `err_effect_handle_replay_removed.vibe` | REFUSED | REFUSED (after repair) |
| `err_async_boundary_mixed_convention.vibe` (control) | REFUSED | REFUSED |
| `async_sleep_boundary_test.vibe` (control) | 42 | 42 |

**Full unit corpus: 1316/1316 passed.**

### Gate 128/128 is proven able to fail

Both halves, with the mutations proven to have landed rather than silently matching nothing:

- positive half fed the higher-order fixture → `FAIL: async_first_order_rowvar_boundary_test.vibe did not compile`
- negative half fed the first-order fixture → `FAIL: err_async_rowvar_higher_order_refused.vibe compiled`

`check_gate_registry.sh`, `check_gate_portability.sh`, `check_fixture_snapshots.mjs`, `check_fixture_execution.sh`, `check_documented_decls.sh`: all ok.

### The new diagnostic

The old wording offered three candidate causes, and for this program none of them is the one to act on — the decisive contributor is a declaration nothing calls:

```
effect Async has closure values whose row carries it (type-directed evidence,
ADR-0076 追記34), but its evidence migration is not fully eligible --
the pass cannot see through `via`, whose declared row has a row variable and
which takes or returns a function. Give it a concrete row, or make its
parameters and return type mention no function type.
The closure VALUES here are the Async-row function parameter of
`takes_async_fn`, which counts even if it is never called -- drop it, or drop
its row, if it is not needed.
Replay fallback is not available for closure-value programs
```

## Three CI failures, and what they were

All three were real and all three were mine. Worth recording because the second one is a hole in how I verified, not just a bug.

1. **`compiler-gate (late)`** — `err_closure_value_evidence_ineligible` compiled. Predicted from reading the code before the first push; repaired by giving its row-variable callee a function parameter.

2. **`unit-tests (shard 2/8)`** — two regressions I did not find. I had run a differential sweep: 238 effect/async fixtures compiled *and executed* under both compilers, answers compared, exactly one difference (the intended one). That was sound work scoped to the wrong corpus — this repo's unit corpus is 1316 files across `fixtures/`, `lib/**/*_test.vibe` and `bench/`, and both regressions were in `lib/`:
   - `evidence_test.vibe` asserted the old behaviour at the unit level, its builder declaring `fn side(n: Int) -> Int with e` — the first-order shape. Now a pair over that builder and a higher-order twin.
   - `prelude_split_bytes_test.vibe` holds a **copy** of `err_effect_handle_replay_removed.vibe` as a source string, so repairing the fixture left the copy behind. Synced, and verified byte-identical to the fixture body rather than eyeballed; its doc comment now says it is a copy, so the next person changing that fixture is told where the second one lives.

   The fix for the process, not just the finding: the full unit runner is what I ran before this push, instead of a hand-picked subset.

3. **`compiler-gate (contracts core)`** — adding `///` blocks turned `evidence_test.vibe` into a documented file with no floor row. Row added at 2; `evidence.vibe` also tightened 75 → 78 for this PR's three new blocks. Only the rows my own change moved.

## One correction to the description

An earlier draft of this work claimed it would unblock TaskGroup. It does not, and I would rather say so here than have a reviewer find it: `TaskGroup::run[T, rg, e](body: (TaskGroup[rg, e]) -> T with e)` takes a function-typed parameter, so it is not first order. Reaching it needs the call-site argument-inertness rule (`scps_callee_fn_args_inert`) ported to the evidence side too, and then wall 2 behind that. `fixtures/err_async_rowvar_higher_order_refused.vibe` pins that boundary with `TaskGroup::run`'s shape named in its header.

---

Issue #2905, opened 2026-09-19
`Console::read_char()` compiles to a module that does not validate, and the build reports success


mizchi:
Found while fixing #2903 by auditing the raw-ABI shim's name lists for other spellings with the same gap.

## Repro

```vibe
fn main() -> Int allows Console {
  Console::read_char()
}
```

Compiles "successfully" — a `.wasm` is produced and the compiler reports nothing. The artifact then fails to load:

```
$ bash scripts/run_wasm_vibe_host_runner.sh --invoke main rc_console.wasm
CompileError: WebAssembly.instantiate(): Compiling function #3 failed:
  not enough arguments on the stack for call (need 1, got 0) @+416
```

and independently:

```
$ node -e "new WebAssembly.Module(require('fs').readFileSync('rc_console.wasm'))"
WebAssembly.Module(): Compiling function #3 failed:
  not enough arguments on the stack for call (need 1, got 0) @+416
```

The sibling is fine, which is what makes the cause legible. With a host stub returning the raw byte 65:

| program | result |
|---|---|
| `Stdin::read_char()` | **65** — correct |
| `Console::read_char()` | **module does not validate** |

## Cause

`Console::read_char` is a registered builtin — `builtin_registry.vibe:243`, `CtFn(reg_p0(), CtInt, Some("Console"))`, so zero parameters returning `Int` — and it appears **nowhere** in `lib/@vibe/compiler/codegen/expr/compile_call.vibe`:

```
$ grep -o 'Console::[a-z_]*' lib/@vibe/compiler/codegen/expr/compile_call.vibe | sort -u
Console::read_stream
Console::write_char
Console::write_err_char
```

Its three siblings are all there. `Stdin::read_char` is in both the return-side shim group (`compile_call.vibe:6619`) and the sorted ladder-membership list (`:539`); `Console::read_char` is in neither, so it falls through to the generic resolved-call path and lands on an import whose arity does not match — "need 1, got 0" is a zero-argument call reaching a one-argument import.

This is the shape #1460 already described for the write side:

> the `Console` spellings share the imports, so they inherit the gap too -- they are registered into the func table the same way, with no shim.

That round fixed `Console::write_char` and `Console::write_err_char`. `Console::read_char` was not in it, and unlike the write side it does not merely produce a wrong byte — it produces an artifact that cannot load.

## Why P1 rather than P0

The program never returns a wrong answer; it fails to load. That is triage's "it crashes", not "silently wrong".

[...]

Issue #2903, opened 2026-09-19
`vibe.sleep` receives a tagged Int and no host decodes it — every `sleep(ms)` sleeps 2× or ½ the requested time


mizchi:
Found while doing #2832 checkbox 1 (inventory the async host imports from the actual emitters).

## The defect

`sleep(ms)` lowers to the core import `vibe.sleep : (i64) -> ()`, and the value the guest passes is the **RC-tagged** integer `ms << 1`. No shipped host decodes it that way, so nobody sleeps for `ms`.

Observed directly — a stub host that records the raw `i64` and does nothing else, so no decoding, no heuristic, nothing to misread:

```
$ node probe.mjs sleep1000.wasm
--- imports ---
  vibe.sleep : function
--- values passed to vibe.sleep ---
  raw i64 = 2000   >>1 = 1000   >>2 = 500   &3 = 0
```

`sleep(1000)` hands the host **2000**. The correct decode is `>> 1` (the RC lane's 1-bit tag — `common_base.vibe:3930` "the RC lane's tagged `n<<1`", and CLAUDE.md's `Int` section).

| host | decode | `sleep(1000)` actually sleeps |
|---|---|---|
| `runtime/viberun/src/main.rs:3449` | `ms as u64` → 2000 | **2000 ms** (2×) |
| `scripts/wasm_vibe_host_runner.js:2814`, ABI auto-detected raw | `Number(2000)` | **2000 ms** (2×) |
| same, `VIBE_IMPORT_ABI=tagged` | `2000 >> 2n` → 500 | **500 ms** (½) |
| `tools/async_host/src/main.rs:27` | `(tagged_ms: i32)`, `>> 2` | would not link — **`i32` against a declared `(i64) -> ()`** |
| `tools/async_host/src/sleep_host.rs:47` | ignores the value | yields once |
| `tools/async_host/src/concurrency.rs:33` | `_ms` unused | yields once |

Timed end to end on the node runner, same module, harness overhead 63 ms measured separately:

```
run under ABI=tagged   570 ms wall  (~507 ms slept)
run under ABI=raw     2069 ms wall (~2006 ms slept)
```

## Why it is P0 rather than P1

It is silently wrong in the triage sense: the program completes, returns the right value, and reports nothing. Only the duration is wrong — and the direction depends on an environment variable, so two runs of the same binary disagree with each other.

[...]

Issue #2832, opened 2026-09-15
0.2.0: specify and verify async host runtime ABI independently of Wasmtime


mizchi:
Extracted from #1346 so the synchronous host contract can be completed before 0.1.0 without pulling async into that release.

- [ ] Inventory future/stream/sleep imports, `VIBE_ASYNC_FUTURES`/`VIBE_ASYNC_STREAMS`, payload layouts, ownership, cancellation and completion rules from actual emitters.
- [ ] Separate the runtime-neutral contract from Wasmtime configuration and Component Model canonical ABI details. Distinguish raw core imports from interfaces expressible in WIT.
- [ ] Align real-source WIT-derived imports (#2064), TaskGroup host waits (#2065), outbound HTTP (#2066) and suspend state (#1537) with that contract.
- [ ] Add focused node/viberun conformance and negative tests for unsupported imports and invalid lifecycle operations; no backend may silently reinterpret the same import.

Existing async functionality stays explicitly unstable in 0.1.0. This is a runtime contract task, not authorization to replace the default backend with WasmFX (#2221).

mizchi:
First checklist item, measured from the emitters rather than from the docs.

## The emitted surface

Every async host import is in core module `vibe`, emitted by `entry/source_compile/wasi_only/component_codegen.vibe`:

| core import | role |
|---|---|
| `vibe.host_future_get` | anonymous future — ADR-0089 step 4, the original single `get-future` |
| `vibe.host_future_get$<name>` | named future (ADR-0089 (c)) |
| `vibe.wit_future_get$<name>` | WIT-derived future — the #2064 shape |
| `vibe.host_future_wait` | the settle/wait the entry boundary drives |
| `vibe.host_stream_get$<name>` | readable end of a named stream |
| `vibe.host_stream_read` | pull one byte |
| `vibe.host_stream_close` | drop the readable end early |
| `vibe.sleep` | blocking sleep; **raw ms, not tagged** (#2903) |

The `$<name>` suffixes are fixed at compile time: `host_future_named` / `host_stream_named` require a **string literal**, because the argument becomes the core import name and then the component import name.

## The guest-side contract (checker/builtins_async.vibe)

| builtin | type | effect |
|---|---|---|
| `host_future_get()` | `Future[Int]` | pure — creating the pair is eager; only `await` carries `Async` |
| `host_future_named(String)` | `Future[Int]` | pure, same reason |
[...]

PR #2907, opened 2026-09-19
#2903: settle `vibe.sleep`'s argument as a raw millisecond count

mizchi:
Fixes the P0 in #2903 (which was auto-closed by #2904 — that PR *documented* the defect as part of the #2832 inventory but did not fix it; the defect was live in the compiler at close time).

## The defect is not the one the issue describes

#2903 framed this as a choice between two candidate fixes. It is not a choice: **option 1 (the guest untags) was already the design.** The RC raw-ABI shim in `codegen/expr/compile_call.vibe` removes the `n << 1` tag before any host import call, every core module declares `vibe.abi: host_import_abi=raw`, and both shipped core runners obey that declaration correctly:

| host | behaviour under `raw` |
|---|---|
| `runtime/viberun/src/main.rs:3449` | `ms as u64` — pass-through |
| `scripts/wasm_vibe_host_runner.js` | `decodeHostInt` → `Number(value)`, unshifted, once `detectHostImportAbi` reads the section |

The **emitter** was the party violating its own declaration. And what existed was not one defect but **a pair on opposite sides that cancelled** on the one lane anybody ran.

### 1. The shim missed a name

`cc_raw_abi_shim_applies` listed `sleep` but not `sleep_blocking` — which had **zero** occurrences in `compile_call.vibe` despite being a registered builtin (`builtin_registry.vibe:301`). Both names resolve to the same `vibe.sleep` import (`linked_compile.vibe:10996`), and the injected `__entry_settle` pays a suspend sleep debt by calling `sleep_blocking` (`:4222`, `:4255`). So an ordinary `sleep(ms)` inside `allows Async` reached the host through the one uncovered name, still tagged.

### 2. The component adapter decoded it back

The adapter's `sleep` body applied `comp_emit_decode_i64_value(…, tag_mode)` — `>> 1` under RC — to its ms argument. That **cancelled defect 1** on the componentized lane, and in the same stroke halved any *direct* `sleep`, which was already shim-covered and so already arrived raw. That half was latent only because an async entry never takes that route.

So fixing (1) alone makes the component lane **worse**. Measured end to end with `viberun` on `let run: () -> Int with Async` sleeping 300ms:

| state | elapsed |
|---|---:|
| pre-fix (both bugs, cancelling) | 318ms |
| shim fixed only | **165ms** ← fails the existing gate's `≥ 0.8×D` (240ms) band |
| **both fixed (this PR)** | **314ms** ×3 runs, 0ms spread |

FS lane, the issue's own repro, with `VIBE_RC=0` as the control that proves the expectation is not simply "whatever the compiler emits":

| | `VIBE_RC=0` | `VIBE_RC=1` |
|---|---:|---:|
| before | 1000 | **2000** |
| after | 1000 | 1000 |

## Changes

- `compile_call.vibe` — `sleep_blocking` added to the shim predicate **and** to `cc_ladder_names`. Adding a name to the predicate alone is **inert**: the list is binary-searched and is how the predicate is reached. The doc comment records this, so the next name added does not hit the same trap.
- `component_codegen.vibe` — the adapter's sleep body no longer re-decodes the ms. `tag_mode` still governs the future/stream values below it, whose core-side producers are *not* shim-covered.
- `tests/gates/early/run.sh` — new gate `27g/27`, plus its registry row.
- `docs/internal/design/async-host-contract.md` — the `vibe.sleep` section said "unsettled, see #2903"; it now states the rule, since there is one.

## The gate

Asserts the **value** handed to `vibe.sleep`, not wall-clock: a duration has no output to diff and a timing band flakes on a shared runner, while the declaration makes the contract exact. The observer decodes nothing, so it cannot paper over a tag the way the runner's own heuristic does.

Two rows — direct `sleep_blocking(1000)`, and `sleep(1000)` under `allows Async` — and **each was proven to fail independently** on a pre-fix stage2 and pass on a fixed one, rather than relying on one row that would mask the other:

```
RED (pre-fix, async row in isolation): FAIL: sleepa.vibe under VIBE_RC=1 handed vibe.sleep '2000', want 1000.
GREEN (fixed):                         raw-ABI shim covers sleep_blocking ok (1000 on both lanes, direct and via Async)
```

## Two caveats about the evidence

**The Codex review did not run** — it posted that it hit its usage limits, so this PR has had no automated code review, only CI.

**CI green is not evidence for the adapter half.** The two lanes this PR touches are asymmetric in coverage: the core half is covered by gate 27g and the unit corpus, but the component half's only real guard (`scripts/test_async_sleep_component_gate.sh`) **runs in no CI workflow** — it is a standalone `pkf` task, and `wasi-p3-gate` runs only the guarantee gate. A green pipeline here would have looked identical at 165ms. Filed as #2908 rather than fixed here, so the CI change is reviewed on its own terms; happy to fold it in if preferred — it would need `VIBE_P3_GATE_REQUIRE_TOOLS=1`, since these gates `require_or_skip` with exit 0.

## Deliberately not in scope

- **#2905** (filed, P1): `Console::read_char` appears nowhere in `compile_call.vibe` and produces a module that does not validate. Found while auditing sibling spellings.
- `VIBE_IMPORT_ABI=tagged` beats a module's own `vibe.abi` declaration (`wasm_vibe_host_runner.js:2470`) and then decodes through a 2-bit-tag heuristic the RC lane stopped using. It is the `½` row in the issue's table, but **no emitter ever declares `tagged` and nothing in the repo sets that variable** — a dead lane, not a live divergence. Recorded in the contract doc's "still not pinned" list; it is the "fifth implementer" hazard the issue names.

## Verification

All on a stage2 built from this tree with `VIBE_REGEN_MODULE_SOURCE=1`, carrying both halves:

**1316/1316 unit-test files passed** · gate 27g Red+Green, both rows independently · component timing 314ms ×3 under `viberun` · `vibe fmt --check` on both changed compiler files · perf report: no drift ≥±2% across 77 tracked series, output checks 10/10 · ten structural lints (doc-classification, doc-path-citations, gate portability / registry / wiring / lane-coverage / self-tests, task-inputs, task-env-sanitized, documented-decls) · pre-commit review-lint · CI + cli-install workflows green · `mergeable_state: clean`.

Numbers in [this comment](https://github.com/mizchi/vibe-lang/pull/2907#issuecomment-5742043167). An earlier corpus run was stopped deliberately — it was measuring an intermediate tree that had only the shim half, which is the "which compiler answered?" trap CLAUDE.md warns about.


mizchi:
Verification complete. Everything the PR body listed as "still running" is now done, all against a stage2 built from this tree with `VIBE_REGEN_MODULE_SOURCE=1` and **carrying both halves** of the fix.

## Unit corpus

```
[unit-test-runner] 1316/1316 active unit-test files passed
[unit-test-runner] ok
```

## The component lane, end to end

`let run: () -> Int with Async` sleeping 300ms, composed with the in-tree composer and run under `runtime/viberun`. This is the measurement that matters, because it is the one no automated check in this repo performs:

| state | elapsed |
|---|---:|
| pre-fix — both defects, cancelling | 318ms |
| shim fixed only | **165ms** |
| **both fixed (this PR)** | **314ms, 314ms, 314ms** (3 consecutive runs) |

The gate this would trip is `ELAPSED >= 0.8 * D` = 240ms. 165ms fails it; 314ms clears it with margin, and the run-to-run spread is 0ms.

## Gate 27g on the final build

```
[compiler-gate] raw-ABI shim covers sleep_blocking ok (1000 on both lanes, direct and via Async)
```

Both rows were each proven to fail independently on a pre-fix stage2 before being trusted here — the async row alone, with the direct row removed so it could not be masked:

```
RED (pre-fix, async row in isolation):
  FAIL: sleepa.vibe under VIBE_RC=1 handed vibe.sleep '2000', want 1000.
```

## Two caveats I would rather state than leave implicit

**1. The Codex review did not run.** It posted `You have reached your Codex usage limits for code reviews`, so this PR has had **no automated code review** — only CI. Combined with the coverage gap below, the adapter half rests on the hand measurement above rather than on anything in the pipeline. That is why I measured it three ways (value at the import, component wall-clock, full corpus) instead of once.

**2. CI green is not evidence for the adapter half.** The two lanes this PR touches are asymmetric in coverage: the core half is covered by gate 27g and the corpus, but the component half's only real guard (`scripts/test_async_sleep_component_gate.sh`) runs in no workflow — filed as #2908. A green pipeline here would have looked identical at 165ms.

## Full check list

Green on the final stage2: 1316/1316 unit corpus · gate 27g (Red+Green, both rows independently) · component timing under `viberun` · `vibe fmt --check` on both changed compiler files · perf report, no drift ≥±2% across 77 tracked series, output checks 10/10 · ten structural lints (doc-classification, doc-path-citations, gate portability / registry / wiring / lane-coverage / self-tests, task-inputs, task-env-sanitized, documented-decls) · pre-commit review-lint · CI + cli-install workflows · `mergeable_state: clean`.

---

mizchi:
**Correction to this PR's body: the "coverage gap" section was wrong.** The fix itself is unaffected — this is about a claim I made alongside it.

I wrote that `scripts/test_async_sleep_component_gate.sh` "runs in no CI workflow" and that "a green pipeline here would have looked identical at 165ms". Both are false. That gate ran on **this PR**, in the `wasi-p3-gate` job (job `105898684811`, head `910fe95`):

```
[async-sleep-component-gate] sleep-only: 42 in 310ms (delay 300ms)
[async-sleep-component-gate] rc0: 42 in 311ms
[async-sleep-component-gate] mixed: 42 in 313ms (sleep and future overlapped; serialized would be ~600ms)
```

310ms, against the 314ms I measured locally. **CI would have caught the 165ms regression.**

The error: I grepped for the filename under `.github/` and `scripts/pkfire/`, found nothing, and stopped. `test_wasi_p3_guarantee_gate.sh` is an aggregator — phase A runs thirteen component gates including this one at `:107`, and the job invokes it with `VIBE_P3_GATE_PHASES: "async,http,stdin"` and `VIBE_P3_GATE_REQUIRE_TOOLS: "1"`. I asked whether a filename appears in a workflow, when the question was whether the gate executes. Corrected inventory: **25 of 27** gate scripts run in CI; the 2 that do not are a gc-backend probe and #1239's build-coordinator gate, neither an async-component gate.

The silent-skip concern was already handled too, and `ci.yml:1559` says so explicitly — "Missing tools are FAILURES (no silent skip)" — including the documented reason this job sits outside `ci-required`. I had read that YAML and missed the comment.

#2908 is closed as invalid with the full measurement.

**What still stands, unchanged:** the two-defect diagnosis, the 318 → 165 → 314ms sequence, the FS-lane before/after (`2000` → `1000` under `VIBE_RC=1`, with `VIBE_RC=0` as control), gate 27g and its independent Red proofs, and 1316/1316. What does **not** stand is my characterisation of the evidence around it: this change had better automated coverage than I credited it with, not worse. I was wrong in the direction of alarm, which is still wrong.

---

PR #2909, opened 2026-09-19
#2828 rung 2: the launcher's capability grants reach `perform?`

mizchi:
Advances #2828 (ADR-0088 rung 2). Does **not** close it — rung 3, the persisted `BindingLock`, still needs the `apply` phase that does not exist.

## The bug

ADR-0088's L2 table existed and was always empty. `optional_perform_artifact_resolution` returns its `resolution` argument verbatim for any non-test artifact, and both production call sites passed `array_empty()`, so every optional capability resolved `NotGranted` regardless of what the run was granted.

Measured before the change, `perform? Fs::read_file` on a real file:

| flags | answer |
|---|---|
| *(none)* | `NotGranted` |
| `--allow-stdout` | `NotGranted` |
| `--allow-fs --allow-stdout` | `NotGranted` |
| `--deny-fs --allow-stdout` | `NotGranted` |

`--allow-fs` and `--deny-fs` were indistinguishable. A control ruled out an environment problem: the same program with a *non-optional* `Fs::read_file`, under the same flags, read the file and printed its contents.

This is also the mechanical reason **`Errored(E)` was unreachable**. `opq_expr` builds the `EHandle` whose arm constructs `$vibe_attempt_errored` only on the `Granted` branch, and nothing outside a test artifact ever resolved to `Granted`. It needed a non-empty table, not a new ABI.

## The change

Rung 1 already computed the granted set at the CLI (`host_granted_from_flags`). This threads it to the lowering. `optional_grants_from_flags` turns the flags into `[(provider, "Granted")]` rows; each hop gains a `_grants` arity while the existing arity delegates with an empty table, so **no existing caller changes**, including the ~20 in the split tests.

Fail-closed by construction: `opq_resolution` falls back to `"*"` → `NotGranted`, so a granted provider needs a row and a denied one needs none. A provider nobody considered cannot be granted by omission.

Measured after, same program:

| flags | answer |
|---|---|
| *(none)* | **`Granted`** — ambient authority, matching rung 1's preflight |
| `--allow-stdout` | `NotGranted` — an allow-list is a list |
| `--allow-fs --allow-stdout` | **`Granted`**, and the call runs |
| `--deny-fs --allow-stdout` | `NotGranted` |
| `--allow-fs --deny-fs` | `NotGranted` — deny still beats allow |

Note the first row is a deliberate semantic change: a no-flag run has full ambient authority, so `perform?` now resolves `Granted` where it previously always said `NotGranted`. That is consistent with rung 1, where a no-flag run already passes the L3 preflight unchanged.

## The gate tests an answer, not a wiring

The first build of this rung had all nine hops threaded, type-checked at every hop, a clean build and a running control — and was **completely inert**. `optional_perform_artifact_resolution` has two call sites, and only the split one was wired; a single-file `vibe run` does not satisfy `module_split_usable`, so it takes the non-split fallback, which still passed `array_empty()`.

Nothing structural separates "wired" from "wired and read". Only running a `perform?` program does. So the four new cases in `check_capability_preflight.sh` assert resolved answers, and the red input is the **real pre-rung-2 compiler** rather than a synthetic mutation — it fails with `got: NOTGRANTED`, exit 1.

Fixing the self-test surfaced a second defect in it: its green control fell back to scanning the generations glob for any artifact carrying rung 1's refusal string, and a rung-1-only stage2 satisfies that. The new cases therefore ran against a compiler predating them and reported a failure saying nothing about the tree. Rung 2 adds behaviour, not text, so no string identifies it and guessing cannot be made safe. The control now refuses and names what it needs.

## Verification

Compiler: `_build/selfhost/generations/bytes-capacity-2026-09-15_fce9080/stage2.wasm`, artifact-verified against a marker red-tested to 0 before the build.

- `perform?` grant table: 5 rows, all correct (above)
- `scripts/check_capability_preflight.sh`: **10 passed, 0 failed** (6 original + 4 new)
- Red test vs. the pre-rung-2 stage2: **fails correctly**, exit 1
- `scripts/check_capability_preflight_test.sh`: 2 passed, 0 failed
- `scripts/check_gate_portability.sh`: exit 0
- Unit corpus: **1318/1318, exit 0**
- `vibe fmt --check`: clean on all five source files

The corpus and the `early`/`mid`/`late` lanes ran green on the pre-rebase base; the gate and self-test were re-run on the rebased artifact. CI covers the full set here.

https://claude.ai/code/session_0122a6eBQhEYutwU7R7sUP9y

---
_Generated by [Claude Code](https://claude.ai/code/session_0122a6eBQhEYutwU7R7sUP9y)_


mizchi:
### `compiler-gate (selftests)` is red on `fce9080`, and I could not reproduce it locally

**What is failing:** one case, the rung-2 one this PR adds.

```
capability-preflight: FAIL: --allow-fs must make perform? Fs::read_file Granted; got: NOTGRANTED
capability-preflight: 9 passed, 1 failed
```

The other nine pass, including every rung-1 case, and the remaining 36 checks on this head are green. The `ci-required` failure is the aggregate of this one. `set -e` aborted the lane at the gate, so the gate's own self-test never ran — the FAIL is the gate with the lane's compiler, not the self-test's deliberate red case.

**It does not reproduce here.** I rebuilt stage2 the way CI does — `VIBE_PREBUILT_MODULE_SOURCE=lib/@vibe/compiler/_cli_adapter_module_source.vibe bash scripts/generations.sh build` — and the gate passes **10/10** against that artifact.

**Eliminated by measurement, not by reasoning:**

| hypothesis | eliminated by |
|---|---|
| stale generated bundle | CI ran `ensure-generated` and regenerated all 5 artifacts at 16:20:07, before the stage2 build |
| the bundle does not carry rung 2 | a freshly regenerated local bundle contains `optional_grants_from_flags`, `no_grants`, `compile_file_fs_mode_rc_grants` |
| a non-RC lane (`mvp`/`shadow`) drops the grants | `VIBE_RC` unset in both environments |
| the compile cache ignores the grant table | deny-then-allow over one shared `VIBE_BUILD_CACHE_DIR` answers `NotGranted` then `Granted`, correctly |
| a stale stage2 restored from cache | the log shows a **miss** on `stage2-v2-<fingerprint>-…`, and that fingerprint is byte-identical to this tree's |

**What blocks fixing it:** the failure message reports the *answer* and not the artifact that produced it, so the CI log cannot distinguish "this compiler has no rung 2" from "it has rung 2 and the `--allow-*` path broke". Those need opposite fixes. I also cannot re-run the job — the token gets `403 Resource not accessible by integration` on `rerun-failed-jobs`.

**What I pushed** (`5baf707b`): the gate now prints, on that failure only, the resolved compiler path, its size and mtime, the runner, and — the decisive part — the **ambient** answer with no capability flags at all. With rung 2 present, ambient authority makes that `Granted`; without it the table is empty and every arm is `NotGranted`. So the next run's log says which of the two bugs this is instead of leaving it ambiguous.

Red-tested both directions: the block fires against a pre-rung-2 stage2 and correctly concludes "does not carry rung 2", and does not fire against a compiler that does.

Not a fix, and I am not calling it a flake — I have no evidence for that. It converts an unreproducible failure into a diagnosable one, and the push re-runs CI, which is the re-run I cannot trigger directly. I will act on the result.


---

Issue #769, opened 2026-07-07
lib/@vibe/compiler/tests: 6 files fail to parse standalone (compiler/index.vibe, entry/compiler/index.vibe facades)


mizchi:
## Background

`scripts/selfhost_unit_test_runner.sh --scan` shows 6 `lib/@vibe/compiler/tests/*_test.vibe` files failing with parse errors that point into compiler-internal facade files, not the test files themselves:

- `ast_binary_test.vibe` — `lib/@vibe/compiler/ast_binary.vibe: line 165:1: expected ';', ',' or '}' in struct`
- `cache_probe_cli_prepare_batch_bench_test.vibe` — `lib/@vibe/compiler/index.vibe: line 3:1: unexpected token in import path: import`
- `selfhost_s5_stage2_test.vibe` — `lib/@vibe/compiler/index.vibe: line 3:1: unexpected token in import path: import` (same root cause)
- `cli_test.vibe` — `lib/@vibe/compiler/entry/compiler/index.vibe: line 1:1: unexpected after 'export': ..`
- `selfhost_s5_test.vibe` — `lib/@vibe/compiler/entry/compiler/index.vibe: line 1:1: unexpected after 'export': ..` (same root cause)
- `coverage_selfhost_suite_lib_test.vibe` — `lib/@vibe/compiler/coverage_selfhost_suite_lib.vibe: line 4:8: unknown name: Json::JNum`

None of these 6 are in `scripts/selfhost_unit_test_allowlist.txt`.

## Hypothesis

`lib/@vibe/compiler/index.vibe` and `lib/@vibe/compiler/entry/compiler/index.vibe` look like directory-import/export facade files whose `import .. / export ..` sugar the loader only resolves correctly when the *package* (not the bare file) is loaded through its normal directory-boundary resolution path — the unit-test runner may be compiling them as a plain standalone file, hitting syntax the parser doesn't expect outside that context. `coverage_selfhost_suite_lib.vibe`'s `Json::JNum` looks like a stale reference to a pre-`Ok`/`Err`-builtin-Result-era JSON API (see #760 gap 2).

## Ask

Root-cause whether this is a genuine parser/loader gap (files that are only ever meant to be entered via bundle/manifest resolution, never compiled standalone) or a real regression, and either fix the compile path or document/allowlist-exclude these test files with a comment explaining why they can't run through the standalone unit-test harness.

Discovered via a `--scan` audit while consolidating stdlib packages into `@vibe/core` (#766).

[...]

PR #2916, opened 2026-09-20
#2905 second half: tag `Console::read_char`'s result on the RC lane

mizchi:
Reopens and finishes #2905. #2911 fixed the half the issue reported; verifying that fix showed the other half was still open — and it is the worse one.

## What #2911 fixed, and what it left

The issue reported: `Console::read_char` reserved no import, so the module **built clean and failed to validate** (`not enough arguments on the stack for call (need 1, got 0)`). #2911 fixed that, correctly, and with a better diagnosis than my issue had — the cause was import *reservation* (`linked_compile.vibe:9645`), not the shim name lists I blamed.

But the module now **loads and returns the wrong byte**. Measured on a stage2 built from `3260d45`, host stub returning a fixed raw byte and decoding nothing:

| host byte | `Console::read_char` | `Stdin::read_char` |
|---:|---:|---:|
| 65 (`'A'`) | **32** (space) | 65 |
| 200 | **100** | 200 |
| 1 | **0** | 1 |

Exactly `host >> 1` at every value. Under `VIBE_RC=0` both answer correctly, which is what identifies the RC lane as the subject.

## Cause

`Console::read_char` **aliases** the same import as `Stdin::read_char` — `linked_compile.vibe:11052` maps it to `stdin_read_char_idx` — and that import speaks **untagged** integers, so the RC lane must tag the result. The shim arm that does this (the #705 group in `compile_call.vibe`) listed `Stdin::read_char` and not the Console spelling. #2911 supplied the reservation; nothing supplied the tag.

So the reservation fix **traded a loud failure for a silent wrong answer**. CLAUDE.md's first pillar is *never be silently wrong*, and triage ranks P0 = silent-wrong above P1 = it crashes — so the issue was not done when it closed.

## The name goes in two lists

`cc_ladder_tests_name` is an early-out to the resolved-call fallback, so an arm is **unreachable** for a name `cc_ladder_names` does not carry. Adding only the arm is inert. That is the #2903 trap, now recorded in a comment beside the fix so the fourth `Console::` alias does not repeat it a fourth time.

This family has diverged four times: #1460 (`Console::write_char`/`write_err_char`), #2903 (`sleep_blocking`), #2905 first half (reservation), and this. #2911 added `check_import_reservation.py` — a real guard, whose red test even names `Console::read_char` — but it checks **reservation**, not the RC result-tag lists, so it could not see this one.

## Gate 27h

Asserts the **value**, against a host stub that decodes nothing so it cannot paper over a missing tag, across three bytes and both RC lanes. Two controls keep a passing row meaningful:

- `Stdin::read_char` (the sibling that was always right) must agree
- `VIBE_RC=0` must already pass — so the expectation is not "whatever the compiler emits"

Module validation is checked **separately** from the value, because the original defect was a module that would not load, and that reads as a trap rather than a wrong number.

Both rows are proven able to fail, **for different reasons**:

```
stage2 @ dbdea5a (pre-#2911):  FAIL: console.vibe under VIBE_RC=0 built a module
                                     that does not validate (#2905)
stage2 @ 3260d45 (post-#2911): FAIL: console.vibe under VIBE_RC=1 returned '32'
                                     for host byte 65 (#2905)
```

## Deliberately not changed

`Console::read_char` is also absent from `ts_int_returning_builtin_name` (#769's static Int-returning list, which keeps `__to_string` off its string-pointer heuristic). I measured before touching it: `"\{Console::read_char()}"` and the `Stdin` spelling both render at length 2, so there is no defect there. Latent, not manifest — and fixing by analogy without a measurement is what this repo's gate rules warn against.

## Verification

**1319/1319 unit-test files passed** · gate 27h Green on the fixed stage2 and Red on the preserved pre-fix binary (both rows, independently) · `vibe fmt --check` · gate registry / wiring / portability · pre-commit review-lint · perf: no drift ≥±2% across 77 series, output checks 10/10 · **`compiler-gate (early)` green in CI**, which is the lane carrying gate 27h.

Numbers in [this comment](https://github.com/mizchi/vibe-lang/pull/2916#issuecomment-5747082861).

**One CI lane is red and it is not this diff's**: `compiler-gate (selftests)` failed on a pkl package-registry fetch (`pkfire@0.14.2` → unreachable `release-assets.githubusercontent.com`), failing before any red case ran. It passes locally on this tree and main is green on it. I cannot re-run it (403) and will not make that self-test tolerate a network outage. Detail [here](https://github.com/mizchi/vibe-lang/pull/2916#issuecomment-5747045946).


mizchi:
**CI red on `compiler-gate (selftests)` — not this PR's.** (`ci-required` then failed as a consequence of it; same root cause, not a second problem.)

The failing self-test is `scripts/check_task_env_sanitized_test.sh`, and it failed on the **unmutated tree**, before any of its red cases ran:

```
[task-env] FAIL: a system binary does not run inside a pkf task (#2877)
pkf: evaluate .../Taskfile.pkl: metadata fetch failed:
     https://pkg.pkl-lang.org/github.com/mizchi/pkfire/pkfire@0.14.2
http fetch failed: NetworkError("InvalidUrl(\"https://release-assets.githubusercontent.com\")")
[task-env-test] FAIL: the gate does not pass on the unmutated tree; the red cases below would prove nothing
```

A network fetch of the pkl package metadata failed, so `pkf` could not evaluate `Taskfile.pkl` at all. Note the gate's headline reports the *symptom* (`a system binary does not run inside a pkf task`) while the cause is one line below — taking the headline at face value would send a reader chasing `LD_LIBRARY_PATH`.

Why it is not this PR's:

- **The diff does not touch it.** Three files: `codegen/expr/compile_call.vibe`, `tests/gates/early/run.sh`, `tests/gates/registry.tsv`. Nothing near `Taskfile.pkl`, `defaults.env`, or pkl package resolution.
- **It passes locally on this exact tree**: `[task-env-test] ok (3 red cases verified to land; RED 4: red)`.
- **The base branch is green on it.** Main's run for `3260d45` (run 6472) concluded `success`, and that is the commit this branch sits on.

**Every other lane is green, including the one this PR changes:**

| lane | result |
|---|---|
| `compiler-gate (early)` — carries the new gate 27h | **success** |
| `compiler-gate (mid)` / `(late)` / `(preflight)` | success |
| `wasi-p3-gate` | success |
| `compiler-gate (selftests)` | **failure** (the pkl fetch above) |

So gate 27h passes in CI, not only locally — which was the open question, since it compiles programs and instantiates them under node and is more environment-sensitive than a pure source check.

There is no fix to port: `pkg.pkl-lang.org` redirecting to an unreachable `release-assets.githubusercontent.com` is outside the tree.

**I cannot re-run the job** — `rerun-failed-jobs` returns `403 Resource not accessible by integration` for this token, so the one confirming re-run the process calls for is unavailable to me, and I would rather say that than imply it happened. I am also not going to make that self-test tolerate a registry outage: masking a network failure inside a gate is exactly the "silence is not safety" failure this repo's #2248 / #2252 rules exist to prevent, and it belongs to a different subsystem than this PR.

What this needs is a re-run by someone with the permission, or — if it recurs — a CI-infra fix so `pkf` resolves `pkfire@0.14.2` from the cached `~/.cache/pkfire` rather than the network on every job. Keeping the PR watched until it is green rather than treating it as finished.

---

mizchi:
Verification complete. Everything the PR body listed as pending is done, all against a stage2 built from this tree with `VIBE_REGEN_MODULE_SOURCE=1`.

## Unit corpus

```
[unit-test-runner] 1319/1319 active unit-test files passed
[unit-test-runner] ok
```

## Gate 27h, both directions

**Green** — stage2 carrying the fix:
```
[compiler-gate] Console::read_char ok (host byte returned intact on both lanes, matching Stdin::read_char)
```

**Red** — the pre-fix stage2, preserved before the rebuild overwrote it:
```
[compiler-gate] FAIL: console.vibe under VIBE_RC=1 returned '32' for host byte 65 (#2905).
```

A note on that second one, because it nearly did not survive: the rebuild reuses the generation directory named after the HEAD the build was *launched* from, so the pre-fix `stage2.wasm` at `_3260d45/` was about to be replaced by the post-fix one under the same name. I copied it aside first. Both directions are therefore re-runnable on demand rather than being a result I once pasted into a commit message — which is the distinction #2248 is about, and it also means the directory name no longer describes what is inside it.

Earlier, against the **pre-#2911** compiler, the same gate fails on its other row for a different reason — `built a module that does not validate` — so the two rows are independently load-bearing rather than always failing together.

## CI

Every lane green except `compiler-gate (selftests)`, which failed on a pkl registry fetch unrelated to this diff — detailed in the comment above, including why I can neither re-run it (403) nor paper over it.

The lane that matters for this change:

| lane | result |
|---|---|
| `compiler-gate (early)` — carries gate 27h | **success** |
| `compiler-gate (mid)` / `(late)` / `(preflight)` | success |
| `wasi-p3-gate` | success |

Gate 27h passing in CI was the open question — it compiles programs and instantiates them under node, so it is more environment-sensitive than a pure source check. It passes there too.

## Full check list

1319/1319 unit corpus · gate 27h Green + Red (both rows, independently) · `vibe fmt --check` · gate registry / wiring / portability · `check_task_env_sanitized_test.sh` locally (the CI-red one) · pre-commit review-lint · perf report, no drift ≥±2% across 77 series, output checks 10/10 · `compiler-gate (early)` green in CI.

---

mizchi:
This is my regression — #2911 landed the reservation half and I closed #2905 as completed. I've reopened it and recorded the three-site table there. Verified the diagnosis independently against main at 3260d45f rather than from the PR body: `cc_ladder_names` (526), the RC tag arm (6527) and the `__to_string` classifier (2243) all list `Stdin::read_char` and none lists `Console::read_char`, while `linked_compile.vibe:11052` maps both to `stdin_read_char_idx`. The fix is right and the two controls in row 27h (the `Stdin` sibling, and `VIBE_RC=0`) are what make it more than "whatever the compiler emits".

**The red CI is not this diff.** Both failures reduce to one cause: `compiler-gate (selftests)` → `check_task_env_sanitized_test.sh`, which reports `[task-env-test] FAIL: the gate does not pass on the unmutated tree`. The underlying line is a dependency fetch, not a gate finding:

```
pkf: evaluate .../Taskfile.pkl: metadata fetch failed: https://pkg.pkl-lang.org/github.com/mizchi/pkfire/pkfire@0.14.2
http fetch failed: NetworkError("InvalidUrl(\"https://release-assets.githubusercontent.com\")")
```

`ci-required` then fails only on `compiler-gate-lanes did not succeed` — every other required row is `success`. Main's last run (3260d45f) was green through the same lane an hour earlier, and nothing in this diff touches `Taskfile.pkl`, pkfire, or the network. It died at install before any test body ran, so it's re-runnable; I get 403 on `rerun-failed-jobs`, so someone with write access needs to kick it.

Two residual sites I checked so they don't turn into a third round, neither needing a change here:

- `ts_int_returning_builtin_name` (2243) — not a live defect. The `__to_string` heuristic is `(v >>u 32) >= 64 && ... < memory_size`; for a byte `0..255` the high word is `0`, and for EOF `-1` it is `0xFFFFFFFF`, which fails the upper bound. Every value this builtin can return already takes the integer path.
- `gc/backend_body.vibe:1830` / `:2075` — correctly absent; `builtin_parity_classification.tsv:9` classifies `Console::read_char` as `linear-only / host-import`.

---

Issue #1962, opened 2026-08-16
Unify direct host builtins with perform and provider semantics


mizchi:
## Problem

The same conceptual operation can be exposed both as `perform Fs::ReadFile(path)` and as a direct `Fs::read_file(path)` builtin. If the builtin bypasses lexical handlers, the two spellings do not share semantics and users must know an implementation detail.

## Decision

Choose and implement one model:

1. public host helpers lower to the same operation algebra as perform, or
2. host primitives become internal provider APIs callable only by the standard provider.

Do not retain a public split where only the perform spelling is interceptable.

## Acceptance criteria

- Mock/record/replay handlers observe every public operation spelling consistently
- Internal host primitives cannot be invoked without provider authority
- Existing capability APIs have a documented migration path
- Checker, WIT lowering, runtime, and tests use one operation identity

Extracted from #1496.

mizchi:
Still blocked on **unifying direct host builtins with perform/provider semantics**. Architecture leftover.

mizchi:
## Measured: the direct spelling silently bypasses the handler, and the two answers differ in the same program

The issue states the problem ("If the builtin bypasses lexical handlers, the two spellings do not share semantics"). Measured against a stage2 built from main today, it does, and the failure is silent.

Same handler, same file, same run — only the spelling differs:

```vibe
fn main() -> Unit allows Stdout + Fs::read_file {
  let x = handle {
    Fs::read_file("/tmp/real2.txt")      // direct
  } with Fs {
    Fs::ReadFile(_p) => resume("MOCKED")
  }
  println(x)
}
```

| spelling inside the handler | answer |
|---|---|
| `perform Fs::ReadFile(path)` | `MOCKED` — intercepted |
| `Fs::read_file(path)` | `REAL-FILE-CONTENTS` — **handler bypassed, real filesystem read** |

The file genuinely contains `REAL-FILE-CONTENTS`, so the second row is not a fallback or an error path: the handler was installed, matched nothing, and the program went to the host.

### Why this is worse than "two spellings, one interceptable"

A mock handler that does not mock is the silent-wrong category CLAUDE.md's triage ranks P0, above crashing. Someone writing

```vibe
handle { run_the_thing() } with Fs { Fs::ReadFile(p) => resume(fixture_for(p)) }
```

[...]

PR #2919, opened 2026-09-20
#1962 follow-up: skip the rewrite walk where no handler names a host operation

mizchi:
## What

Follow-up to #2918, which landed before this could ride along. The perf report on that PR flagged the cost of the pass itself, measured against `10b621879`:

| series | value | Δ |
|---|---:|---:|
| prelude cold | 959 MiB | **+2.01% ⚠️** |
| selfcompile memory | 988 MiB | +1.96% |
| prelude warm | 622 MiB | +1.56% |

Paid by programs the pass never changes.

## Cause

`dhhc_expr` rebuilds **every** node it visits — `ETuple(items) => ETuple(dhhc_exprs(items, ops))` allocates a fresh node and a fresh array whether or not anything was rewritten — and it runs twice, once in the checker and once in the codegen prelude. Over the compiler's own sources, which contain no `handle` naming a host operation at all, that is a full `Expr` clone of the whole program for nothing.

## Fix

`desugar_inspect_calls` next door already solves this, and its comment calls the guard "load-bearing rather than cosmetic" — I should have carried that property over with the rewrite shape. Same guard here: an allocation-free pre-scan over `expr_children_push` / `expr_child_at` answers *"does this statement contain a `handle` whose arms name a host operation?"*, and only a statement that does gets the rebuilding walk.

One detail that is correctness, not performance: `EHandle` and `EMatch` are spelled out in the scan rather than left to the generic child walk, because **that walk does not descend into arm bodies** — the same reason `collect_push_elem_shapes` spells them out a few hundred lines up. Without it, a handler nested inside a match arm would scan as "nothing here" and its file would be skipped silently. That is a wrong answer, not a slow one.

## Verified

On a stage2 built from this tree:

- `fs_handled_direct_test.vibe` — 3/3, so the direct spelling is still intercepted.
- `desugar_handled_host_calls_test.vibe` — 7/7, including the four negative controls.
- `fs_test.vibe`, `index_import_test.vibe`, `fs_readdir_test.vibe`, `fs_effect_test.vibe` — unchanged.

**Whether the 2% is actually recovered is for this PR's own perf report to say.** This change does not claim it; the number in the table above is the one to compare against.

Refs #1962

https://claude.ai/code/session_0122a6eBQhEYutwU7R7sUP9y

---
_Generated by [Claude Code](https://claude.ai/code/session_0122a6eBQhEYutwU7R7sUP9y)_


mizchi:
Perf verdict, which this PR's description said it would not claim in advance: **the guard recovers essentially all of the regression.**

Reading the percentage alone would mislead. This PR baselines against `ef154efa6` — main *with* the unguarded pass — so its `−1.80%` is a recovery, not an improvement over the original state. Chaining through the absolute numbers gives the comparison that matters:

| series | pre-pass (`10b621879`) | with the pass (#2918) | with the guard (this PR) | net vs pre-pass |
|---|---:|---:|---:|---:|
| prelude cold | 940.1 MiB | 959 MiB (+2.01% ⚠️) | 941 MiB | **+0.10%** |
| selfcompile memory | 969.0 MiB | 988 MiB (+1.96%) | 971 MiB | **+0.21%** |
| prelude warm | 612.4 MiB | 622 MiB (+1.56%) | 614 MiB | **+0.25%** |

The pre-pass column is recovered from #2918's own reported percentages against its `10b621879` baseline, not measured separately.

The residual 0.10–0.25% is the pre-scan itself, which still walks every expression — it just does so without allocating a replacement node per visit. That is the floor for a pass wired into both the checker and the codegen prelude, and it is small enough that I am not proposing to chase it further.

CI on this head: 26 passed, 9 running, 0 failures.

---

Issue #2914, opened 2026-09-19
`vibe grep --where` traps (out of bounds) on a repo-wide sweep; the untyped tier handles the same corpus


mizchi:
## What

`vibe grep`'s **typed tier** (`--where`) traps on a repo-wide query. The untyped tier answers the same query over the same corpus.

```
$ vibe grep --pattern '__to_string($(x:exp))' --where '$x : Array[_]' lib
viberun: error while executing at wasm backtrace:
    0:   0x3a6c - <unknown>!<wasm function 28>
    ...
Caused by:
    wasm trap: out of bounds memory access
```

## Isolation

Measured 2026-09-19 on a stage2 built from `claude/sweet-sagan-6gy3cz` at 97ff33361. Same pattern throughout; the only variable is `--where`.

| query | corpus | result |
|---|---|---|
| untyped | `lib` | **ok — 2551 matches, 213s** |
| typed | `lib` | **`wasm trap: out of bounds memory access`** |
| typed | `lib/@vibe/compiler` | **timeout** (>150s) |
| typed | `lib/@vibe/compiler/checker` | ok, 0 matches |
| typed | `lib/@vibe/core` | ok, 0 matches |
| typed | `lib/@vibe/lsp` | ok, 0 matches |
| typed | one file | ok |
| untyped | one file / small dir | ok |

So it is **not corpus size** — the untyped tier walks all of `lib` fine. It is the typed tier, which degrades to a timeout at package scale and then traps at tree scale.

## Why this matters more than an ordinary crash

`AGENTS.md` makes `vibe grep`'s typed tier the answer to a specific class of question, and names it as the difference from moongrep / ast-grep:

> **文法だけで止まらない**のが moongrep / ast-grep との差: filter は checker の答え (推論型・effect row・解決済み名・型エラー) で書く

and the CLI section sets the standard the tool is held to:

> **判定に使える** — 「このファイルはコンパイルが通るか」に CLI 単体で答えられる。答えられないなら、それは診断の穴であって呼び出し側の作法の問題ではない

A typed repo-wide sweep is the thing this tier exists for, and it is exactly the shape that fails. The workaround — sweep package by package and aggregate — is the "ワークアラウンドを覚えて先へ進まない" case that section forbids, which is why this is an issue rather than a note in my scrollback.

## What it blocked

[...]

Issue #2912, opened 2026-09-19
`__to_string` on an `Array::map` result renders a pointer, silently


mizchi:
## What

`__to_string` on an `Array::map` result renders a **pointer** instead of the array, and nothing reports an error. The array itself is correct — only the rendering is wrong.

```vibe
test "p" {
  let xs = [1, 2, 3]
  inspect(__to_string(Array::map(xs, (v: Int) -> Int { v * 2 })), "[2, 4, 6]")
}
```

```
actual:   300
expected: [2, 4, 6]
```

`vibe check` is silent. `300` is an allocation address — it varies with what ran before it in the same file.

## Measured boundary

Stage2 built from `claude/sweet-sagan-6gy3cz` at 1565c28b4, single-file `vibe test` lane:

| program | result |
|---|---|
| `__to_string(Array::map(xs, (v) -> v * 2))` | **pointer** |
| `__to_string(Array::map(xs, (v: Int) -> Int { v * 2 }))` | **pointer** |
| `let ys = Array::map(..); __to_string(ys)` | **pointer** |
| `let ys: Array[Int] = Array::map(..); __to_string(ys)` | `[2, 4, 6]` |
| `__to_string(Array::map(ss, (v: String) -> String { .. }))` | **pointer** |
| `__to_string(Array::get(Array::map(..), 1))` | `4` |
| `__to_string(Array::length(Array::map(..)))` | `3` |
| `__to_string(Array::filter(xs, (v) -> v > 1))` | `[2, 3]` |
| `__to_string([2, 4, 6])` | `[2, 4, 6]` |

So:

- **Annotating the LAMBDA does not help; annotating the BINDING does.** The renderer needs the value's declared type and does not get it from `Array::map`'s inferred return.
- **The array is fine.** `Array::length` gives 3 and `Array::get(.., 1)` gives 4. Only the whole-array render is wrong.
- **`Array::filter` is the control and renders correctly** in the same file, unannotated. So it is not "any Array-returning builtin".
- Not `Int`-specific: `Array[String]` does the same.

## Not a regression

Identical on the pre-change stage2 (`fce9080`) — pointer both ways. I measured that explicitly because I first read it as fallout from #2900 and it is not. Filing it separately rather than folding it in.

## Why P0 rather than P1

[...]

Issue #2773, opened 2026-09-13
#2745's erased-formal refusal approximates types in a pass that has none, and twenty-six review rounds found sixty-three routes past it


mizchi:
Filed from PR #2761, where the same defect has now been found **twenty-six rounds in a row**. Every instance was real and every fix was correct; the point of this issue is that finding them one at a time is not converging. #2761 landed at the author's direction with the enumeration where it stands; **this issue is now the tracker for the remainder.**

## What the refusal is

`"\{a}"` where `a`'s type is a formal a **nested** binder bound renders the value's representation — the tagged pointer. #2745 refuses it. To do that, `dtd_erased_interp_formal` has to answer "is this interpolated value an erased formal?", and it lives in `normalize`, which has **no types**. So it reads the value's syntax and carries an ad-hoc environment.

## The measurement

**Twenty-six rounds, sixty-three distinct routes, seven of them over-refusals** (a correct program rejected — the other direction of the same gap). Each route measured before and after against an explicitly built stage2; each pinned by a refusal fixture plus a green-control row. 70 refusal fixtures, 96 green rows. The per-route table is in #2761's description.

The rate did not fall, and the *shape* of the remainder is legible in the record:

- **Seven times** closing one spelling of a pair left its sibling open, and the sibling was the next round's finding: `xs[0]` vs `Array::get`, a call inline vs through a name, a constructor inline vs through a name, an aggregate argument written vs named, `if` vs `match`, local vs top-level, array vs map.
- **Four times** a descent read a body in the *outer* environment instead of shadowing the binder it introduced. That direction produces **over-refusals**, which the `_refused` fixtures structurally cannot see — they reach us essentially only through review.
- **Five refusal fixtures have lost their isolation.** They used a local generic call as "the initializer this walk declines", which is what made the route named in each header the only thing that could refuse the file. Round 15 taught the walk to classify that call, so those five now refuse for two reasons; each header records it.
- **Trying to restore that isolation produced a new silent-wrong.** Reaching for a `T`-typed expression this pass cannot classify, I tried `let xs = []` plus `Array::push(xs, a)`. It was not opaque; it was route 54, printing the pointer on the commit that had just passed every gate in the PR.

## Open routes, carried rather than fixed

Reported against `dca9478` and **not verified by me**:

| route | claim | status |
|---|---|---|
[...]

PR #2920, opened 2026-09-20
#2912: refuse an unrenderable array instead of printing its address

mizchi:
Closes the remaining half of #2912 — the floor the issue asked for: *"whatever the fix, the floor should be that it refuses rather than renders an address."*

> **Verified.** Self-build exits 0, 1321/1321 unit corpus, early gate lane 39/39, gate Red+Green. Numbers in [this comment](https://github.com/mizchi/vibe-lang/pull/2920#issuecomment-5748199999).
>
> This branch briefly also carried a second fix (#2914(b)); it hit a design conflict and was **reverted** — [detail here](https://github.com/mizchi/vibe-lang/pull/2920#issuecomment-5748278603). The PR is #2912 only, and none of #2912's files were touched by that revert.

## The defect

`__to_string` on an array whose element type does not resolve fell through to `compile_call`'s general arm — a runtime heuristic that guesses string-pointer vs integer from the bit pattern. An array pointer takes the integer branch and prints as a decimal address.

| program | before | after |
|---|---|---|
| `Array::map(xs, (v: Int) -> Int { v * 2 })` | `[2, 4, 6]` | `[2, 4, 6]` |
| `Array::map(xs, (v) -> v * 2)` | **`296`** | **refused** |
| `Array::filter(xs, (v) -> v > 1)` | `[2, 3]` | `[2, 3]` |
| `let ys: Array[Int] = Array::map(..)` | `[2, 4, 6]` | `[2, 4, 6]` |

`296` is an allocation address that moves with whatever ran before it, and `vibe check` is silent. #2911 landed the usability half (`dtd_map_callback_shape` reads the callback's **declared** return type), which is why row 1 renders; row 2 is what remained.

```
cannot render the result of `Array::map` -- its element type is not known here,
and rendering it would print a heap address instead of the array. Annotate the
binding (`let xs: Array[Int] = ...`) or, for `Array::map`, give the callback a
declared return type (`(v: Int) -> Int { .. }`) (#2912)
```

## No new refusal path

**#1445 already refuses exactly this, four lines up in the same function:**

> `nothing rendered it structurally AND the type is a declared aggregate with no renderer -- that is a missing derive(Show), and falling through here would print a pointer decimal.`

Declared aggregates were covered; arrays were the hole. So this widens one existing refusal by one case rather than inventing a mechanism.

It also stays out of the **#2773 trap**. Because the check sits *after* the shape machinery has run and failed, it never has to decide "is this an array?" from syntax — the pass that accumulated 63 bypass routes over 26 review rounds before being deleted (57 functions, 1,970 lines).

## No third name list

The signal is that **both existing shape functions returned empty** — `dtd_elem_preserving_shape` (the four builtins that reuse their argument's element type) and `dtd_map_callback_shape` (`Array::map`).

This family has diverged from hand-maintained name lists **four times**: #1460 (`Console::write_char` / `write_err_char`), #2903 (`sleep_blocking`), and both halves of #2905. A new list naming the array-returning builtins would be a fifth, and a name missing from it would silently reopen the pointer path. The predicate that knows those five names sits beside the two functions that define them, with a comment fixing that they are edited together.

A call in neither list never had a shape to begin with, so there is nothing to refuse and its behaviour is unchanged — fail-open here is the status quo, not a new hole.

## Cost, measured rather than predicted

The same one #2475 records for structural `==`: **a render that is written but never executed no longer compiles.** The [estimate on the issue](https://github.com/mizchi/vibe-lang/issues/2912#issuecomment-5746649124) was "at most 15 sites, every one a test" — the real number of files the refusal rejects is **0**, because all 15 resolve. 1321/1321 unit-test files pass.

`vibe check` does not run this pass, so it stays clean; the refusal lands on `vibe build` / `vibe run` / `vibe test`.

## The compiler still builds itself

This was the open risk. #2475 had to leave some rungs as runtime traps **precisely because the compiler's own sources reached them** (`List` and `AvlTree` at a formal argument), and refusing there would have rejected the compiler. A full seed → stage1 → stage2 build exits 0 under this refusal, so `lib/@vibe/compiler` contains no unresolved array render.

## The gate, and why the green controls carry it

`fixtures/err_array_map_render_refused.vibe` plus a gate in the early lane, asserting the **message and the edit it names** — the standard the #2475 loop directly above it uses.

The three **green controls** are the part that actually constrains the change: a predicate that refused *everything* would satisfy the refusal rows and still be catastrophic, so the controls are the three spellings that do resolve. A refusal-only assertion cannot see that failure mode.

```
GREEN (stage2 with the refusal):    unrenderable array refusal ok (message names the edit; 3 resolvable spellings still compile)
RED   (pre-refusal stage2 3260d45): FAIL: fixtures/err_array_map_render_refused.vibe compiled; expected a compile-time refusal (#2912)
```


mizchi:
Verification complete. The open risk is cleared and the blast radius turned out to be smaller than the issue's estimate.

## The compiler still builds itself

This was the question that mattered. #2475 had to leave some rungs as runtime traps **precisely because the compiler's own sources reached them** (`List` and `AvlTree` at a formal argument), and refusing there would have rejected the compiler itself.

A full seed → stage1 → stage2 build **exits 0** under this refusal, so `lib/@vibe/compiler` contains no unresolved array render. No narrowing needed.

## Behaviour, re-measured under the refusal

| program | before | after |
|---|---|---|
| `Array::map(xs, (v: Int) -> Int { v * 2 })` | `[2, 4, 6]` | `[2, 4, 6]` |
| `Array::map(xs, (v) -> v * 2)` | **`296`** | **refused** |
| `Array::filter(xs, (v) -> v > 1)` | `[2, 3]` | `[2, 3]` |
| `let ys: Array[Int] = Array::map(..)` | `[2, 4, 6]` | `[2, 4, 6]` |

Only the silently-wrong row moved.

```
cannot render the result of `Array::map` -- its element type is not known here,
and rendering it would print a heap address instead of the array. Annotate the
binding (`let xs: Array[Int] = ...`) or, for `Array::map`, give the callback a
declared return type (`(v: Int) -> Int { .. }`) (#2912)
```

## Blast radius: zero, not fifteen

**1321/1321 unit-test files pass.** The [estimate on the issue](https://github.com/mizchi/vibe-lang/issues/2912#issuecomment-5746649124) was "at most 15 sites, every one a test" — the real number of files the refusal rejects is **0**, because all 15 resolve. That estimate was an upper bound and it behaved like one.

The **early gate lane also runs to its last gate (39/39)**, so no existing refusal contract moved.

## The gate, and why the green controls are the load-bearing half

Added `fixtures/err_array_map_render_refused.vibe` plus a gate in the early lane. It asserts the **message and the edit it names**, not merely that something was refused — the standard the #2475 loop directly above it uses.

But the three **green controls** are the part that actually constrains the change. A predicate that refused *everything* would satisfy the refusal rows and still be catastrophic; the controls are the three spellings that do resolve, so over-refusal fails the gate instead of passing quietly. A refusal-only assertion cannot see that failure mode.

Both directions proven on artifacts that can be re-run:

```
GREEN (stage2 with the refusal):  unrenderable array refusal ok (message names the edit; 3 resolvable spellings still compile)
RED   (pre-refusal stage2 3260d45): FAIL: fixtures/err_array_map_render_refused.vibe compiled; expected a compile-time refusal (#2912)
```

I copied the fixed build aside **before** re-running. Last time a rebuild overwrote the pre-fix artifact under the same generation name and nearly cost the ability to reproduce the Red — a pasted number is not the same as a re-runnable proof.

## One small thing worth recording

`check_gate_registry.sh` rejected my first closing banner because `is_success_banner` keys on the literal `ok`. That is the ratchet working as designed, and it caught the wording before CI did.

## Full check list

1321/1321 unit corpus · self-build exits 0 · early gate lane 39/39 · gate Red+Green with three green controls · `vibe fmt --check` · gate registry / wiring / portability · fixture execution + snapshot checks · pre-commit review-lint.

---

mizchi:
**CI was red on my own change, and I have reverted the part that caused it.** This PR is back to being #2912 only, which is the state that was verified.

## What went wrong

I put a second fix (#2914(b), a diagnostic for `vibe grep --where` instead of a trap) on this branch, because my instructions forbid pushing to a different branch without permission. Reading the heap needs `Profiler::heap_bytes`, a **capability builtin**, so `Profiler` joins the effect row and propagates up every caller.

I claimed the ripple had stopped **twice**, and was wrong both times — first at the CLI dispatch, then at `selfhost_cli_user_grep_args`. Each time I traced a single call path and concluded from it; the compiler found the next one. Tracing it properly shows where it actually lands:

```
cli_adapter.vibe:1242  fn cli_main_with_lanes(..)  with Exception + Fs + Env + Stdin + Stdout
cli_adapter.vibe:2754  fn cli_main_compile_only()  with Exception + Fs + Env + Stdin + Stdout
```

That is the **compile-only lane**, which exists to carry as little as possible and which #2492 / #2497 are open to strip further. Threading a profiling capability into it to improve one error message pushes against work already in flight — a design decision, not a bug I should patch through unilaterally.

So I reverted rather than pressed on. 29 of the red checks across two commits were this one cause.

## This PR is unaffected

The revert touches only `grep_fs.vibe`, `grep_fs_test.vibe`, `runtime/index.vpkg` and `user_dispatch.vibe`. **None of #2912's files appear in it** — verified by name, not assumed.

Re-verified on a stage2 built from the reverted tree (`3614622`, matching HEAD):

```
[compiler-gate] unrenderable array refusal ok (message names the edit; 3 resolvable spellings still compile)
```

The self-build exits 0 again, and the earlier #2912 numbers stand: 1321/1321 unit corpus, early lane 39/39, gate Red+Green.

## #2914 stays open, with the blocker now named

Two things from the attempt are worth keeping, and are on the issue:

- The guard belongs **before** typing each file. The allocation that crosses the ceiling is inside `grep_fs_typed_context`, and once it traps there is nothing left to report with.
- The budget has to be a **parameter**, not a constant — otherwise the firing path is reachable only through its predicate, and a predicate that is right and a loop that never consults it look identical from outside.

The open question for whoever takes it: may `Profiler` enter the compile-only entry, or does the tier need a signal that is not a capability?

---

